PR #14, opened 2019-10-03
Add build cache for duniverse projects

talex5:
This avoids pinning all the opam packages, relying on the project to be entirely self-contained.

Requested by @avsm.

I'm not sure this is a good idea however, since it means *everything* must be in the duniverse, including depexts.


talex5:
This is now a more minimal solution that just adds a build cache to duniverse builds (we could add it to everything, but it's experimental so start with these projects, which really need it). It also avoids pinning empty opam files, since realworldocaml needs this.

PR #21, opened 2019-10-08
Pipeline: Install precise version of the dependencies

kit-ty-kate:
The CI doesn't install depexts of test dependencies, this PR fixes this issue

### DIFF
--- a/service/opam_build.ml
+++ b/service/opam_build.ml
@@ -68,6 +68,6 @@ let dockerfile ~base ~info ~repo =
   workdir "/src" @@
   run "sudo chown opam /src" @@
   pin_opam_files groups @@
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
I was wrong on what was going wrong. This should now fix the pipeline.

The problem was that only the packages name were gotten and installed but some depext might be changed/added/removed between versions.

talex5:
Is this thing with `--dry-run` just the same as using `--show-actions` and `awk '/- install/{print $3 "." $4}'`?

kit-ty-kate:
probably yes, I'm not really a awk expert but yes it's to get packages of the form `name.ver`

talex5:
My way doesn't work because sometimes there's a star at the end. Your dry run system seems to work nicely though!

PR #23, opened 2019-10-09
Pipeline: Install precise version of the dependencies

talex5:
Possibly simpler version of #21.

/cc @kit-ty-kate


talex5:
This doesn't work because the version sometimes has a star on the end! e.g. `index.dev*`

PR #38, opened 2019-10-14
Only call opam depext on the packages being tested if it contains some depexts

kit-ty-kate:
Mitigate ocaml/opam-depext#121 encountered by some users of ocaml-ci


talex5:
Rebased as #39.

PR #39, opened 2019-10-14
Only call opam depext on the packages being tested if it contains some depexts

talex5:
Rebased #38, plus sort the output from find to make caching more reliable.


talex5:
Also, squash all the depext commands to make things a bit faster.

PR #49, opened 2019-10-23
Add linting pipeline stage to projects with .ocamlformat files

CraigFe:
This is a port of https://github.com/ocaml-ci/ocurrent/pull/72 to `ocaml-ci` directly.

### DIFF
--- a/service/analyse.ml
+++ b/service/analyse.ml
@@ -25,6 +25,7 @@ module Examine = struct
     type t = {
       is_duniverse : bool;
       opam_files : string list;
+      ocamlformat_version : string option;
     }
     [@@deriving yojson]
 
@@ -38,14 +39,65 @@ module Examine = struct
     let opam_files t = t.opam_files
 
     let is_duniverse t = t.is_duniverse
+
+    let ocamlformat_version t = t.ocamlformat_version
   end
 
   let id = "ci-analyse"
+
+  let ocamlformat_version_from_string =
+    let re =
+      Re.(
+        seq
+          [
+            start;
+            rep space;
+            str "version";
+            rep space;
+            char '=';
+            rep space;
+            group (rep1 @@ diff graph (set "#"));
+            rep space;
+            eol;
+          ]
+        |> compile)
+    in
+    fun path ->
+      Re.exec_opt re path |> function
+      | Some g -> Some (Re.Group.get g 1)
+      | None -> None
+
+  let ocamlformat_version_from_file job path =
+    let ( let+ ) = Lwt.Infix.( >|= ) in
+    if not (Sys.file_exists path) then
+      let () = Current.Job.log job "No .ocamlformat file found" in
+      Lwt.return (Ok None)
+    else
+      let+ versions = Lwt_io.with_file ~mode:Lwt_io.input path (fun channel ->
+          Lwt_io.read_lines channel
+          |> Lwt_stream.filter_map ocamlformat_version_from_string
+          |> Lwt_stream.to_list
+        )
+      in
+      match versions with
+      | [ v ] ->
+          let () =
+            Current.Job.log job "Found OCamlformat version '%s' in dotfile" v
+          in
+          Ok (Some v)
+      | _ -> Error (`Msg "Unable to parse .ocamlformat file")
+
+  let get_ocamlformat_version job root =
+    Fpath.(to_string (root / ".ocamlformat")) |> ocamlformat_version_from_file job
+    >|= function
+    | Ok result -> result
+    | Error (`Msg e) -> failwith e
 
   let build ~switch No_context job src =
     Current.Job.start job ~level:Current.Level.Harmless >>= fun () ->
     Current_git.with_checkout ~switch ~job src @@ fun tmpdir ->
     let is_duniverse = is_directory (Filename.concat (Fpath.to_string tmpdir) "duniverse") in
+    get_ocamlformat_version job tmpdir >>= fun ocamlformat_version ->
     let cmd = "", [| "find"; "-name"; "*.opam" |] in
     Current.Process.check_output ~cwd:tmpdir ~switch ~job cmd >|= Stdlib.Result.map @@ fun output ->
     let opam_files =
@@ -62,7 +114,7 @@ module Examine = struct
               true
         )
     in
-    let r = { Value.opam_files; is_duniverse } in
+    let r = { Value.opam_files; is_duniverse; ocamlformat_version } in
     Current.Job.log job "@[<v2>Results:@,%a@]" Yojson.Safe.(pretty_print ~std:true) (Value.to_yojson r);
     r
 
--- a/service/lint.ml
+++ b/service/lint.ml
@@ -17,4 +17,4 @@ let v_from_opam ~ocamlformat_version ~base ~src =
   let img =
     Docker.build ~label:"OCamlformat" ~pull:false ~dockerfile (`Git src)
   in
-  Docker.run ~label:"lint" img ~args:[ "dune"; "build"; "@fmt" ]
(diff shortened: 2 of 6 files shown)


CraigFe:
Currently uses `list_map` to mock out a selective applicative. 

Example of a pipeline that uses the linting stage:
image

And one that doesnt use the linting stage:
image

service/lint.ml
```diff
@@ -0,0 +1,19 @@
+open Current.Syntax
+module Docker = Current_docker.Default
```
talex5: This will need updating to master (we have multiple Docker engines now). It's easiest just to take the base image as a string.

service/lint.ml
```diff
@@ -0,0 +1,19 @@
...
+
+let format_dockerfile ~base ~ocamlformat_version =
+  let open Dockerfile in
+  from (Docker.Image.hash base)
+  @@ run "git -C /home/opam/opam-repository pull"
```
talex5: This will prevent most caching of the installation of ocamlformat, since every build will get a slightly different version of the repository. Is it really needed?

service/lint.ml
```diff
@@ -0,0 +1,19 @@
...
+
+let v_from_opam ~ocamlformat_version ~base ~src =
+  let dockerfile =
+    let+ base = base and+ ocamlformat_version = ocamlformat_version in
+    format_dockerfile ~base ~ocamlformat_version
+  in
+  let img =
+    Docker.build ~label:"OCamlformat" ~pull:false ~dockerfile (`Git src)
+  in
+  Docker.run ~label:"lint" img ~args:[ "dune"; "build"; "@fmt" ]
```
talex5: This doesn't appear to get run with access to the source code, which seems strange. Maybe it should just be another build step?

service/pipeline.ml
```diff
@@ -65,11 +65,23 @@ let build_with_docker ~repo src =
     build "debian-10-ocaml-4.08";
   ]
 
+let lint ~analysis ~src =
+  analysis
+  |> Current.map Analyse.Analysis.ocamlformat_version
+  |> Current.map (function Some v -> [ v ] | None -> [])
```
talex5: (we could simplify this by adding a new combinator to OCurrent for optional steps)

service/analyse.ml
```diff
@@ -38,14 +39,67 @@ module Examine = struct
...
+    fun path ->
+      Re.exec_opt re path |> function
+      | Some g -> Some (Re.Group.get g 1)
+      | None -> None
+
+  let ocamlformat_version_from_file path =
+    let ( let+ ) = Lwt.Infix.( >|= ) in
+    let ( let* ) = Lwt.Infix.( >>= ) in
+    if not (Sys.file_exists path) then
+      let () = Logs.info (fun m -> m "No .ocamlformat file found") in
```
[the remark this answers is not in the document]
CraigFe: Done. I wanted to pass a `logger` around rather than a `job`, but couldn't get the polymorphism of the `format` strings to line up correctly...

service/lint.ml
CraigFe: Not entirely sure what you mean by 'take the base image as a string'. Pretty-print it and use that as the host?

service/lint.ml
CraigFe: Indeed; it's not really needed.

service/lint.ml
CraigFe: Not sure what you mean here; the source is built into the image on the previous step, no?

CraigFe:
I've tested it with `ocaml-ci-local` on various repositories on my machine, but not on the full GitHub version as I don't have a non-disruptive way to test that.

service/lint.ml
talex5: Ah, actually this isn't a problem. Although we build on multiple engines, we only lint using `Default`.

service/lint.ml
talex5: No, it's just made available in the context. But there are no `COPY` steps, so it isn't used for anything. Do you have an example of the check failing on something?

CraigFe: Of course you're right. I managed to refactor away a `copy` step because I didn't understand what it was doing... have now included it again.

talex5:
Seems to be working now! But if it fails then it reports the error as `Command "docker" "--context" "ci.ocamllabs.io" "run" "--rm" "-i" "sha256:1bc3be338b6dd1648678a60436a868605c52c60e3e0443fd92a7e846f959e1ba" 
"dune" "build" "@fmt" exited with status 1`. Maybe replace it with something more friendly?

CraigFe:
Updated to include https://github.com/ocaml-ci/ocurrent/pull/78, and now any prerequisite stages to an unselected step are executed. e.g. the pull in the following case:

image

I don't think this was happening before, and seems sub-optimal to me.

talex5:
OCurrent always runs things as soon as it can and you need to use `Current.gate` if you want to prevent that. In this case, I don't think we do want to prevent it, because we're bound to want the format base image sooner or later.

Note that the old version would already have started fetching it when it didn't know what the list contained. The difference is that it now doesn't cancel the operation once it discovers that the list is empty.

Always delaying running any step inside a list pipeline until the list is ready and non-empty would also be a reasonable choice, but that would require more changes, and I'm not sure we want that behaviour here anyway.

CraigFe:
In this case, I agree with you that it's worth pulling the image each time, but I find the semantics of always performing such operations at least once (even in an empty list case) confusing in general. However, it seems OK to require a gate if the alternative case ever arises :slightly_smiling_face:

CraigFe:
Have made an attempt to fix the error message from within the Docker container using:
```ocaml
  Docker.run ~label:"lint" img ~args:[ "sh"; "-c"; "dune build @fmt || (echo \"Error: dune build @fmt failed\"; exit 1)" ]
```
Which gives a log output like this:
```
@@ -19,7 +19,7 @@ open Cmdliner
 open Astring
 module IntSet = Core.IntSet
 
-module type S=sig
+module type S = sig
   include Core.S
 
   val run :
Error: dune build @fmt failed
2019-10-28 12:49.18: Job failed: Command "docker" "run" "--rm" "-i" "sha256:64295dc7e27737ac741e25bfbacf6144486456391d9f4743ca2d27cad5a39339" 
"sh" "-c" "dune build @fmt || (echo "Error: dune build @fmt failed"; exit 2)" exited with status 2
```

talex5:
I made a small change so that the lint step also works in service mode, not just local testing mode: https://github.com/talex5/ocaml-ci/commit/01f65abc16080204d5184054b422a892a3fabc93

However, it breaks some existing builds: https://github.com/avsm/duniverse.git#improve-cmdliner-args fails with `"Unable to parse .ocamlformat file"`.

I think we should also remove the `Error:` prefix from the error string (which I suggested - sorry!). The echo command gets logged whatever happens, and the log analyser always thinks this is an interesting string. Without the prefix, we can make a more specific match that only matches at the start of a line.

CraigFe:
Thanks for the service integration; I've cherry-picked it onto this branch. I have also removed the `Error:` prefix from the error message as you suggest.

The https://github.com/avsm/duniverse.git#improve-cmdliner-args PR is broken because there is no OCamlformat version in the `.ocamlformat` file on that branch. @NathanReb has already fixed this issue in https://github.com/avsm/duniverse/pull/71, so the fix for the `improve-cmdliner-args` PR is to rebase onto Duniverse master; this seems fine to me.

CraigFe:
In fact, it looks like the `improve-cmdliner-args` branch was already merged in https://github.com/avsm/duniverse/pull/65, so perhaps the solution is to just delete the branch :slightly_smiling_face:

talex5:
Merged as #52 - thanks!

CraigFe:
For what it's worth, I think the current graph of a pipeline with inactive linting is confusing:

image

The 'option box' contains stages that are not dependent on the result of the option, and it's not clear why the `summarise` step should be able to run with an apparently unexecuted dependency.

PR #58, opened 2019-11-01
Index the analysis job too

talex5:
This means that errors during the analysis phase also appear in the web UI, etc.

### DIFF
--- a/service/pipeline.ml
+++ b/service/pipeline.ml
@@ -46,8 +46,8 @@ let lint ~analysis ~src =
       Lint.v_from_opam ~ocamlformat_version ~base ~src
     )
   |> Current.map (function
-      | Some () -> ()
-      | None -> ()
+      | Some () -> `Checked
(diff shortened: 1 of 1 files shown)


service/pipeline.ml
```diff
@@ -148,14 +153,15 @@ let v ~app () =
...
+                 (variant, x)
                )
              |> Current.list_seq
   in
   let index =
     let+ commit = head
+    and+ analysis = job_id analysis
     and+ jobs = jobs in
-    Index.record ~commit jobs
+    Index.record ~commit @@ ("ANALYSIS", analysis) :: jobs
```
avsm: Why so shouty here? Most of the other job names are lower case

talex5: It's to make it appear first in the list. Otherwise, if it fails then it's not obvious why the first items in the list haven't started (and "alpine" comes before "analysis", so there are a lot of them).

PR #75, opened 2019-11-19
Cache dockerfile generation

talex5:
This isn't very clever, but it might help a bit. I tried using `lru` instead of this dumb system of wiping the cache when it gets too big, but for some reason that was much slower than having no cache at all!

Another option would be to replace the generic `Docker.build` with a custom build operation. Then we could generate the Dockerfile as part of the build step rather than on every evaluation of the pipeline.

### DIFF
--- a/service/opam_build.ml
+++ b/service/opam_build.ml
@@ -54,7 +54,14 @@ let get_opam_packages = List.fold_left (fun acc (_, _, pkgs) -> pkgs @ acc) []
 
 let download_cache = "--mount=type=cache,target=/home/opam/.opam/download-cache,uid=1000"
 
-let dockerfile ~base ~info ~repo ~variant =
+type key =  {
+  base : string;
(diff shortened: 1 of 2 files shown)


service/opam_build.ml
```diff
@@ -83,3 +90,16 @@ let dockerfile ~base ~info ~repo ~variant =
     ) @@
   copy ~chown:"opam" ~src:["."] ~dst:"/src/" () @@
   run "%s opam install -tv ." caches
+
+let cache = Hashtbl.create 10000
```
avsm: Why not just make this `7` or something small instead and let it size up?  It'll keep the default memory profile of ocurrent a bit lower (not relevant for Docker builds I realise, but possibly if used in other contexts)

talex5: This is in ocaml-ci, not OCurrent, so it won't affect anything else.

PR #77, opened 2019-11-29
Detect dune projects

Julow:
Hi !

This PR adds detection of `dune-project` files and parse them using Sexplib.

This is usefull in a follow-up PR to detect if OCamlformat is vendored, to avoid installing it from Opam or failing on the version check.

Sexplib was already an indirect dependency. I choose to parse `dune-project` files instead of just looking at `.opam` files because we may need other fields in the future (eg. dune's version for formatting).

Thanks !

### DIFF
--- a/ocaml-ci-service.opam
+++ b/ocaml-ci-service.opam
@@ -21,4 +21,5 @@ depends: [
   "ocaml-ci-api"
   "conf-libev"
   "dockerfile" {>= "6.3.0"}
+  "sexplib"
 ]
--- a/service/analyse.ml
+++ b/service/analyse.ml
@@ -26,10 +26,13 @@ module Examine = struct
   end
 
   module Value = struct
+    type ocamlformat_version = Version of string | Vendored
+    [@@deriving yojson]
+
     type t = {
       is_duniverse : bool;
       opam_files : string list;
-      ocamlformat_version : string option;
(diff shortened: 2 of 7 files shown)


CraigFe:
It's difficult to justify merging this independently of #78, as it is dead code currently. Without a way to exercise this code path (and no tests), I'd rather merge this diff along with a corresponding feature when I have more assurance that it works in practice.

Julow:
That makes sense. I split the PRs because I feel like this one may evolve a bit during reviews (eg. why parsing dune-project if we have the opam files).
I think I can reverse the dependencies between the PRs, implementing #78 differently and improving it in this PR?

The list of detected project is logged (is it too verbose?). To test this, is there an example repo used as a test ?

Julow:
I rebased this PR on top of #78. The detected projects are used to detect if OCamlformat is vendored instead of looking at opam files.
This PR now makes sense on its own and is optional.

CraigFe:
Closing this as we decided to go a different route with vendored OCamlformat in https://github.com/ocurrent/ocaml-ci/pull/78 that is more inline with other analysis steps being performed

If we ever need to parse `dune-project` files for other reasons (like installing appropriate OCamlformat versions), this will form a good starting point :slightly_smiling_face:

PR #78, opened 2019-11-29
Support vendored ocamlformat

Julow:
~~Rebased on top of https://github.com/ocurrent/ocaml-ci/pull/77~~

To avoid installing OCamlformat and having to parse its version from the `.ocamlformat` file.
Projects that have OCamlformat vendored (eg. OCamlformat itself) may not set a `version`.

### DIFF
--- a/lib/analyse.mli
+++ b/lib/analyse.mli
@@ -1,12 +1,9 @@
 module Analysis : sig
-  type ocamlformat_version = Version of string | Vendored of string
-  [@@deriving yojson, eq]
-
   type t
 
   val opam_files : t -> string list
   val is_duniverse : t -> bool
-  val ocamlformat_version : t -> ocamlformat_version option
+  val ocamlformat_source : t -> Analyse_ocamlformat.source option
 
   val of_dir : job:Current.Job.t -> Fpath.t -> (t, [ `Msg of string ]) result Lwt.t
 end
--- a/lib/analyse.mli
+++ b/lib/analyse.mli
@@ -1,5 +1,5 @@
 module Analysis : sig
-  type ocamlformat_version = Version of string | Vendored
+  type ocamlformat_version = Version of string | Vendored of string
   [@@deriving yojson, eq]
 
   type t
--- a/lib/analyse.mli
+++ b/lib/analyse.mli
@@ -1,9 +1,11 @@
 module Analysis : sig
+  type ocamlformat_version = Version of string | Vendored
+
   type t
 
   val opam_files : t -> string list
   val is_duniverse : t -> bool
-  val ocamlformat_version : t -> string option
+  val ocamlformat_version : t -> ocamlformat_version option
 
   val of_dir : job:Current.Job.t -> Fpath.t -> (t, [ `Msg of string ]) result Lwt.t
 end
--- a/lib/analyse.mli
+++ b/lib/analyse.mli
@@ -1,5 +1,6 @@
 module Analysis : sig
   type ocamlformat_version = Version of string | Vendored
+  [@@deriving yojson, eq]
 
   type t
 
--- a/lib/analyse_ocamlformat.ml
+++ b/lib/analyse_ocamlformat.ml
@@ -1,6 +1,9 @@
 open Lwt.Infix
 
-type version = Version of string | Vendored of string
+type source =
+  | Opam of { version : string }
+  | Vendored of { path : string }
+[@@deriving yojson,eq]
 
 let ocamlformat_version_from_string =
     let re =
@@ -44,15 +47,15 @@ let ocamlformat_version_from_file job path =
         Ok (Some v)
     | _ -> Error (`Msg "Unable to parse .ocamlformat file")
 
-let get_ocamlformat_version ~opam_files job root =
+let get_ocamlformat_source job ~opam_files root =
   let proj_is_ocamlformat p = Filename.basename p = "ocamlformat.opam" in
   match List.find_opt proj_is_ocamlformat opam_files with
   | Some opam_file ->
     let path = Filename.dirname opam_file in
-    Lwt.return (Some (Vendored path))
+    Lwt.return (Some (Vendored { path }))
   | None ->
     Fpath.(to_string (root / ".ocamlformat")) |> ocamlformat_version_from_file job
     >|= function
-    | Ok (Some v) -> Some (Version v)
+    | Ok (Some version) -> Some (Opam { version })
     | Ok None -> None
     | Error (`Msg e) -> failwith e
--- a/lib/analyse_ocamlformat.ml
+++ b/lib/analyse_ocamlformat.ml
@@ -1,5 +1,7 @@
 open Lwt.Infix
 
+type version = Version of string | Vendored
+
 let ocamlformat_version_from_string =
     let re =
       Re.(
@@ -42,8 +44,13 @@ let ocamlformat_version_from_file job path =
         Ok (Some v)
     | _ -> Error (`Msg "Unable to parse .ocamlformat file")
 
-let get_ocamlformat_version job root =
-  Fpath.(to_string (root / ".ocamlformat")) |> ocamlformat_version_from_file job
-  >|= function
-  | Ok result -> result
-  | Error (`Msg e) -> failwith e
+let get_ocamlformat_version ~opam_files job root =
+  let proj_is_ocamlformat p = Filename.basename p = "ocamlformat.opam" in
+  if List.exists proj_is_ocamlformat opam_files then
+    Lwt.return (Some Vendored)
+  else
+    Fpath.(to_string (root / ".ocamlformat")) |> ocamlformat_version_from_file job
+    >|= function
+    | Ok (Some v) -> Some (Version v)
+    | Ok None -> None
+    | Error (`Msg e) -> failwith e
--- a/lib/analyse_ocamlformat.ml
+++ b/lib/analyse_ocamlformat.ml
@@ -47,7 +47,7 @@ let ocamlformat_version_from_file job path =
         Ok (Some v)
     | _ -> Error (`Msg "Unable to parse .ocamlformat file")
 
-let get_ocamlformat_source job ~opam_files root =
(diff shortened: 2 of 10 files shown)


Julow:
I simplified this PR to detect OCamlformat by only looking at opam files.
This way this PR only does one thing and #77 can be discussed on its own.

lib/analyse.mli
```diff
@@ -1,9 +1,12 @@
 module Analysis : sig
+  type ocamlformat_version = Version of string | Vendored
+  [@@deriving yojson, eq]
```
Julow: This could be defined in `Analyse_ocamlformat` only and still be part of the API. Also, the derivings are necessary in the implementation and in the tests but the tests could redefine them if we want to remove them from the API.

service/lint.ml
```diff
@@ -1,18 +1,28 @@
...
-  let open Dockerfile in
-  from (Docker.Image.hash base)
-  @@ run "opam depext ocamlformat=%s" ocamlformat_version
-  @@ run "opam install ocamlformat=%s" ocamlformat_version
-  @@ copy ~chown:"opam" ~src:["./"] ~dst:"./src" ()
-  @@ workdir "src"
+type ocamlformat_version = [
+  | `Vendored
+  | `Version of string
+]
```
Julow: Now that Analyse is a library, we could use it here (or just `Analyse_ocamlformat`).

lib/analyse_ocamlformat.mli
```diff
@@ -1,2 +1,3 @@
+type version = Version of string | Vendored
 
-val get_ocamlformat_version : Current.Job.t -> Fpath.t -> string option Lwt.t
+val get_ocamlformat_version : opam_files:string list -> Current.Job.t -> Fpath.t -> version option Lwt.t
```
CraigFe: This is an odd type signature now. Some suggestions for improvements:
 - keep `Job.t` first as this is distinct from the others;
 - shift `opam_files` after the `Fpath.t` argument to respect their relative dependency;
 - add comment to clarify that `opam_files` are file-paths relative to the root, not e.g. file contents or absolute paths.

```suggestion
val get_ocamlformat_version : Current.Job.t -> root:Fpath.t -> opam_files:string list -> version option Lwt.t
(* Get the OCamlformat version to use for linting a given project directory, given the locations of the [.opam] files in the repository relative to the project root *)
```

test/test_analyse.ml
```diff
@@ -111,4 +111,48 @@ let test_duniverse =
...
+      is_duniverse = true;
+      ocamlformat_version = Some Vendored;
+    }
+  in
+  expect_test "ocamlformat_vendored" ~project ~expected
+
+let test_ocamlformat_self =
+  let project =
+    let open Gen_project in
+    [ File ("ocamlformat.opam", opam); File (".ocamlformat", empty_file) ]
```
CraigFe: Have you tested this on OCamlformat itself? I remember there being other blocking problems with that (such as OCamlformat using different configurations for subtrees of its source).

service/lint.ml
```diff
@@ -1,18 +1,28 @@
...
 
-let v_from_opam ~ocamlformat_version ~base ~src =
+let ocamlformat ~ocamlformat_version ~base ~src =
```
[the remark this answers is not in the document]
Julow: This function is only about ocamlformat, should it be `Ocamlformat.v` or `v_ocamlformat` instead?
Any future lint job will need their own `v` or module I think.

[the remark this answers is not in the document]
CraigFe: It's not only about OCamlformat. It runs `dune build @fmt`, which applies linting to dune files too. I agree that other linting tasks (e.g. `lint-doc`) will at least be different functions here, but I think we can worry about that when it happens.

test/test_analyse.ml
Julow: That doesn't work on ocamlformat yet. Ocamlformat sources can be fixed so that `dune build @fmt` works.

service/pipeline.ml
```diff
@@ -44,13 +44,19 @@ let job_id x =
...
+    Conf.Builder_amd1.pull ~schedule:weekly "ocurrent/opam:alpine-3.10-ocaml-4.08"
+  in
   analysis
   |> Current.map Analyse.Analysis.ocamlformat_version
   |> Current.option_map (fun ocamlformat_version ->
-      let base =
-        Conf.Builder_amd1.pull ~schedule:weekly "ocurrent/opam:alpine-3.10-ocaml-4.08"
+      let ocamlformat_version =
+        let+ v = ocamlformat_version in
+        match v with
```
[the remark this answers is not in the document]
Julow: I did that.

lib/analyse_ocamlformat.ml
```diff
@@ -1,5 +1,7 @@
 open Lwt.Infix
 
+type version = Version of string | Vendored
```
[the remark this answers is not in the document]
Julow: I added more doc and changed that type. Tell me what you think.

service/lint.ml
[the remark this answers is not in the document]
Julow: It's more than `dune build @fmt`, it also installs OCamlformat and worry about it being vendored (but doesn't compute that).
I'm pretty sure every linting step will have to be called independently from `pipeline.ml`.
What is you don't like about `Lint.Ocamlformat.v` or `Lint.v_ocamlformat` ?

lib/analyse_ocamlformat.mli
Julow: I did most of that, what do you think?

service/lint.ml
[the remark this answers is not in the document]
CraigFe: I agree with you that every linting step will be called independently from `pipeline.ml`. I also agree with you that it does more than `dune build @fmt`. However, it's _not just_ an OCamlformat linting step, it's a general source linting step as triggered by Dune. `ocamlformat` is not a sufficient name for describing the linting responsibilities of this pipeline segment. 

Call it something like `v_fmt` if you're concerned by where to put the documentation linting in future, but it seems just as likely to me that we'll want to put that in a separate file, making this a premature concern :slightly_smiling_face:

lib/analyse_ocamlformat.mli
```diff
@@ -1,2 +1,8 @@
+type source =
+  | Opam of { version : string } (** Should install OCamlformat from Opam. *)
+  | Vendored of { path : string } (** OCamlformat is vendored. [path] is relative from the project's root. *)
```
CraigFe:
```suggestion
  | Vendored of { path : string } (** OCamlformat is vendored. [path] is relative to the project's root. *)
```

service/lint.ml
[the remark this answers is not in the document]
Julow: I see what you mean, this step doesn't necessary use (or may use more than) OCamlformat. `v_fmt` seems good !

lib/analyse_ocamlformat.mli
```diff
@@ -1,2 +1,8 @@
+type source =
+  | Opam of { version : string } (** Should install OCamlformat from Opam. *)
+  | Vendored of { path : string } (** OCamlformat is vendored. [path] is relative to the project's root. *)
+[@@deriving yojson,eq]
 
-val get_ocamlformat_version : Current.Job.t -> Fpath.t -> string option Lwt.t
+val get_ocamlformat_source : Current.Job.t -> opam_files:string list -> Fpath.t -> source option Lwt.t
+(** Detect the required version of ocamlformat or if it's vendored.
```
CraigFe:
```suggestion
(** Detect the required version of OCamlformat or if it's vendored.
```

lib/analyse_ocamlformat.mli
```diff
@@ -1,2 +1,8 @@
...
+    Vendored OCamlformat is detected by looking at file names in [opam_files]. *)
```
CraigFe:
```suggestion
    Vendored OCamlformat is detected by inspecting the [opam_files] paths, which must be specified relative to the given root directory. *)
```
Still needs more explanation to make it clear what `opam_files` really is, and I still think a label on `Fpath.t` is a good idea :slightly_smiling_face:

service/lint.ml
```diff
@@ -1,18 +1,26 @@
...
-  @@ run "opam install ocamlformat=%s" ocamlformat_version
-  @@ copy ~chown:"opam" ~src:["./"] ~dst:"./src" ()
-  @@ workdir "src"
-
-let v_from_opam ~ocamlformat_version ~base ~src =
+let v_fmt ~ocamlformat_source ~base ~src =
   let dockerfile =
-    let+ base = base and+ ocamlformat_version = ocamlformat_version in
-    format_dockerfile ~base ~ocamlformat_version
+    let open Dockerfile in
```
CraigFe: Not sure why you chose to inline this function here? It was clearer to me as it was written before.

lib/analyse_ocamlformat.ml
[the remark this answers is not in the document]
CraigFe: LGTM

lib/analyse_ocamlformat.mli
Julow: This function doesn't require that the path is relative.
I agree `~root` label feels better.

service/lint.ml
Julow: Fixed.

service/lint.ml
```diff
@@ -1,19 +1,28 @@
...
+      @@ run "opam install ocamlformat=%s" version
+  in
   from (Docker.Image.hash base)
-  @@ run "opam depext ocamlformat=%s" ocamlformat_version
-  @@ run "opam install ocamlformat=%s" ocamlformat_version
-  @@ copy ~chown:"opam" ~src:["./"] ~dst:"./src" ()
+  @@ run "opam install dune" (* Not necessarily the dune version used by the project *)
   @@ workdir "src"
+  @@ install_ocamlformat
+  @@ copy ~chown:"opam" ~src:["./"] ~dst:"./" ()
```
Julow: How is the install commands working in the vendored case if the copy is done after ?
Are depending on that the `base` image already contains the opam files ? Does it ?

PR #89, opened 2020-01-06
Move build and lint steps to lib/

Julow:
This PR moves the code for the build step and the lint step to `lib/`.
The goal is to later test this code.
The module type `DOCKER_CONTEXT` is added in `lib/s.ml` and replaces `Conf.BUILDER`, it abstracts some arguments to `Current_docker` and allow `lib/` to not depend on it.

An extra step could be to move `summarise` and maybe the list of system/version variants to `lib/`.


service/conf.ml
```diff
@@ -18,27 +18,46 @@ module Capnp = struct
...
+  let image_hash = Docker.Image.hash
+
+  let pull name =
+    Docker.pull ~schedule:pull_schedule name
+
+  let build ~label ~dockerfile source =
+    Docker.build ~timeout:build_timeout ~pool ~label ~pull:false ~dockerfile source
+
+  let run ~label image ~args =
+    Docker.run ~label ~pool image ~args
```
Julow: `~pool` was not passed to `run` for the lint step, now it is.

lib/opam_build.ml
```diff
@@ -116,3 +114,21 @@ let dockerfile ~base ~info ~repo ~variant =
...
+    if opam_files = [] then failwith "No opam files found!";
+    info
+  in
+  let dockerfile =
+    let+ base = Docker.pull ("ocurrent/opam:" ^ variant)
+    and+ repo = repo
+    and+ info = info in
+    dockerfile ~base:(Docker.image_hash base) ~info ~repo ~variant
+  in
+  let build = Docker.build ~label:"" ~dockerfile source in
```
Julow: I made the `~label` mandatory but in this case it doesn't look nice. Should I add a label here or make the argument optional ?

hannesm:
I have not looked into the details of this PR (sorry), but am a bit concerned this will make `ocaml-ci` very docker specific -- is this the goal in the end (if so, why abstract?)? -- I'd appreciate a more general interface to be used on platforms where "Docker" is not <the way to isolate your throwaway-jobs> (i.e. macos/windows/BSD).

Julow:
`ocaml-ci` depends on Docker for everything except for the analyse step. (`lib/` is not `ocurrent/`)
Is the code in `lib/` (a private library currently) expected to be reused in other projects ?

talex5:
Support for non-Docker platforms would be very welcome, but I don't think it needs to hold up this PR. If someone contributes e.g. BSD support then they can either use their own interface or generalise the Docker interface at that point.

hannesm:
ok sorry for the noise

lib/opam_build.ml
CraigFe: I say make it optional: it's optional in `Current_docker`, and I don't see a particular reason for introducing a difference in this case.

lib/lint.mli
```diff
@@ -0,0 +1,12 @@
+module Make (Docker : S.DOCKER_CONTEXT) : sig
+
+  val v :
+    analysis:Analyse.Analysis.t Current.t ->
+    source:Docker.source ->
+    [> `Checked | `Check_skipped ] Current.t
+  (** [v ~analysis ~src] runs the linting step:
+      - Checks formatting using "dune build @fmt".
+        Ensures OCamlformat is installed depending on [Analysis.ocamlformat_source].
```
CraigFe:
```suggestion
      - Ensures OCamlformat is installed depending on {!Analysis.ocamlformat_source}.
```

Also, these list items should probably be in the opposite order.

lib/dune
```diff
@@ -1,4 +1,4 @@
 (library
   (name ocaml_ci)
   (preprocess (pps ppx_deriving.std ppx_deriving_yojson))
-  (libraries logs current current_github current.cache ppx_deriving_yojson.runtime))
+  (libraries logs current current_github current.cache dockerfile ppx_deriving_yojson.runtime))
```
CraigFe: `service/dune` now no longer depends on `dockerfile`; you should remove it there,

service/pipeline.ml
```diff
@@ -43,9 +43,28 @@ let job_id x =
...
+
+  type image = Builder.Image.t
+
+  let image_hash = Builder.Image.hash
+
+  let pull ~schedule name =
+    Builder.pull ~schedule name
+
+  let build ~label ~dockerfile source =
+    Builder.build ~label ~pool:Builder.pool ~pull:false ~dockerfile source
```
CraigFe: I don't see why you specialised `~pull:false` here, even if that's the only value that we happen to use right now.

service/pipeline.ml
```diff
@@ -49,8 +49,8 @@ module Docker_of_builder (Builder : Conf.BUILDER) : Ocaml_ci.S.DOCKER_CONTEXT =
 
   let image_hash = Builder.Image.hash
 
-  let pull ~schedule name =
-    Builder.pull ~schedule name
+  let pull name =
+    Builder.pull ~schedule:weekly name
```
CraigFe: As in the `pull` case, this feels like unnecessary specialisation.

lib/lint.ml
```diff
@@ -21,11 +20,23 @@ module Make (Docker : S.DOCKER_CONTEXT) = struct
     @@ install_ocamlformat
     @@ copy ~chown:"opam" ~src:["./"] ~dst:"./" ()
 
-  let v_fmt ~ocamlformat_source ~base ~src =
+  let ocamlformat_run ~ocamlformat_source ~base ~src =
```
CraigFe: We discussed this before: this _isn't_ just an OCamlformat check! `run_fmt` or similar will do.

lib/opam_build.mli
```diff
@@ -1,7 +1,10 @@
-(** Generate a Dockerfile for building all the opam packages in the build context. *)
-val dockerfile :
-  base:string ->
-  info:Analyse.Analysis.t ->
-  repo:Current_github.Repo_id.t ->
+(** Build and test all the opam packages.
```
CraigFe:
```suggestion
(** Build and test all the opam packages in a given build context.
```
It's not building and testing _all_ the opam packages.

lib/opam_build.mli
```diff
@@ -1,7 +1,10 @@
...
+    [~variant] is the variant of the "ocurrent/opam" image.
+    [~repo] is used to identify a build cache for duniverse projects. *)
```
CraigFe: That's a weird thing to expose in documentation; if I wanted to see what it's used for, I'd read the code. Shouldn't it tell me what `repo` actually _is_?

lib/opam_build.ml
Julow: Fixed.

lib/dune
Julow: Done

service/pipeline.ml
Julow: Does `~pull:true` makes sense when building from a Dockerfile ? If that is useful in the future, would a separate function makes sense ? Or at least an optional argument with the right default ?

service/pipeline.ml
Julow: I think this is part of the global config and should be abstract to `lib`. I think it's not the job of `Lint` of `Opam_build` to set that value.

lib/opam_build.mli
Julow: Sorry, I couldn't find a use for this argument that fits. What would you think of resolving this earlier and taking an optional list of caches here instead ?

service/pipeline.ml
CraigFe: I don't think the use of a Dockerfile or otherwise really matters; `--pull` updates the base image specified in the Dockerfile.

service/pipeline.ml
CraigFe: I agree with you, but that wasn't what I was suggesting. 

You could lift this argument from the `Buidler` API to the `Lint` / `Opam_build` APIs, allowing those schedules to be different without needing to construct another `DOCKER_CONTEXT` module. This would define the schedules in `service/pipeline.ml` rather than `service/conf.ml` (which I think is slightly better, since it's one fewer difference between `Current_docker.S` and `Ocaml_ci.S.DOCKER_CONTEXT`).

lib/opam_build.mli
CraigFe: Resolving this earlier means not being able to test it :slightly_smiling_face: Why not just something simple like `[~repo] is the ID of the repository-under-test on GitHub`. 

You could even make it optional and have it default to `{  Github.Repo_id.owner = "local"; name = "test"  }`, which is what's being done in `Pipeline.local_test`.

lib/lint.ml
Julow: It is ! And I wanted to make that obvious (it takes `ocamlformat_source and calls `ocamlformat_dockerfile`).
I agree we should split concern so I moved the "run" part in an other function.

CraigFe: We'll have to agree to disagree. Your separation works, anyway.

Julow: Note that this function is not exposed.

service/pipeline.ml
Julow: I did it that way because passing that value is not something the current implementation does. My change shows that.
I agree that it's an arbitrary limitation, but I argue that it was already like that is the code.
I think this should be discussed in an other PR and when this is necessary and when we know how this value should change.

CraigFe: I think this is a limitation introduced by this PR. The current implementation defines the schedules of `lint` and `build` separately in `service/pipeline.ml` (which is where I would expect to see it):
- https://github.com/ocaml-ci/ocaml-ci/blob/master/service/pipeline.ml#L48
- https://github.com/ocaml-ci/ocaml-ci/blob/master/service/pipeline.ml#L69

lib/opam_build.mli
Julow: I updated the doc :)
I don't think we should give it a default value. This value will be used depending on a flag in `analysis`.

service/pipeline.ml
talex5: What matters is whether the Dockerfile specifies a tag (which might change) or a hash (which won't). I think ocaml-ci always uses hashes in the Dockerfiles, so `pull` doesn't do anything.

service/pipeline.ml
Julow: I put the `schedule` argument back and passed it through. I think this is not the right way and a refactor is needed.
I think the `Docker.pull` should be done in `pipeline.ml` (and hidden in `DOCKER_CONTEXT`)

lib/opam_build.mli
```diff
@@ -0,0 +1,11 @@
+(** Build and test all the opam packages in a given build context.
+    [~variant] is the variant of the "ocurrent/opam" image.
+    [~repo] is the ID of the repository-under-test on GitHub. *)
+val v :
+  docker:(module S.DOCKER_CONTEXT with type source = 'source) ->
+  pull_schedule:Current_cache.Schedule.t ->
```
CraigFe:
```suggestion
  schedule:Current_cache.Schedule.t ->
```
No need to expose this detail, I think.

Julow: I made the change. But as I argued [above](https://github.com/ocurrent/ocaml-ci/pull/89#discussion_r371179171), I think this logic should go into `pipeline.ml` (it needs to be better defined to be in lib/)

PR #90, opened 2020-01-09
Add OCaml 4.10 to the build matrix

kit-ty-kate:
In the effort to have the most packages available and tested on OCaml 4.10 at or before its release here is my proposal to add it to the build matrix of ocaml-ci. This is for instance needed for `mdx` and other packages (see https://github.com/realworldocaml/mdx/pull/204)

### DIFF
--- a/service/pipeline.ml
+++ b/service/pipeline.ml
@@ -78,6 +78,7 @@ let build_with_docker ~repo ~analysis src =
   let lint_result = lint ~analysis ~src in
   [
     (* Compiler versions:*)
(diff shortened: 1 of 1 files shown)


CraigFe:
It will also be necessary to ensure that 4.10 images are being built, so that the pipelines don't all fail on the `pull` step. 

This requires a change in `ocaml-version` [here](https://github.com/ocurrent/ocaml-version/blob/master/ocaml_version.ml#L128) and/or a change in `docker-base-images` [here](https://github.com/ocurrent/docker-base-images/blob/master/src/conf.ml#L73).

PR #104, opened 2020-02-12
Add status icons on the status page

Julow:
So it's easier to see which job failed.

### DIFF
--- a/api/client.ml
+++ b/api/client.ml
@@ -6,20 +6,25 @@ type variant = string
 
 module Ref_map = Map.Make(String)
 
+module State = struct
+  open Raw.Reader.JobInfo.State
+
+  type t = Raw.Reader.JobInfo.State.unnamed_union_t
+
+  let pp f =
+    function
+    | NotStarted -> Fmt.string f "not started"
+    | Aborted -> Fmt.string f "aborted"
+    | Failed m -> Fmt.pf f "failed: %s" m
+    | Passed -> Fmt.string f "passed"
+    | Active -> Fmt.string f "active"
+    | Undefined x -> Fmt.pf f "unknown:%d" x
+end
+
 type job_info = {
   variant : variant;
-  outcome : Raw.Reader.JobInfo.State.unnamed_union_t;
+  outcome : State.t;
 }
-
-let pp_state f =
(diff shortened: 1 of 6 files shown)


api/client.ml
```diff
@@ -6,21 +6,26 @@ type variant = string
 
 module Ref_map = Map.Make(String)
 
+module State = struct
+  open Raw.Reader.JobInfo.State
+
+  type t = Raw.Reader.JobInfo.State.unnamed_union_t
```
CraigFe:
```suggestion
  type t = unnamed_union_t
```

web-ui/github.ml
```diff
@@ -36,6 +36,26 @@ let github_branch_url ~owner ~name ref =
...
+  .statuses > li:before {
+    display: inline-block;
+    width: 1em;
+    margin-right: 0.5em;
+    margin-left: -1.5em;
+    text-align: center;
+    line-height: 1.1em;
+  }
+  .statuses > li.not-started:before { content: "●"; color:grey; }
+  .statuses > li.aborted:before { content: "●"; color:red; }
```
CraigFe:
```suggestion
  .statuses > li.aborted:before { content: "A"; color:red; }
```
I still prefer 'A' here, as is being done in Citty. A red dot doesn't look like 'aborted' to me, and this symbol is already being shared.

web-ui/github.ml
```diff
@@ -49,6 +69,22 @@ let breadcrumbs steps page_title =
     List.rev steps
   )
 
+let statuses ss =
+  let open Tyxml.Html in
+  let render_status (s, elms) =
+    let s_class =
```
CraigFe:
```suggestion
    let status_class_name =
```
I don't like the similarity with `a_class` which is something completely different.

talex5:
Having status icons here is really nice. It would be even more useful if we could do something different for `[SKIP]` results.

web-ui/github.ml
Julow: Good idea

Julow:
Having a different status for `[SKIP]` would be great ! I'll try to add it

Julow:
Note that I haven't been able to test this PR yet

CraigFe:
Great, thanks!

Issue #94, opened 2020-01-21
Lint documentation comments in the CI


CraigFe:
It might be nice to have a linting stage that checks that the documentation comments in the project are well-formed.

ODoc now supports a 'fatal warnings' flag that could be used for this purpose. This feature is scheduled to be released in ODoc 1.5 (see https://github.com/ocaml/odoc/issues/405).

PR #105, opened 2020-02-12
Add a stage to lint documentation errors

CraigFe:
This resolves https://github.com/ocurrent/ocaml-ci/issues/94.

This stage follows the same structure as the existing `lint` step, but shares some logic with the Opam build steps in order to install the appropriate dependencies before building the documentation.

### DIFF
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -12,7 +12,9 @@ let install_ocamlformat =
     run "opam depext ocamlformat=%s" version
     @@ run "opam install ocamlformat=%s" version
 
-let fmt_dockerfile ~base ~ocamlformat_source =
+let fmt_dockerfile ~base ~info ~variant ~for_user =
+  ignore (variant, for_user); (* (todo) *)
+  let ocamlformat_source = Analyse.Analysis.ocamlformat_source info in
   let open Dockerfile in
   from base
   @@ run "opam install dune" (* Not necessarily the dune version used by the project *)
@@ -23,7 +25,11 @@ let fmt_dockerfile ~base ~ocamlformat_source =
   @@ copy ~chown:"opam" ~src:["./"] ~dst:"./" ()
   @@ run "opam exec -- dune build @fmt || (echo \"dune build @fmt failed\"; exit 2)"
 
-let dockerfile ~base ~info ~variant ~for_user =
-  ignore (variant, for_user); (* (todo) *)
-  let ocamlformat_source = Analyse.Analysis.ocamlformat_source info in
-  fmt_dockerfile ~base ~ocamlformat_source
+let doc_dockerfile ~base ~info ~variant ~for_user =
+  let open Dockerfile in
+  Opam_build.install_project_deps ~base ~info ~variant ~for_user
+  @@ run "opam depext odoc"
+  (* Warnings-as-errors was introduced in Odoc.1.5.0 *)
+  @@ run "opam install dune odoc>=1.5.0"
+  @@ run "ODOC_WARN_ERROR=true opam exec -- dune build @doc \
+          || (echo \"dune build @doc failed\"; exit 2)"
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -50,7 +50,7 @@ let rec get_root_opam_packages = function
 
 let download_cache = "--mount=type=cache,target=/home/opam/.opam/download-cache,uid=1000"
 
-let dockerfile ~base ~info ~variant ~for_user =
+let install_project_deps ~base ~info ~variant ~for_user =
   let opam_files = Analyse.Analysis.opam_files info in
   let groups = group_opam_files opam_files in
   let root_pkgs = get_root_opam_packages groups in
@@ -61,10 +61,6 @@ let dockerfile ~base ~info ~variant ~for_user =
       run "sudo dnf install -y findutils" (* (we need xargs) *)
     else
       empty
-  in
-  let build_cmd =
-    copy ~chown:"opam" ~src:["."] ~dst:"/src/" () @@
-    run "opam exec -- dune build @install @runtest && rm -rf _build"
   in
   (if for_user then empty
    else comment "syntax = docker/dockerfile:experimental@sha256:ee85655c57140bd20a5ebc3bb802e7410ee9ac47ca92b193ed0ab17485024fe5") @@
@@ -79,5 +75,10 @@ let dockerfile ~base ~info ~variant ~for_user =
   crunch_list (List.map (fun pkg ->
       run {|test "$(opam show -f depexts: %s)" = "$(printf "\n")" || opam depext -ty %s|} pkg pkg) root_pkgs
     ) @@
-  (if Analyse.Analysis.is_duniverse info then run "opam pin remove $(opam pin -s) -n" else empty) @@
-  build_cmd
+  (if Analyse.Analysis.is_duniverse info then run "opam pin remove $(opam pin -s) -n" else empty)
+
+let dockerfile ~base ~info ~variant ~for_user =
+  let open Dockerfile in
+  install_project_deps ~base ~info ~variant ~for_user @@
+  copy ~chown:"opam" ~src:["."] ~dst:"/src/" () @@
+  run "opam exec -- dune build @install @runtest && rm -rf _build"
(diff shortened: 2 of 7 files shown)


lib/lint.ml
```diff
@@ -24,6 +24,27 @@ module Make (Docker : S.DOCKER_CONTEXT) = struct
...
+    in
+    let open Dockerfile in
+    from (Docker.image_hash base)
+    @@ run "opam depext odoc"
+    @@ run "opam install dune odoc>=1.5.0"
+    (* Warnings-as-errors was introduced in ODoc.1.5.0 *)
+    @@ workdir "src"
+    @@ copy ~chown:"opam" ~src:["./"] ~dst:"./" ()
+    @@ run "opam depext --update %s" packages
+    @@ run "opam install --deps-only --yes %s" opam_files
```
talex5: How will this work with the caching? It looks like it would force reinstalling all dependencies each each build.

CraigFe: Indeed, that looks like a mistake :slightly_smiling_face: We could just make the `copy` step last?

talex5: Yes, though you'll have to install just the opam files first so the install step works.

CraigFe: Done in https://github.com/ocurrent/ocaml-ci/pull/105/commits/313a788fece55261f5bcb70a131ec58ceed94c76. I've done some local testing to confirm that this is caching the dependency installation appropriately.

CraigFe:
Rebased over https://github.com/ocurrent/ocaml-ci/pull/108 to get the CI working.

CraigFe:
Rebased again. @talex5 / @avsm, can I get another review?

talex5:
CI is failing...

CraigFe:
Thanks; bad rebase. Should be fixed now.

service/pipeline.ml
```diff
@@ -69,7 +71,8 @@ let build_with_docker ~repo ~analysis source =
     "fedora", build (module Conf.Builder_amd3) @@ "fedora-30-ocaml-" ^ default_compiler;
     (* oraclelinux doesn't work in opam 2 yet: *)
     (* build (module Conf.Builder_amd3) @@ "oraclelinux-7-ocaml-" ^ default_compiler; *)
-    "lint", (lint_result, job_id lint_result);
+    "lint-src", (lint_fmt_result, job_id lint_fmt_result);
+    "lint-doc", (lint_doc_result, job_id lint_doc_result);
```
Julow: Are these names presented to the user ?
Even if no, what about `Formatting` and `Documentation` ? No need to mangle "lint" in the name, every lint steps will most probably be independent and shown separately to the user (no single "lint" step doing everything).

lib/lint.ml
```diff
@@ -35,10 +61,18 @@ module Make (Docker : S.DOCKER_CONTEXT) = struct
     in
     Docker.build ~label:"fmt" ~dockerfile source
 
+  (** An image with ODoc installed *)
+  let odoc_base_image ~base ~opam_files ~source =
+    let dockerfile =
+      let+ base = base and+ opam_files = opam_files in
+      `Contents (odoc_dockerfile ~base ~opam_files)
+    in
+    Docker.build ~label:"ODoc" ~dockerfile source
```
Julow: Nitpick: On https://github.com/ocaml/odoc, it's very rarely capitalized and when it is, it's `Odoc`.

lib/opam_build.ml
```diff
@@ -58,6 +58,11 @@ let pin_opam_files groups =
       )
   )
 
+let packages_of_opam_files x =
+  group_opam_files x
```
Julow: What about moving the logic that find the package names (from `group_opam_files`) to `Analysis` ?
This logic can benefit from having its own interface and unit tests and to be reused.

CraigFe: I don't think it's worthwhile right now: it's not part of the analysis step, and it's still possible to unit-test it from `Opam_build`.

service/pipeline.ml
CraigFe: They are visible; I've made them clearer in https://github.com/ocurrent/ocaml-ci/pull/105/commits/674cc7df3dfe8e8e757afb5bc7e681956905d273, but ultimately I think the solution here is to have a hierarchical (or graph-based) display of jobs.

I disagree with not including 'lint' in the name. "Documentation" could mean anything from '_building documentation_' to '_linting documentation_' to '_deploying documentation to web API_' (all of which we have plans to do).

lib/opam_build.ml
Julow: `Analysis` is not the analysis step but its result. But yes, it would be the first function of its kind there.
This function can only become more complex over time (especially `group_opam_files`), the comments in its implementation says that.
What about making it a field of `Analysis.t` ? For now, it can be computed outside of analysis but this might change as it get generalized.
My point is that this logic needs a proper abstraction.

service/pipeline.ml
Julow: The "Lint" prefix is unnecessary for now and can be added when an other documentation related job is added.

CraigFe: As discussed in person, the "Lint" prefix is helpful in describing what this job does (and any related jobs can be considered when they arise).

lib/opam_build.ml
CraigFe: As discussed in person, I don't think this is worth worrying about until either (a) we add unit tests or (b) the function becomes more complex.

talex5:
This failed on the first two packages I tried it on, e.g.
```
#10 [6/8] RUN opam depext --update opam-zi.dev
#10 5.608 # Detecting depexts using vars: arch=x86_64, os=linux, os-distribution=alpine, os-family=alpine
#10 8.503 [ERROR] No solution for opam-zi.dev: Your request can't be satisfied:
#10 8.503           - unknown package
```
How are you testing it? I'm using `ocaml-ci-local`.

I think we should be using the same code to generate the Dockerfile installation lines for the lint stages as for the main build stages.

Is `odoc` just a binary? Perhaps we could build that in a separate Docker build stage and then `COPY --from=odoc ...`. Then we could cache both the building of the project and the building of odoc.

CraigFe:
I've rebased and changed the approach slightly. The logic for installing Opam dependencies of the project is now properly shared with `opam_build.ml` by exposing some more logic inside that module.

Tested to be working on Alcotest, Irmin and OCurrent in `ocaml-ci-local` mode.

talex5:
I've deployed it... let's see what happens!

avsm:
Odoc is just a binary, and is installed that way in the duniverse_build.ml (install_bin).  @CraigFe, why is the doc target disabled in the duniverse build?  The intention is very much to build docs there, so that the stage can be further extended in the future to a deployment/staging branch for visual inspection.

CraigFe:
> The intention is very much to build docs there, so that the stage can be further extended in the future to a deployment/staging branch for visual inspection.

@avsm: I'm keen for that feature too, but it's not clear to me why it should apply only to Duniverse projects. Shouldn't Opam builds get a deployment/staging branch for their docs too? That's why I proposed taking it out of  `duniverse_build.ml`: right now it seems like duplicated functionality unique to Duniverse projects.

avsm:
But you've taken out the feature without adding it anywhere else for them.  Remember that right now, duniverse builds are distinct from opam builds.  I'm happy to refactor it when we add the staging support, but right now I'd like to ensure that it stays building correctly with duniverse projects.

CraigFe:
This linting step should still run for Duniverse builds currently, no? These jobs are defined outside of the regular build matrix.

It seems to be running correctly on e.g. `realworldocaml/book`:

image

talex5:
For me, `book` is now failing:
```
#125 4.218 The following dependencies couldn't be met:
#125 4.218   - rwo -> ocaml >= 4.10.0
#125 4.218       base of this switch (use `--unlock-base' to force)
```

https://ci.ocamllabs.io:8100/job/2020-04-21/184252-docker-build-be7e04

avsm:
Right, that'll fail because the duniverse build is different from the opam build. The doc linting step is trying to copy all the opam files, and to pin them, and then run depexts, and then install them, etc.  The duniverse build doesn't install opam deps, and uses the locally vendored copies instead with a single dune build.

There's a material difference in how the opam and the duniverse builds work.  I think in the interests of pushing this PR forward, it should leave them distinct, and we can work on converging them in a next step.

CraigFe:
Linting steps are now disabled for Duniverse projects, and I've reverted the change to not build docs during the build stages of Duniverse project.

The current solution uses a bind to conditionally execute linting according to the results of the analysis step.

PR #108, opened 2020-02-13
Constrain Alcotest to pre 1.0.0

CraigFe:
While we wait for 1.0.1 to become available in CI.

### DIFF
--- a/ocaml-ci-api.opam
+++ b/ocaml-ci-api.opam
@@ -19,3 +19,16 @@ depends: [
   "alcotest" {< "1.0.0" & with-test}
   "alcotest-lwt" {with-test}
 ]
+pin-depends: [
+  ["current.dev" "git+https://github.com/ocurrent/ocurrent.git#5814658ee0bc885c0e3b2417fe9fcfd8d30d0a08"]
+  ["current_ansi.dev" "git+https://github.com/ocurrent/ocurrent.git#5814658ee0bc885c0e3b2417fe9fcfd8d30d0a08"]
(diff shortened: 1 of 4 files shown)


CraigFe:
I'm not sure why the CI is still failing here.
```
#19 30.25 # Error: Conflict between the following libraries:
#19 30.25 # - "current_rpc" in _build/default/ocurrent/lib_rpc
#19 30.25 # - "current_rpc" in /home/opam/.opam/4.08/lib/current_rpc
#19 30.25 #   -> required by library "ocaml-ci-api" in
#19 30.25 #      /home/opam/.opam/4.08/lib/ocaml-ci-api
#19 30.25 # -> required by executable main in client/dune:4
```

talex5:
Looks like something changed in dune 2.2.0 - probably https://github.com/ocaml/dune/pull/3075.

Maybe we should use an opam pin instead of a submodule now?

talex5:
I've asked about this in https://github.com/ocaml/dune/issues/3139 in case there's a better solution.

CraigFe:
Done in https://github.com/ocurrent/ocaml-ci/pull/108/commits/556fc0e7a87bc848c9f559ced3aae1826cf16835. I'm not against pinning in principle, but:
 - it's messy here since we need to pin transitive dependencies in the same repository;
 - Dune's `.opam` file autogeneration doesn't support pinning (and likely never will, as I understand it).

talex5:
Fixed by #111 and #121.

Issue #107, opened 2020-02-13
Enable `lint` stage even if `.ocamlformat` is missing


pascutto:
For some projects, we might want to use dune formatting feature without ocamlformat, e.g. to format dune files only. 
This is currently not supported since the `.ocamlformat` file is the trigger for the `lint` stage and is not present in that scenario.

While you could parse the `dune-project` file to decide whether formatting should be enabled, an even simpler solution would just be to trigger it every time. 
It is always possible to disable the dune formatting:
- With `formatting disabled` for dune-lang >= 2.0
- By not having the `(using fmt ...)` stanza for dune-lang < 2.0

In both cases, `dune build @fmt` does not fail if dune formatting is disabled, it just does nothing.

PR #109, opened 2020-02-13
Lint project even if .ocamlformat is missing

CraigFe:
This resolves https://github.com/ocurrent/ocaml-ci/issues/107.

A simple solution that conditionally installs OCamlformat according to the result of the analysis phase.

### DIFF
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -2,38 +2,41 @@ open Current.Syntax
 
 module Make (Docker : S.DOCKER_CONTEXT) = struct
 
-  let ocamlformat_dockerfile ~base ~ocamlformat_source =
+  let install_ocamlformat =
     let open Dockerfile in
-    let install_ocamlformat =
-      match ocamlformat_source with
-      | Analyse_ocamlformat.Vendored { path } ->
-        let opam_file = Filename.concat path "ocamlformat.opam" in
-        copy ~chown:"opam" ~src:[ opam_file ] ~dst:opam_file ()
-        @@ run "opam pin add -k none -yn ocamlformat %S" path
-        (* Only the opam file is necessary to install the deps
-           [-k none] above ensures that opam doesn't try to make an installable package *)
-        @@ run "opam depext ocamlformat"
-        @@ run "opam install --deps-only -y ocamlformat"
-      | Opam { version } ->
-        run "opam depext ocamlformat=%s" version
-        @@ run "opam install ocamlformat=%s" version
-    in
+    function
+    | Analyse_ocamlformat.Vendored { path } ->
+      let opam_file = Filename.concat path "ocamlformat.opam" in
+      copy ~chown:"opam" ~src:[ opam_file ] ~dst:opam_file ()
+      @@ run "opam pin add -k none -yn ocamlformat %S" path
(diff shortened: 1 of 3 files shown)


lib/lint.ml
```diff
@@ -2,52 +2,52 @@ open Current.Syntax
...
-        let img = ocamlformat_base_image ~base ~ocamlformat_source ~source in
+    |> (fun ocamlformat_source ->
+        let img = fmt_base_image ~base ~ocamlformat_source ~source in
         run_fmt ~img
       )
-    |> Current.map (function
-        | Some () -> `Checked
-        | None -> `Check_skipped
-      )
+    |> Current.map (fun () -> `Checked)
```
Julow: Why not keeping `unit` here ?

lib/lint.ml
```diff
@@ -2,52 +2,52 @@ open Current.Syntax
...
       let+ base = base and+ ocamlformat_source = ocamlformat_source in
-      ocamlformat_dockerfile ~base ~ocamlformat_source
+      fmt_dockerfile ~base ~ocamlformat_source
     in
-    Docker.build ~label:"OCamlformat" ~dockerfile source
+    Docker.build ~label:"fmt" ~dockerfile source
 
   let run_fmt ~img =
-    Docker.run ~label:"lint" img ~args:[ "sh"; "-c"; "dune build @fmt || (echo \"dune build @fmt failed\"; exit 2)" ]
+    Docker.run ~label:"lint-fmt" img ~args:[ "sh"; "-c"; "dune build @fmt || (echo \"dune build @fmt failed\"; exit 2)" ]
```
Julow: What about a more descriptive label ? eg. "formatting"
This is shown to the users in the reports.

CraigFe: I don't think this is shown to the users at any point. The user-facing name is given here: https://github.com/CraigFe/ocaml-ci/blob/lint-dune-files/service/pipeline.ml#L72

lib/lint.ml
CraigFe: `` `Checked`` is used by the `summarise` step (https://github.com/CraigFe/ocaml-ci/blob/lint-dune-files/service/pipeline.ml#L103) after constructing the list of jobs (https://github.com/CraigFe/ocaml-ci/blob/lint-dune-files/service/pipeline.ml#L53-L73).

lib/lint.ml
Julow: True

lib/lint.ml
Julow: I think the map `` unit -> `Checked `` should be moved there. It is used in a very specific logic that is internal to `summarise`. The fact that the variant have to be open in `Lint.v` signature shows that, I think.

CraigFe:
Rebased over https://github.com/ocurrent/ocaml-ci/pull/108 to get the CI working.

CraigFe:
Now rebased over https://github.com/ocurrent/ocaml-ci/pull/111 (merged) rather than https://github.com/ocurrent/ocaml-ci/pull/108.

CraigFe:
Have locally verified that this works for `vocal-project/vocal#master`.

PR #115, opened 2020-02-19
Update to latest OCurrent API

talex5:


talex5:
Hmm, looks like pinning is broken in CI:

```
2020-02-19 14:53.17: WARNING: ignoring opam file "ocurrent/current.opam" as not in root or duniverse subdir
```

Looks like this changed in https://github.com/ocurrent/ocaml-ci/pull/81

PR #122, opened 2020-02-26
Fix check for pinning with a version number

talex5:
We need to check for a version number in the directory name, not the opam file name. Was broken by 70aa06c0ba.


talex5:
`realworldocaml` was previously failing with an opam constraint error due to the pinning. With this change, it's now failing with:

```
Fatal error: exception Configurator__V1.Fatal_error("Configurator scripts must be run with Dune. To manually run a script, use $ dune exec.")
```

PR #124, opened 2020-02-28
Upgrade to OCaml 4.10 to get new memory allocator

talex5:


avsm:
That was a refreshingly drama-free upgrade within days of a new release :-)

PR #126, opened 2020-02-29
Only install dependencies of top-level opam packages

talex5:
We still need to pin vendored opam files, but we don't necessarily need to install them. In particular, this change means that we don't try to install test dependencies of vendored packages.


talex5:
After deploying this change, the number of OK stages went from 7664 to 7677.

PR #130, opened 2020-03-03
Update to new Prometheus for faster GC stats

talex5:


talex5:
This shows the collection time with the old version:

gc-time-before

This is the time with the new one:

gc-time-after

(times are now around 40 us)

PR #146, opened 2020-04-17
Fix unprecise failure tracking in dynamic pipelines

kit-ty-kate:
This PR fixes an issue that can arise when the body of `builds` fails but `analysis` didn't.
This can create a green state even though nothing is checked. Currently it shouldn't as the only two static jobs do not count in the summary, but if we add another static check this can happen and it helps getting the coding style right for the rest of the pipeline.

### DIFF
--- a/service/pipeline.ml
+++ b/service/pipeline.ml
@@ -67,17 +67,21 @@ let build_with_docker ~repo ~analysis source =
      it's dynamic here in anticipation of future changes where the set of platforms
      to use comes from the analysis phase. *)
   let platforms =
-    let+ analysis = analysis in
-    if Analyse.Analysis.is_duniverse analysis then (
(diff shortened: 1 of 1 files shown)


talex5:
OK, I think I understand. There are three places where we could get an error:

1. During analysis.
2. During an individual build job.
3. In a per-platform build pipeline, but outside of an individual job.

At the moment, (3) can't happen, but if it did then the error would be silently ignored. With this change:

1. An error during analysis leads to an empty set of platforms, so the builds succeed, and we report the error against "(analysis)" as before.
2. An error during a build is handled by summarise, as before.
3. An error during the `list_map` outside of a build job results in the whole pipeline failing (which shouldn't happen, but at least it will get reported).

Is that right?

kit-ty-kate:
Yes that's right. Thanks

PR #147, opened 2020-04-17
Custom build

talex5:
Instead of using the generic `Docker.build` from OCurrent, use our own version. This has several advantages:

1. The build log starts with a meaningful comment saying which project we're building and which commit, rather than a generic build command with the hash of the Dockerfile. We could put more useful stuff here later (such as a version of the Dockerfile that can run without BuildKit).

2. The Docker context is no longer part of the key, so if we reassign a platform to a new build host, we don't have to rebuild everything. We can more easily use a cluster.

3. The Dockerfile is generated during the build, not before. This avoids keeping Dockerfiles in memory for archived builds and means we can make changes to the generated files without having to rebuild everything (unless we want to).

### DIFF
--- a/lib/build.ml
+++ b/lib/build.ml
@@ -23,7 +23,7 @@ module Op = struct
 
     let to_json { commit; analysis; base; variant; repo } =
       `Assoc [
-        "commit", `String (Current_git.Commit.id commit);
+        "commit", `String (Current_git.Commit.hash commit);
         "analysis", digest_analysis analysis;
         "base", `String (Raw.Image.digest base);
         "variant", `String variant;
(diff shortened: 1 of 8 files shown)


lib/build.ml
```diff
@@ -1,30 +1,88 @@
...
+    let dockerfile =
+      let base = Raw.Image.hash base in
+      Dockerfile.string_of_t (
+        if Analyse.Analysis.is_duniverse analysis then
+          Duniverse_build.dockerfile ~base ~repo ~variant
+        else
+          Opam_build.dockerfile ~base ~info:analysis ~variant
+      )
+    in
+    Current.Job.log job "@[<v2>Using Dockerfile:@,%a@]" Fmt.lines dockerfile;
```
avsm, line 52: Would it be possible to log the Dockerfile in a cut-and-paste-friendly way? A minor nit, but the intending makes it difficult :-)

talex5: Done - how does this look?

https://ci.ocamllabs.io/github/ocurrent/ocaml-ci/commit/40a6e807b83fe183365e66baad636bed39a9e93b/variant/4.10

PR #148, opened 2020-04-20
Duniverse: Avoid failure when ocaml_compilers field is invalid

kit-ty-kate:
This should fix the issue encountered in https://github.com/realworldocaml/book/pull/3302 if it happens again in the future.

cc @avsm

Unknown errors are slightly tricky, should we put an error message web-ui per-commit pages to let the users know what happened, if that's even possible? @talex5


avsm:
LGTM; I'm modifying duniverse to work with both x.y versions (which defaults to the latest patch version) and x.y.z (for a specific one). So this check is good.

talex5:
Unit-tests are failing.

talex5:
@kit-ty-kate : is this PR still needed? If so, it should be rebased. If not, let's close it.

talex5:
Please re-open if this is still needed.

PR #150, opened 2020-04-22
Always show the full command for each build steps in the logs

kit-ty-kate:
Recently buildkit switched its output to use `PROGRESS_NO_TRUNC` by default to show the full command whether or not the line is more than 80 characters: https://github.com/moby/buildkit/pull/1435

However it would still be interesting to get this behaviour regardless of the version of docker installed.


talex5:
Needs rebasing to fix `Option` problem.

kit-ty-kate:
done

PR #159, opened 2020-04-24
Avoid concurrent builds, as this confuses BuildKit

talex5:
Make sure we never build the same (commit, variant) twice at the same time, as this is likely to trigger BuildKit bug https://github.com/moby/buildkit/issues/1456. This often happens if you enable ocaml-ci for both an upstream repository and a fork, since for every PR there is a commit in both repositories, and half of the builds will always fail with a missing-output error.

### DIFF
--- a/lib/build.ml
+++ b/lib/build.ml
@@ -32,6 +32,28 @@ module Spec = struct
   let label t = t.label
 end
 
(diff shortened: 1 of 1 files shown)


avsm:
Nice catch!

PR #164, opened 2020-04-28
Get OCaml compiler version from base image

talex5:
This simplifies the compiler search for duniverse builds, and is useful for the solver branch too.

### DIFF
--- a/lib/analyse.ml
+++ b/lib/analyse.ml
@@ -56,6 +56,9 @@ module Analysis = struct
     | exception exn ->
       Fmt.failwith "Exception parsing dune-get: %a" Fmt.exn exn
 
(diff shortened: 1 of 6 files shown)


avsm:
Looks good to me, but how does it affect duniverse builds?

avsm:
Scratch that last comment. I just noticed the top of the diff which simplifies the matching of compilers.

talex5:
It's actually not quite right at the moment because duniverse wants e.g. 4.10, not 4.10.0.

avsm:
That's something we need to refine the meaning of. I think that if a package requests "4.10", that should translate to "the latest patch version". 

Otoh, some packages might specifically ask for "4.10.1" since they need a specific version. This is rarer, but can happen sometimes.

PR #167, opened 2020-04-30
Run solver in the analysis phase

talex5:
This is faster, and avoids the need to run separate Docker builds on platforms for which there is no solution. More importantly, it allows us to follow changes in opam-repository more closely, since updating that no longer requires pulling a new set of base images and rebuilding everything.

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -70,7 +70,7 @@ let install_project_deps ~base ~opam_files ~selection ~for_user =
   distro_extras @@
   workdir "/src" @@
   run "sudo chown opam /src" @@
-  run "cd ~/opam-repository && (git reset --hard %s || (git pull origin master && git reset --hard %s)) && opam update" commit commit @@
+  run "cd ~/opam-repository && (git reset --hard %s || (git pull origin master && git reset --hard %s)) && opam update -u" commit commit @@
   pin_opam_files groups @@
   env ["DEPS", String.concat " " non_root_pkgs] @@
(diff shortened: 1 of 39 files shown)


talex5:
This is missing support for opam's `pin-depends`, e.g. https://ci.ocamllabs.io:8100/job/2020-04-30/120634-ci-build-133485

talex5:
Also, in opam 1.x format files it looks like the `available` field is ignored: https://ci.ocamllabs.io/github/0install/0install/commit/680f1b2342a3d07533179f6aaed7e26cd51c3d8a/variant/debian-10-ocaml-4.02

avsm:
This seems like a big improvement even before pin-depends support. As far as I can tell, pin-depends is broken before this PR in the case of pins to git tags, since they would be cached to the first reference the tag was resolved to.

talex5:
Yeah, my plan is to finish and upstream the changes to the solver libraries, then merge this, then investigate how to fix pin-depends.

lib/opam_build.ml
```diff
@@ -68,15 +70,14 @@ let install_project_deps ~base ~opam_files ~variant ~for_user =
   distro_extras @@
   workdir "/src" @@
   run "sudo chown opam /src" @@
+  run "cd ~/opam-repository && (git reset --hard %s || (git pull origin master && git reset --hard %s)) && opam update" commit commit @@
```
avsm, line 73:
```suggestion
  run "cd ~/opam-repository && (git reset --hard %s || (git pull origin master && git reset --hard %s)) && opam update -u" commit commit @@
```

Useful to have the update step here in case there has been a change to the compiler description files at this stage of the build.

PR #168, opened 2020-04-30
duniverse build: update to latest versions of tools, remove odoc

avsm:
odoc was in the class of 'this can never work', since of course
the tool depends on reading cmt files and therefore is dependent
on the precise compiler version.

This PR updates ocamlformat, duniverse and dune to the latest
versions as well. The next step in the duniverse tool is to do
a complete (and cache-friendly) bootstrap of ocaml and tools
from within `duniverse install`, so we should be able to remove
the hardcoded versions in the CI soon.


avsm:
needs #169 to work correctly.

PR #183, opened 2020-05-17
Add support of OCaml 4.10+32bit on Debian

### LINKED ISSUE #182: Add OCaml+32bit
In some case, a test with OCaml +32bit is useful, specially for:
- [`optint`](https://github.com/mirage/optint) which the compilation differs between 32bit/64bit
- `mirage-crypto` which use some specials instructions (eg. mirage/mirage-crypto#60

More globally, you should be able to test our libraries to ensure a support for arm32/esp32. So, I don't have any clues to add it.

dinosaure:
Want to fix #182, 4.10 seems enough I think - however, a docker image is missing.

### DIFF
--- a/service/conf.ml
+++ b/service/conf.ml
@@ -69,6 +69,8 @@ let platforms =
       v "centos"   Builders.amd3 "centos-8"      default_compiler;
       v "fedora"   Builders.amd3 "fedora-31"     default_compiler;
       (* oraclelinux doesn't work in opam 2 yet *)
+      (* Variants *)
+      v "4.10+32bit" Builders.amd4 "debian-10" "4.10+32bit";
     ]
   | `Dev ->
     [
--- a/service/conf.ml
+++ b/service/conf.ml
@@ -70,7 +70,7 @@ let platforms =
       v "fedora"   Builders.amd3 "fedora-31"     default_compiler;
       (* oraclelinux doesn't work in opam 2 yet *)
       (* Variants *)
-      v "4.10+32bit" Builders.amd4 "debian-10" "4.10+32bit";
+      v (default_compiler ^ "+32bit") Builders.amd4 "debian-10" (default_compiler ^ "+32bit");
     ]
   | `Dev ->
     [


service/conf.ml
```diff
@@ -69,6 +69,8 @@ let platforms =
       v "centos"   Builders.amd3 "centos-8"      default_compiler;
       v "fedora"   Builders.amd3 "fedora-31"     default_compiler;
       (* oraclelinux doesn't work in opam 2 yet *)
+      (* Variants *)
+      v "4.10+32bit" Builders.amd4 "debian-10" "4.10+32bit";
```
CraigFe, line 73:
```suggestion
      v (default_compiler ^ "+32bit") Builders.amd4 "debian-10" (default_compiler ^ "+32bit");
```

At some point, `default_compiler` will probably be determined dynamically to be the highest OCaml version supported by the project. I think we might as well avoid hard-coding `4.10` here.

talex5:
This seems reasonable to me. I guess you'll need to add the extra variant here:

https://github.com/ocurrent/ocaml-version/blob/master/ocaml_version.ml#L231

avsm:
I don't think this is quite right.  Why not just create a base i386 image from https://hub.docker.com/r/i386/debian/ and just do a normal OCaml compiler compilation into that? That will work with depexts as well without any multilib complexity.

dinosaure:
> I don't think this is quite right. Why not just create a base i386 image from https://hub.docker.com/r/i386/debian/ and just do a normal OCaml compiler compilation into that? That will work with depexts as well without any multilib complexity.

I agree with you but I don't have any clue to add such image and integrate it into `ocaml-ci`.

talex5:
The pipeline that builds the base images is here:

https://github.com/ocurrent/docker-base-images/blob/f4282e58ac2eec50c4e85423122b4c8c3f5ce965/src/pipeline.ml#L132

The set of architectures to build for is:

https://github.com/ocurrent/docker-base-images/blob/f4282e58ac2eec50c4e85423122b4c8c3f5ce965/src/conf.ml#L85

```ocaml
let arches_for ~distro = Dockerfile_distro.distro_arches Ocaml_version.Releases.latest distro
```

`Dockerfile_distro.distro_arches` is:

https://github.com/avsm/ocaml-dockerfile/blob/d94e531638d03397b8c2adb8d62da9b0aa10fe31/src-opam/dockerfile_distro.ml#L91-L96

So maybe you could add ``` `X86``` as another option here?

dinosaure:
I did:
- ocurrent/ocaml-version#12
- ocurrent/docker-base-images#40

Where I added `i386` architecture. Then, I don't know what I need to do.

avsm:
Closing this in favour of #219

PR #189, opened 2020-05-22
Use git-unix to read opam-repository

talex5:
Instead of checking out the whole of opam-repository on each analysis run, just pass the hash to the solver. The solver maintains a bare opam-repository clone and reads it directly.

We also parse all the opam files in opam-repository and keep them in memory until the head commit changes, which should be faster when doing lots of solves.


talex5:
This handled an update to opam-repository overnight. Instead of the usual 80 minutes to recalculate, it only took 8 minutes this time :-)

PR #155, opened 2020-04-23
Web-ui improvements

kit-ty-kate:
I did split this PR into three separate commits:
* the first one just fixes some issues with the status separator for hierarchical statuses
* the second one improves the failure icon
* the third is an update to #127 noted in https://github.com/ocurrent/ocaml-ci/pull/127#issuecomment-593721456. However my code still hasn't changed here and even though I'm still slightly unhappy with it, I didn't experienced any issues with it in opam-ci, so I prefer to open a PR now and get some feedbacks on it or push it now and improve it further later (I'm fine with both)

Feel free to cherry-pick the two first commits separately if you prefer.

Here is a screenshot of the result of this PR overall:
Screenshot from 2020-04-23 20-24-22

PR #192, opened 2020-05-28
[META-PR] opam-ci (CI for opam-repository) integation

kit-ty-kate:
Here is a draft META-PR listing all things that need to be done to integrate opam-ci into ocaml-ci. The current opam-ci works as-is but has several issues, however it also bring new things that can also improve ocaml-ci alone.

The PR is not meant to be merge as-is, but rather a place to discuss, list what needs to be integrated (bit by bit) in order to integrate the final pipeline. The final change once everything is integrate should be rather small, and I'll do my best to keep the branch up-to-date when an item of the list is integrated in a separate PR into ocaml-ci.

TODOs for opam-ci integration into ocaml-ci:
* [ ] Create an archive cache service / Remove buildkit
* [ ] Add a way to get the revdeps in the solver (opam-ci currently uses `opam list`)
* [ ] Make the client API return a tree-like structure for statuses. This allows easier `citty` support. (opam-ci currently reconstructs the tree in the web-ui service, see https://github.com/ocurrent/ocaml-ci/pull/155)
* [ ] Check the merge commit for PRs instead of the branch HEAD (opam-ci currently merges by hand in each `docker build`)
* [ ] Dealing with failed states better (see https://github.com/ocurrent/ocurrent/issues/181)


talex5:
Moved to https://github.com/ocurrent/opam-repo-ci/issues/49.

PR #198, opened 2020-06-06
duniverse: do not pin the tool

avsm:
The ocamlformat failure on the live CI is because the opam metadata
in the release branch is wrong. Since these tools are all guaranteed
to be in the opam repo, simply do not pin the tool while building
the binary.

This should fix the live CI which is failing with ocamlformat
trying to use base.v0.14.0 and thus not building.

### DIFF
--- a/lib/duniverse_build.ml
+++ b/lib/duniverse_build.ml
@@ -23,7 +23,6 @@ let install_bin ~compiler ~repo ~tag ~bins name =
     workdir "/home/opam/src" @@
     run "echo \"(lang dune 2.0)\" > dune-workspace" @@
     run "echo \"(env (_ (flags -cclib -static)))\" >> dune-workspace" @@
-    run "opam pin add -n ." @@
     run "opam depext %s" name @@
     run "opam install --deps-only %s" name @@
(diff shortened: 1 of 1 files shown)


talex5:
Uh-oh:

```
Fatal error: exception (Failure
  "Failed to read tree: Chunk loaded from opam-repository/objects/17/2d5d5f5193d5a1a407466bceba45eea20121ed is too big (32768 bytes) for the decoder (max: 32768 bytes)")
```

Looks like https://github.com/mirage/irmin/issues/1001 has hit the production CI too :-(

talex5:
With this change, RWO master fails with:
```
#17 [duniverse_binary 10/13] RUN opam depext duniverse
#17 sha256:42b7f83154ab60ca399481853ec1c7f3602a4e55807bc79d526f5e68849c9eab
#17 4.243 [ERROR] No solution for duniverse: Your request can't be satisfied:
#17 4.243           - unknown package
```

avsm:
Sorry about the half working PR. I've been shifting this functionality into `duniverse` itself, so that it does a full tool bootstrap. I'll have a fixed PR soon that removes all this adhoc build logic in the CI.

PR #205, opened 2020-06-09
Report statistics about the state of all master branches

talex5:
Reporting states for all PRs and branches isn't very useful because they may be stale, but the master branchs should be a good indicator of things that are supposed to be passing.

Also, we now track the set of active repositories. Before, we listed all repositories for which there were entries in the database, even if they are no longer being monitored. Now, we just list the ones that are still active.

### DIFF
--- a/api/client.ml
+++ b/api/client.ml
@@ -21,6 +21,18 @@ module State = struct
     | Undefined x -> Fmt.pf f "unknown:%d" x
 end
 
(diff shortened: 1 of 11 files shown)


talex5:
This now also updates the web UI a bit:

- When listing refs, show an indicator of the status.
- When listing repositories, show the status of the master branch.

PR #210, opened 2020-06-21
Add @favonia to the whitelist

favonia:


avsm:
Hi @favonia - this service is still iterating rapidly, so we just want to check what you're using it for and what your expectations are.

favonia:
I plan to use it for my own OCaml projects. (Is there something else the service is providing?) It seems this service does not suffer from many GitHub Actions problems, and that's why I am interested. I know this is experimental, and I can always switch back to GitHub Actions or Travis if there's an issue. Does this answer your question?

favonia:
Hi, I am a bit confused by the current state of the PR. Is there something I should do to get into the whitelist, or is the service not ready for public usage? I am okay with the latter but would appreciate an answer so that I can move on to finding other resources. Thank you.

avsm:
@favonia, I'll approve this one on the understanding that this is not yet a public service. We're iterating fairly rapidly on it to replace the existing opam-repository CI, so there will be some instability. In return, we'd appreciate any feedback (or indeed, PRs) about how it's working for you and any limitations you observe -- issues on this repository are fine.

favonia:
Thank you! Yes, I will report any problem I encounter.

PR #217, opened 2020-07-08
Use a project name that is a package name

emillon:
Otherwise `dune subst` (invoked when pinning) fails.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -1,5 +1,5 @@
 (lang dune 2.0)
-(name ocaml-ci)
+(name ocaml-ci-api)
 (formatting disabled)
 
 (generate_opam_files true)


talex5:
This seems wrong - the project isn't called "ocaml-ci-api".

I don't understand how `dune subst` is supposed to work - the repository contains multiple packages so there is no package name for the whole repository. Can we disable it somehow?

emillon:
I think that's a dune limitation at the moment - the project name needs to be one of its package names. If watermark is not used, I don't think that hurts, but dune subst also sets a version number based on git, so disabling it would not do the right thing either.

emillon:
We can also make a dummy package ocaml-ci that depends on all packages in the repo maybe?

talex5:
OK, let's merge it until dune is fixed then.

PR #219, opened 2020-07-14
support 32-bit and multiarch builds

### LINKED ISSUE #182: Add OCaml+32bit

avsm:
This plumbs through the arch option in order to allow the selection of 32bit builds.  Its only activated for i386 right now, but when we use the builder pool it should also allow for selecting arm32 as well.

Only lightly tested so far. Depends on ocurrent/ocurrent#213

Fixes #182

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,6 +4,9 @@
 [submodule "opam-0install-solver"]
 	path = opam-0install-solver
 	url = https://github.com/talex5/opam-0install-solver
+[submodule "ocaml-dockerfile"]
+	path = ocaml-dockerfile
+	url = https://github.com/avsm/ocaml-dockerfile.git
 [submodule "ocluster"]
 	path = ocluster
 	url = https://github.com/ocurrent/ocluster.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,9 +4,6 @@
 [submodule "opam-0install-solver"]
 	path = opam-0install-solver
 	url = https://github.com/talex5/opam-0install-solver
-[submodule "ocaml-dockerfile"]
-	path = ocaml-dockerfile
-	url = https://github.com/avsm/ocaml-dockerfile.git
 [submodule "ocluster"]
 	path = ocluster
 	url = https://github.com/ocurrent/ocluster.git
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM ocurrent/opam:debian-10-ocaml-4.10 AS build
 RUN sudo apt-get update && sudo apt-get install libev-dev capnproto m4 pkg-config libsqlite3-dev libgmp-dev -y --no-install-recommends
-RUN cd ~/opam-repository && git pull origin -q master && git reset --hard a5e373ef1d13748cb092ede3c5b74ce6b6c03349 && opam update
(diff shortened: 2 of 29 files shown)


.gitmodules
```diff
@@ -1,6 +1,12 @@
...
+	url = https://github.com/avsm/ocurrent.git
 [submodule "opam-0install-solver"]
 	path = opam-0install-solver
 	url = https://github.com/talex5/opam-0install-solver
+[submodule "ocaml-dockerfile"]
+	path = ocaml-dockerfile
+	url = https://github.com/avsm/ocaml-dockerfile
+[submodule "ocaml-version"]
+	path = ocaml-version
+	url = https://github.com/ocurrent/ocaml-version
```
avsm, line 12: This will be ocaml-version.2.6.1 as soon as https://github.com/ocaml/opam-repository/pull/16822 is merged

avsm:
merged against master to work with the cluster

lib/platform.mli
```diff
@@ -13,16 +13,19 @@ val pp : t Fmt.t
 val compare : t -> t -> int
 
 val get :
+  arch:Ocaml_version.arch option ->
   label:string ->
   builder:Builder.t ->
   pool:string ->
   distro:string ->
   ocaml_version:string ->
+  Current_docker.Raw.Image.t Current.t *
```
talex5, line 22: Would be clearer to use a labelled argument here I think. I try to avoid having two unnamed arguments with same type and no obvious order.

Dockerfile
```diff
@@ -39,4 +39,5 @@ RUN apt-get update && apt-get install docker-ce -y --no-install-recommends
 WORKDIR /var/lib/ocurrent
 ENTRYPOINT ["dumb-init", "/usr/local/bin/ocaml-ci-service"]
 ENV OCAMLRUNPARAM=a=2
+ENV DOCKER_CLI_EXPERIMENTAL=enabled
```
talex5, line 42: What's this for?

avsm: The image now does docker manifest pull to calculate opam vars (before it uses the pool), so the ocaml-ci image needs it enabled too.  I've added a comment to the Dockerfile in 2fd97826a0022d24cb3b85943f64354f1eb6a598

lib/platform.ml
```diff
@@ -90,29 +93,31 @@ module Query = struct
...
   let latched = true
 end
 
 module QC = Current_cache.Generic(Query)
 
-let query builder ~variant image =
+let query builder ~variant (host_image, image) =
   Current.component "opam-vars" |>
-  let> image = image in
+  let> host_image, image = Current.pair host_image image in
```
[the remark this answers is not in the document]
avsm: done in c871088a8137d1f2c5d764172669497e53e14f59

lib/platform.mli
avsm: done in c871088a8137d1f2c5d764172669497e53e14f59

PR #229, opened 2020-07-26
opam_build: check that the source tree is unmodified post build

### LINKED ISSUE #220: CI check suggestion: dune-project and opam files are in sync
some projects generate their opam file from dune-project, but keep the opam files in the git repository for technical reasons (to keep `opam pin --dev` working).

the generated opam file contain a comment "please do not edit, generated by ..." - that is fine. but already more than once it happened to me that I modify (out of habit) the opam file, get the PR merged, and discover only at a later point that the dune-project and opam are now out of sync.

it would be great if the ocaml CI could check that after a `dune build` the `git diff` is empty (i.e. there are no modified files). thanks.

avsm:
This will catch the case of modifying a dune-project file but
forgetting to commit the corresponding opam file (or vice versa).

Fixes #220


avsm:
The source tree is not a git repository any more, so the easy way doesn't work.

PR #230, opened 2020-07-26
add an opam lint checking phase

### LINKED ISSUE #151: Opam files are not checked during lint stage
The CI currently does not fail or notice when an invalid opam file (w.r.t. `opam lint`) is used in the repo, e.g. when there is no synopsis/description. 
Maybe that could be checked during the `lint` stage?

avsm:
This runs opam lint on all the detected opam files. It should
perhaps only run them on the toplevel ones, but its useful to
know if any sub-files are also wrong I think.

I considered making the check only copy the opam files, but
there may be other dependencies required from the source
tree (via extra-files), so its best to just take the whole
source tree over. It's very fast to recompute this lint, so
rebuilds shouldn't be too much worse.

Fixes #151


avsm:
Pushed to live and seems to be working fine.

PR #232, opened 2020-07-27
duniverse: use opam-tools plugin to get environment

avsm:
The duniverse build previously hardcoded the set of tools to
build, which is obviously problematic when they get out of sync
(most notably for ocamlformat).  This now shifts the CI to using
the soon-to-be-released opam-tools plugin to get those tools
installed in a local switch.

I'll be refining this more, but this immediately unbreaks ocaml-ci
for realworldocaml/book, where the ocamlformat build has failed
for some time due to the old way of building it here.

### DIFF
--- a/lib/duniverse_build.ml
+++ b/lib/duniverse_build.ml
@@ -1,52 +1,3 @@
-let latest_patch_release_with_flambda compiler =
-  let open Ocaml_version in
-  of_string_exn compiler |> fun ov ->
-  Releases.all_patches |>
-  List.filter (fun ov' -> with_just_major_and_minor ov' = ov) |>
-  List.sort compare |> List.rev |> function
-  | [] -> failwith ("Unknown compiler release" ^ compiler)
-  | hd::_ -> Opam.V2.name @@ with_variant hd (Some "musl+flambda") (* TODO add `Musl to ocaml-version *)
-
-let install_bin ~compiler ~repo ~tag ~bins name =
-  let open Dockerfile in
-  let alias = Printf.sprintf "%s_binary" name in
-  let base = Printf.sprintf "ocurrent/opam:debian-10-ocaml-%s" compiler in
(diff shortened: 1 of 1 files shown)


avsm:
pushed to live to test on rwo

lib/duniverse_build.ml
```diff
@@ -65,24 +16,30 @@ let build_cache repo =
 
 let download_cache = "--mount=type=cache,target=/home/opam/.opam/download-cache,uid=1000"
 
+let install_opam_tools =
+  let opam_tools_hash = "6c56ab9fedd7b3f6c143cb606a0ea6fe6a384013" in
+  let open Dockerfile in
+     run "opam pin add -n https://github.com/avsm/opam-tools.git#%s" opam_tools_hash
+  @@ run "opam depext -iy opam-tools"
```
kit-ty-kate, line 23: wouldn't that come with a risk of interference between packages installed in opam and packages install via the duniverse (I'm not too sure how they interact).

I'd suggest to add the following two lines:
```
@@ "cp \"$(opam exec -- sh -c 'command -v opam-tools')\" ~/"
@@ "run opam remove -a opam-tools"
```
then later use the copied binary `~/opam-tools` directly instead of through opam.

avsm: It does indeed run that risk, but I'm modifying this in opam-tools to not actually have any dependencies.

Note that the contamination of packages isn't actually a problem for a well-formed duniverse though, since there will never be any use of the opam installed libraries, since _all_ packages are resolved from within the duniverse. This is guaranteed by the duniverse pull.

However, it's still good practise to keep the opam environment as clean as possible. I'll have a few more PRs to refine this behaviour as I continue iterating on opam-tools.

avsm:
Merging this, will have the standalone opam-tools in a followup PR.

PR #234, opened 2020-07-29
Add jeffa5 to the whitelist

jeffa5:


avsm:
This CI service is still under development; could you let us know roughly what you're planning to do with it @jeffa5?  Primarily, I want to set the expectation that it might break frequently as we're iterating through changes towards a stable release.

jeffa5:
Sure, I just wanted to explore it and try it for some of my personal projects. Breaking won't be a problem :)

PR #235, opened 2020-07-31
add bikallem to allowlist

bikallem:


avsm:
This CI service is still under development; could you let us know roughly what you're planning to do with it @bikallem? Primarily, I want to set the expectation that it might break frequently as we're iterating through changes towards a stable release.

bikallem:
Mostly curious if it will be more convenient, faster than github workflows.

PR #236, opened 2020-08-01
Add support for building OCaml compiler repositories and variants

avsm:
This adds support for parsing compiler variants (e.g. flambda) in
the Variant module, which in turns allows for the compilation of
OCaml compilers.

OCaml compiler repositories are detected (using the experimental
dune flag that signifies that the compiler dune rules are active).
We then use the base distros and variants defined for the trunk
compiler (in service/conf.ml) to generate Dockerfiles that run
the full test suite.

In order to prevent an explosion of tests for non-OCaml compilers
(i.e. normal repositories), we filter out all the non-standard
variants before solving for normal repositories.  We could bring
in opt-in support for (e.g. flambda or nnp) by looking for x-
fields in their opam files, but this is not implemented yet.

### DIFF
--- a/lib/analyse.ml
+++ b/lib/analyse.ml
@@ -124,7 +124,7 @@ module Analysis = struct
       List.filter_map (fun (variant,_) ->
         let ov1 = OV.with_just_major_and_minor OV.Sources.trunk in
         let ov2 = OV.without_variant (Variant.ocaml_version variant) in
-        if OV.compare ov1 ov2 = 0 then Some variant else None) platforms in
+        if OV.equal ov1 ov2 then Some variant else None) platforms in
     Lwt_result.return (`Ocaml_compiler variants)
 
   let check_opam_version =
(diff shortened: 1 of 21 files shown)


lib/analyse.ml
```diff
@@ -236,8 +274,11 @@ module Analysis = struct
     else if List.filter is_toplevel opam_files = [] then Lwt_result.fail (`Msg "No top-level opam files found!")
     else (
       begin
-        if is_duniverse then duniverse_selections ~job ~platforms dir
-        else opam_selections ~solver ~job ~platforms ~opam_repository_commit ~opam_files dir
+        match ty with
+        | `Duniverse -> duniverse_selections ~job ~platforms dir
+        | `Ocaml_repo -> opam_selections ~solver ~job ~platforms ~opam_repository_commit ~opam_files dir
+        | `Ocaml_compiler -> ocaml_compiler_selections ~platforms
+        | `Opam | `Dune -> failwith "TODO"
```
avsm, line 281: Going to test out the OCaml compiler support before adding in runes for Opam and Dune (both of which have special bootstrap requirements too)

avsm:
This works pretty well, and found a few bugs in running the testsuite on different architectures: https://ci.ocamllabs.io/github/avsm/ocaml/commit/8b6241c64c8e02a02daa7c4aa6c2181238023834

The only hitch is that the vast number of branches in ocaml/ocaml means that we run into the paging limit, so we need to do some branch filtering before it can be activated on the live ocaml repo. Also a lot of mysterious BuildKit failures but for ppc64le mainly

avsm:
One possible optimisation for building the compiler Dockerfile more incrementally is to copy everything but `testsuite/` and then then copy that. It's difficult to copy everything-but-one-directory with COPY though

avsm:
I'm inclined not to merge this right now. The divergence in the usage of platforms is pretty significant, so perhaps ocaml-compiler-ci should be its own service rather than trying to shoe horn it into the package CI.

avsm:
Closing in favour of #241; OCaml compiler support will reappear in the future.

PR #238, opened 2020-08-07
Use "git fetch", not "git pull", to update opam-repository

talex5:
We're going to do a `git reset --hard` in the next step, so updating the working tree in the pull step is a waste of time and produces lots of pointless log output.

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -74,7 +74,7 @@ let install_project_deps ~base ~opam_files ~selection ~for_user =
   distro_extras @@
   workdir "/src" @@
   run "sudo chown opam /src" @@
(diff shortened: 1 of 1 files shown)


avsm:
Lgtm. This will cause a large rebuild - I need to update the base images (to add an opam-2.1 binary into the base images for future use) so i can merge this with #236 and push live ...

talex5:
It shouldn't cause a rebuild. The build (and cluster-build) operations are keyed using the opam-repository commit, but the exact content of the Dockerfile is generated only in the `run` function.

avsm:
Sounds good. I'll merge this and push it live shortly once the cluster settles.

PR #240, opened 2020-08-10
Update OCurrent to remove limit on number of installations

talex5:


avsm:
merging this, backing out #236 and pushing the result live

PR #241, opened 2020-08-11
Add support for building OCaml variant versions

avsm:
This adds support for parsing compiler variants (e.g. flambda) in
the Variant module.

The arch value is now explicitly specified, so the build modules can
choose whether to introduce a personality or not in their
Dockerfiles.

The configuration now uses Ocaml_version to setup the various
architecture choices.  This just adds flambda to the set of
tested defaults, using the latest stable compiler version.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,2 +1,2 @@
-version = 0.12
+version = 0.15.0
 disable = true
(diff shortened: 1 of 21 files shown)


avsm:
This is the "good bits" of #236 to support compiler variants like flambda

avsm:
I'll test this on live later on this afternoon. It shouldn't change much externally except flambda being added to the test matrix.

avsm:
Pushing this live as a test

talex5 (approved):

avsm:
I'll push this live again now.

talex5 (changes requested):
This PR reverted some changes in OCurrent. That should be fixed.

avsm:
It should be on ocurrent master now I think

talex5:
Ah, I see the problem: https://github.com/ocurrent/ocurrent/pull/219 hadn't been merged, so updating to master reverted that.

avsm:
Ok merged it in and also cherry-picked d787bd0176409044dc2e29bbb8519224fcfb8c3e from #245 so the history is clean now

avsm:
Pushed this branch to live-engine

talex5 (approved):

PR #243, opened 2020-08-24
Use per-pool rate limits

talex5:
Otherwise, we might limit x86 builds because we're doing lots of arm builds, which isn't helpful.


avsm (approved):
lgtm

PR #244, opened 2020-09-03
Handle 'post' dependencies

Julow:
Post dependencies must be installed after other dependencies and are often used to break dependency cycles.

This patch only detects 'post' dependencies that are directly required by the root packages.
Installing the released version of current package is now allowed.

Because of this, the pinning of root packages must be done after installing dependencies.
We must now handle pin-depends from root packages explicitly (it was done implicitly by `opam`).

This is very adhoc but it seems to be working as expected.

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -56,6 +56,9 @@ let install_project_deps ~base ~opam_files ~selection ~for_user =
   let root_pkgs = get_root_opam_packages groups in
   let non_root_pkgs = packages |> List.filter (fun pkg -> not (List.mem pkg root_pkgs)) in
   let download_cache_prefix = if for_user then "" else download_cache ^ " " in
+  (* Remove root packages from post packages. This can happen when a package
+     has a 'post' dependency on an other root package. *)
+  let post_packages = List.filter (fun p -> not (List.mem p root_pkgs)) post_packages in
   let open Dockerfile in
   let distro_extras =
     if Astring.String.is_prefix ~affix:"fedora" (Variant.id variant) then
@@ -79,7 +82,7 @@ let install_project_deps ~base ~opam_files ~selection ~for_user =
   run "sudo chown opam /src" @@
   run "cd ~/opam-repository && (git cat-file -e %s || git fetch origin master) && git reset -q --hard %s && git log --no-decorate -n1 --oneline && opam update -u" commit commit @@
   env ["DEPS", String.concat " " non_root_pkgs] @@
-  env ["POST_DEPS", String.concat " " post_packages] @@
+  env ["POST_DEPS", match post_packages with [] -> "\"\"" | p -> String.concat " " p] @@
   run "%sopam depext --update -y $DEPS $POST_DEPS" download_cache_prefix @@
   run "%sopam install $DEPS" download_cache_prefix @@
   install_post_deps @@
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -51,7 +51,7 @@ let rec get_root_opam_packages = function
 let download_cache = "--mount=type=cache,target=/home/opam/.opam/download-cache,uid=1000"
 
 let install_project_deps ~base ~opam_files ~selection ~for_user =
-  let { Selection.packages; commit; variant } = selection in
+  let { Selection.packages; post_packages; commit; variant } = selection in
   let groups = group_opam_files opam_files in
   let root_pkgs = get_root_opam_packages groups in
   let non_root_pkgs = packages |> List.filter (fun pkg -> not (List.mem pkg root_pkgs)) in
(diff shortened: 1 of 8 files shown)


avsm:
As context, this is needed by odoc.

lib/opam_build.ml
```diff
@@ -51,18 +58,27 @@ let rec get_root_opam_packages = function
...
 let install_project_deps ~base ~opam_files ~selection ~for_user =
-  let { Selection.packages; commit; variant } = selection in
+  let { Selection.packages; post_packages; commit; variant; pin_depends } = selection in
   let groups = group_opam_files opam_files in
   let root_pkgs = get_root_opam_packages groups in
   let non_root_pkgs = packages |> List.filter (fun pkg -> not (List.mem pkg root_pkgs)) in
   let download_cache_prefix = if for_user then "" else download_cache ^ " " in
+  (* Remove root packages from post packages. This can happen when a package
+     has a 'post' dependency on an other root package. *)
+  let post_packages = List.filter (fun p -> not (List.mem p root_pkgs)) post_packages in
```
Julow, line 68: This case happens on Irmin, `irmin` requires `irmin-mem.dev` but `irmin-mem.dev` was pinned to an empty source tree, Opam seems to do nothing. We need to filter them because now this case will be an error.

talex5 (commented):
This is a quite a big patch and I don't understand what it's doing. Could you say more about what the problem is, with an example of something that doesn't work, and explain the changes you're making?

I would expect that someone checking out a project to work on it locally would be able to do `opam install --deps-only -t .` and then `dune build`. It sounds like things can be more complicated than that... is it something we can fix in opam?

Julow:
Odoc depends on Mdx (`with-test`) which depends on Odoc. With the `post` flag, `opam install -t .` works as expected but `--deps-only` would not install Mdx.

With this changes, the released version of Odoc then Mdx will be installed. This is not equivalent to `opam install -t .` but is more cache-friendly and avoid unrelated errors in building Mdx with the dev version of Odoc.

This is how I setup my local environment for Odoc: `opam install --deps-only -t .`, Mdx is not installed, `opam install mdx`, the released version of Odoc and Mdx are installed.

talex5:
Thanks for the example. I see there was some discussion on the #opam slack channel too on the 11th Sep. That seemed to conclude with:

> if pinning works this looks like a better solution, compared to adding a post dependency in opam-repository

Does that affect this PR?

In general, it seems concerning that the CI can now install a different version of the package being tested. e.g. a passing test for odoc only says that the *previous* version of odoc works with mdx, not the one under test.

But if this is still needed, perhaps the refactoring parts could be moved to a separate PR so they can be reviewed and merged first? e.g. it looks like this PR switches to using JSON for communicating solver-worker results. And there's a new `No_solution` in the solver API. Is that related to the post-dependencies problem?

Julow:
> > if pinning works this looks like a better solution, compared to adding a post dependency in opam-repository
>
> Does that affect this PR?

Possibly. There is an other solution that I'm investigating: Install the released version of the current package, if required.
This might not work as easily on Irmin, which has `post` dependencies on root packages, which this PR handles.

This PR is doing that too but only for dependencies marked `post` to avoid changing the way the CI works for other packages.
In Odoc's case, the `post` flag is working as expected in the case of `opam install -t`, but the CI is doing the equivalent of `--deps-only -t`, which in Opam, doesn't work with the `post` flag (deps marked `post` are not installed, meaning the tests cannot run).
The behavior added in this PR doesn't exist in Opam and is only what I expect, in the case of Odoc.

I really don't want to build Mdx with the version of Odoc that is currently being tested. This would break caching but more importantly, would make the CI unusable in case of breaking changes.

I had to use a richer type to pass the additional field. The `No_solution` case is because calling code cannot just test if the list is empty because there is now 2 lists. I'll split PRs if this solution is still useful.

talex5:
Summary of slack discussion:

- The CI wants to install (and cache) all dependencies before building and testing.
- It can't do that if it has to download the mdx dependency after building odoc.
- It might be possible to build the new version of odoc using an mdx built with the previous version in some cases, but not if they have incompatible dependencies. Also, it would be useful to know that odoc changes don't break mdx (otherwise it will only work in CI, but not once released).
- odoc wants to use mdx because the OCaml top-level printer is useful in its unit tests.
- A simple short-term fix is to use git-submodule to add mdx to the odoc repository. Then the CI will treat odoc and mdx as a single project and there won't be a problem. This is similar to the case for Irmin (a cyclic dependency within the repository, where one edge is a test edge), and that works OK.
- A longer-term solution is to move the odoc parser (which is the bit mdx needs) to a separate repository. In fact, it was imported originally from [octavius](https://github.com/ocaml-doc/octavius), so it could just go back there.

Julow:
Thanks for the help !

PR #245, opened 2020-09-04
Adjust arch opam variable for other architectures 

talex5:
At the moment, we only query the opam variables on `x86_64` and assume they are the same on other platforms. But we need to override the `arch` value at least!


avsm:
Ahh this explains why we get occasional solver failures from the opam invocation itself.

talex5:
Now included in #241.

PR #248, opened 2020-09-13
Add joelburget to alpha testers list.

joelburget:


talex5:
Note: we're iterating fairly rapidly on the CI, so there will be some instability. In return, we'd appreciate any feedback (or indeed, PRs) about how it's working for you and any limitations you observe -- issues on this repository are fine.

PR #249, opened 2020-09-18
Add jonludlam to stack.yml

jonludlam:
Please! :-)


talex5:
Done.

Note: you seem to have a lot of branches. If they don't all need to be kept up-to-date, consider deleting them or converting them to tags so the CI doesn't have to keep checking them.

(we should probably filter them out, but at the moment the CI will keep the status of all branches fresh)

Issue #250, opened 2020-09-24
Documentation linting job does not include the source code


CraigFe:
It appears that the `lint-doc` Dockerfile does not include the source code of the project when running `dune build @fmt` when triggered from the GitHub pipeline.

Running `dune build @doc` inside a directory that is not a Dune project (i.e. contains no `dune-project` files in the ancestry of the root nor `dune` files in its descendents) will succeed, leaving an empty `_build` directory:

```bash
❭mkdir /tmp/doc && cd /tmp/doc && dune build @doc
Done: 0/0 (jobs: 0) •
```

This has the effect of creating false-positives on all project with `odoc` dependencies and a false-negative for the Odoc repository itself (since the analysis phase correctly determines that `odoc` does not need to be installed from Opam, then tries to build the documentation inside an empty tree.)

The fix is to copy the project source tree into the current directory before performing `dune build @doc`. Thanks to @jonludlam for the report.

PR #251, opened 2020-09-24
lint-doc: copy project source into container before building docs

CraigFe:
Fix https://github.com/ocurrent/ocaml-ci/issues/250. The implementation of `Lint.doc_dockerfile` is now almost identical to that of [`Opam_build.dockerfile`](https://github.com/ocurrent/ocaml-ci/blob/655edd03d94c44874bf1561efe8429ab1d2dab53/lib/opam_build.ml#L80), as it should be.

Also fixes an escaping issue in the use of "odoc>=1.5.0" which was causing the creation of a file named "=1.5.0" inside the lint container.


talex5 (approved):
I tried it on your `a-doc-test` and it found the error :-)

CraigFe:
`live-engine` is currently on this branch.

PR #253, opened 2020-10-07
Generate Dockerfiles from OBuilder specs

talex5:
Instead of generating Dockerfiles directly, generate OBuilder specs and then convert them to Dockerfiles.

This lets us test that the conversion works, before enabling OBuilder builds on the main cluster.


avsm:
Looks good to me.

PR #258, opened 2020-10-16
Work around opam depext bug

talex5:
https://github.com/ocaml-opam/opam-depext/pull/132


emillon:
FYI, there's now a fixed version of depext on opam-repository: https://github.com/ocaml/opam-repository/pull/17435

talex5:
Thanks! I'll leave it as it is for now, as it will take a while for the base images to update.

PR #259, opened 2020-10-16
Pull the list of distributions to test from ocaml-dockerfile

kit-ty-kate:
This PR is a proposal to integrate ocaml-ci more closely with ocaml-dockerfile.
Currently every CIs have each their own list of distributions and I feel this could be better done by sharing it via ocaml-dockerfile which would get updated only once when a new version of a distribution comes out.

This PR requires https://github.com/avsm/ocaml-dockerfile/pull/25 in which I'm proposing a new Tier system which reflects more the state of opam-repository.

### DIFF
--- a/service/conf.ml
+++ b/service/conf.ml
@@ -61,27 +61,27 @@ let platforms =
   let v ?(arch=`X86_64) label distro ocaml_version =
     { arch; label; builder = Builders.local; pool = pool_of_arch arch; distro; ocaml_version }
   in
+  let master_distro = DD.resolve_alias DD.master_distro in
   let make_distro distro =
+    let distro = DD.resolve_alias distro in
     let label = DD.latest_tag_of_distro distro in
     let tag = DD.tag_of_distro distro in
     let ov = OV.(Releases.latest |> with_just_major_and_minor) in
-    if distro = `Debian `V10 then
+    if distro = master_distro then
       v label tag (OV.with_variant ov (Some "flambda")) ::
       List.map (fun arch -> v ~arch label tag ov) (DD.distro_arches ov distro)
     else
       [v label tag ov]
   in
   let make_release ?arch ov =
-    let distro = DD.tag_of_distro (`Debian `V10) in
+    let distro = DD.tag_of_distro master_distro in
     let ov = OV.with_just_major_and_minor ov in
     v ?arch (OV.to_string ov) distro ov in
   match profile with
   | `Production ->
-      let distros = List.map make_distro [
-        `Debian `V10; `Alpine `V3_12; `Ubuntu `V20_04;
-        `Ubuntu `V18_04; `OpenSUSE `V15_2; `CentOS `V8;
-        `Fedora `V32 ] |> List.flatten
-      in
+      let distros =
+        DD.active_tier1_distros `X86_64 @ DD.active_tier2_distros `X86_64 |>
+        List.map make_distro |> List.flatten in
       (* The first one in this list is used for lint actions *)
       let ovs = List.rev OV.Releases.recent @ OV.Releases.unreleased_betas in
       List.map make_release ovs @ distros


talex5:
This is failing CI because it has a `pin-depends` on a package that depends on ocaml-dockerfile. Rebasing on master (which uses submodules instead) should fix it.

avsm:
Pushed a merge and to use the master branch of dockerfile-opam

talex5 (changes requested):
It seems that with this change, we'll only test on Debian and Alpine. Are you sure that's what we want?

talex5:
I also rebased and squashed; the downgrades and upgrades of the submodules were confusing me ;-)

avsm:
That's not right indeed. It should be using latest_distros and not tier1_distros.

avsm:
Take 3 pushed, @talex5 :-) See https://github.com/avsm/ocaml-dockerfile/pull/30

talex5:
With this change, the distros tested will be:
```
alpine-3.12          
debian-10
alpine-3.11
centos-7
centos-8
debian-9
debian-8
debian-testing
debian-unstable
fedora-31
fedora-32
oraclelinux-7
opensuse-15.1
opensuse-15.2
ubuntu-16.04
ubuntu-18.04
ubuntu-20.04
```
Previously, it was:

```
debian-10            
alpine-3.12
ubuntu-20.04
ubuntu-18.04
opensuse-15.2
centos-8
fedora-32
```

This is a pretty large increase. Are we sure we want to do that?

avsm:
I think it's smaller than that. From https://github.com/avsm/ocaml-dockerfile/blob/23e07f6fbf29cdb59bef5a61bda8de43b57c5438/src-opam/dockerfile_distro.ml#L51 it should be:

```
alpine-3.12
archlinux-latest
centos-7
centos-8
debian-10
debian-testing
debian-unstable
fedora-32
fedora-33
oraclelinux-8
opensuse-15.2
ubuntu-16.04
ubuntu-18.04
ubuntu-20.04
ubuntu-20.10
```

Of those we could trim out fedora-32 and make it deprecated, but the rest seem fine. The goal is to constrain them away in the project opam file so they wont be selected if broken, which in turn makes the submissions to upstream opam-repository more accurate.

talex5:
Ah, right. Forgot the submodule update :-)

This will still more than double the load on the cluster, though. If you're happy with that I'll deploy it.

avsm:
I think it's worth a try. I'll trim down the distros if it gets too unmanageable, but I'm curious to see how many bugs we ferret out with this wider testing.

talex5 (approved):
OK. I've rebased on master to fix the CI failure.

kit-ty-kate:
I don't think this PR should be merged.

kit-ty-kate:
The original goal of this PR was to give a list of distributions that are considered well supported in opam-repository. This was the original intent behind https://github.com/avsm/ocaml-dockerfile/pull/25. In that PR (which wasn't merged), I had split the distributions into 2 categories:
* tier 1: well supported in opam-repository
* tier 2: not well supported in opam-repository

Tier 1 included:
* Alpine 3.12
* CentOS 8 (this one was a mistake, I should've given CentOS 7 and move CentOS 8 in Tier2, even both in Tier2 as CentOS 7 is way too old now)
* Debian stable, testing and unstable
* OpenSUSE 15.2
* Ubuntu 18.04 (LTS) and 20.04 (latest stable at the time)

Tier 2 included:
* Oraclelinux 8

At the time the Archlinux support wasn't yet merged but I would've classified it as Tier2 for the moment if it were.

This PR diverges greatly from this original proposal and gets redundant, out of date and not well supported distributions tested, which isn't suitable for ocaml-ci.

talex5:
Yes, this doesn't seem very useful now. Perhaps we should scan the project for `*.c` files in the analysis phase and only include rare distributions if some are found?

talex5:
I've reverted the live deployment to master, to avoid annoying people with errors from old distros.

avsm:
But the whole point of getting older distros into ocaml-ci is that the packages will constrain them here if they are not supported _before_ being submitted to opam-repo.  

Is the issue here actually that there's no way to constrain the errors using opam 2.0 since we needed integrated depext?  In that case, I can modify this so that the older distros shift to Tier 3 and we use opam 2.1 in Tier3.  The ocaml-ci errors are only annoying if there's no way to fix them -- but with opam 2.1 there is a way to fix them.

avsm:
Note also the terminology:

- Tier 1 : build all possible base images (only Debian-10 and Alpine to cover musl)
- Tier 2: supported on opam-repo
- Tier 3: not well supported on opam-repo or older-but-still-EOL distributions

kit-ty-kate:
> But the whole point of getting older distros into ocaml-ci is that the packages will constrain them here if they are not supported _before_ being submitted to opam-repo.

Yes and no. Supporting a (version-specific or not) distribution doesn't really mean much and it's something that cannot and should not be enforced in opam-repository in my opinion.

There are several types of errors related to distributions:
* Missing dedicated content in one or sereval `conf-*` packages. This can:
  - require a change in opam-repository (in this case since it's failing on ocaml-ci, we're basically saying it's to the task of the developer to fix it, otherwise they won't ever have a passing CI)
  - be more tricky. Some packages are not available in the base distributions and are only available in external repositories (ppa for Ubuntu 16.04 for instance, …).
* Failures related to the version number of some of the components (too old GCC, too old library, …). This is tricky. On one side we could say that due to that the distribution is not compatible with the package. However does not say anything about what is actually available on the user's setup (maybe they installed a newer version of GCC or the library manually, or via an external PPA or something). Currently in opam-repository we simply ignore these kind of errors because they do not inherently make the package "unavailable" on a distribution.

I'm probably forgetting other types of failures, but anyway, in conclusion, in my opinion testing old and badly-supported* distributions in ocaml-ci is a bad idea.

*: IMO this currently covers at least oraclelinux, centos, ubuntu < 18.04 and archlinux

kit-ty-kate:
That said, what could be done to improve this issue would be to support the custom `x-ci-accept-failures: [unsupported-distributions]` opam field in ocaml-ci. This is what I'm using in opam-health-check to get rid of untestable packages.

This way we could support/test every the distributions without any issues.

kit-ty-kate:
I gave a shot at implementing the feature in https://github.com/ocurrent/opam-repo-ci/pull/62. Hopefully this should be easy enough to port to ocaml-ci as well, I'm guessing the way to do it would be to handle it directly in the analysis phase and skim off the distributions that won't be supported directly. I can't do that in opam-repo-ci since we don't really have the list of packages being installed before we actually run opam, so the solution is a bit more crude but it's basically the same thing as in opam-health-check so this should work.

Feel free to test it, I'll go to bed for now. Good night!

kit-ty-kate:
Support for `x-ci-accept-failures` has been tested successfully and merged in opam-repo-ci.

avsm:
I've pushed this to `live` now with fewer distros...

avsm:
Seems to be working fine, so merging. The distro list now has Ubuntu 20.10 instead of 18.04, as we might as well track the bleeding edge Ubuntu and the latest LTS. It's easy to put Ubuntu 18.04 back here if there is demand.

PR #265, opened 2020-10-25
Disable warnings-as-errors on old OCaml versions

kit-ty-kate:
Older versions of the compiler often trigger warnings that we might want to get rid of. For instance unsupported (at the time) documentation comments or something of the sort. There is easy way to deal with those for users of ocaml-ci so my suggestion would be to disable warnings-as-errors for all version of the compiler but the latest.

Ideally it would be best to detect the latest available version for the current packages but I couldn't get it to work in a reasonable amount of time so that'd be for another time, the regular extra checks such as flambda, and the diverse other distributions also suffer from this anyway.

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -85,12 +85,22 @@ let install_project_deps ~opam_files ~selection =
     run ~network ~cache "opam install $DEPS"
   ]
 
+let is_latest_ocaml v =
+  Ocaml_version.equal Ocaml_version.Releases.latest v
+
+let get_dune_option selection =
+  if is_latest_ocaml (Variant.ocaml_version selection.Selection.variant) then
+    "" (* default options *)
+  else
+    " --profile=release" (* Disable warnings-as-errors on older compilers *)
+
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
oops I forgot to link the compelling real world example: https://ci.ocamllabs.io/github/ocaml/opam-file-format/commit/2170573c17aeeaba1328910af733602d171a7147/variant/debian-10-4.02

talex5:
I think it would be better for projects where this is a problem to disable the warning themselves in their dune files. We're already running odoc using the latest version of the compiler, so I guess that would find important doc errors, even when it's disabled in dune.

kit-ty-kate:
The documentation issues are just an example, there can be other warnings that are triggered by older compilers but not on new one and dune does not have a way to give different flags to the compilers according to its version easily. I guess what I want is https://github.com/ocaml/dune/issues/90

Currently it can be done with a custom rule and `(:include ...)` directive but it's quite painful.
Fair enough, closing.

PR #268, opened 2020-11-05
Hotfix duniverse builds

emillon:
This is a set of assorted hotfixes for duniverse builds. Some of these are specific to the realworldocaml/book build but until duniverse is released this is all this workflow needs to support.

- install pkg-config before opam-tools (it used not to have any depexts, but now does due to uri -> angstrom -> bigstringaf -> pkg-config).
- fix installed dune version to 2.6.2
- install duniverse to the last version that shipped a "duniverse" binary (the latter commands refer to it)


emillon:
cc @NathanReb @avsm. This makes a green build for realworldocaml/book#master with `ocaml-ci-local` so I think this is the right way to go until we add support for duniverse (well, opam-monorepo) 0.1.0 in a couple weeks.

avsm:
Many thanks! Pushing this one live

PR #269, opened 2020-11-09
switch to ocaml/opam as the base image source

avsm:
see https://github.com/ocurrent/docker-base-images/pull/68


talex5 (approved):
Deployed. Waiting for everything to rebuild with the new base images...

PR #271, opened 2020-11-13
Add OCaml 4.12 to the production platforms

kit-ty-kate:
OCaml 4.12 was added to `ocaml-version` in https://github.com/ocurrent/ocaml-version/pull/20 and still need a release. Vendoring it allows us to test OCaml 4.12 in the meantime.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "ocaml-version"]
 	path = ocaml-version
 	url = git://github.com/ocurrent/ocaml-version.git
+[submodule "ocaml-dockerfile"]
+	path = ocaml-dockerfile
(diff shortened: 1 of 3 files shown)


talex5:
We might want to mark it as vendored in the `dune` file too.

talex5:
Hmm, with Lwt being incompatible with OCaml 4.12, there are pretty no much packages that can make use of this.

kit-ty-kate:
I don't think that's an issue, it'll just not show up in the build results for those, will for the packages that don't use lwt and when lwt is released everything will be already automatically in place

talex5:
OK. I've enabled it. Maybe we could get ocaml-ci enabled for Lwt itself?

PR #276, opened 2020-11-25
opam monorepo 0.1.0 support

emillon:
This contains support for opam-monorepo 0.1.0.

It detects the project based on the presence of a `pkg.opam.locked` file
with a `x-opam-monorepo-version` field.

Such builds perform the following steps:
- install `opam-monorepo` and the dune at the right version
- call `opam monorepo pull` on the repository to fetch dependencies
- call `dune build @install`
- call `dune runtest`

As follow-up for this work we'll need to:

- update the installation method to use opam once the plugin is released
- add lint builds
- install the plugin in a separate switch by hand or using opam-tools

### DIFF
--- a/lib/analyse.ml
+++ b/lib/analyse.ml
@@ -45,7 +45,11 @@ module Analysis = struct
   type t = {
     opam_files : string list;
     ocamlformat_source : Analyse_ocamlformat.source option;
-    selections : [ `Opam_build of Selection.t list | `Duniverse of Variant.t list ];
+    selections :
+      [ `Opam_build of Selection.t list
+      | `Duniverse of Variant.t list
+      | `Opam_monorepo of Opam_monorepo.selection
+      ];
   }
   [@@deriving yojson]
 
@@ -58,18 +62,13 @@ module Analysis = struct
 
   let opam_files t = t.opam_files
 
-  let is_duniverse t =
-    match t.selections with
-    | `Duniverse _ -> true
-    | `Opam_build _ -> false
-
   let ocamlformat_source t = t.ocamlformat_source
 
   let selections t = t.selections
 
   let is_test_dir = Astring.String.is_prefix ~affix:"test"
 
-  let get_ocaml_versions dir =
+  let duniverse_get_ocaml_versions dir =
     (* TODO Just manual right now to avoid a build dep on Duniverse_lib *)
     let dune_get_path = Filename.concat (Fpath.to_string dir) "dune-get" in
     let open Sexplib.Sexp in
@@ -90,26 +89,17 @@ module Analysis = struct
   let pp_ocaml_version f (_variant, vars) =
     Fmt.string f vars.Worker.Vars.ocaml_version
 
-  let ocaml_major_version vars =
-    Ocaml_version.with_just_major_and_minor (Ocaml_version.of_string_exn vars.Worker.Vars.ocaml_version)
-
   let duniverse_selections ~job ~platforms dir =
-    let vs = get_ocaml_versions dir in
+    let vs = duniverse_get_ocaml_versions dir in
     if vs = [] then failwith "No OCaml compilers specified!";
-    let find_compiler v =
-      let v = Ocaml_version.with_just_major_and_minor (Ocaml_version.of_string_exn v) in
-      let matches_compiler (_variant, vars) =
-        Ocaml_version.compare v (ocaml_major_version vars) = 0
-      in
-      match List.find_opt matches_compiler platforms with
-      | None ->
-        Current.Job.log job "WARNING: Unsupported compiler version %a (have: %a)"
-          Ocaml_version.pp v Fmt.(Dump.list pp_ocaml_version) platforms;
-        None
-      | Some (variant, _vars) ->
-        Some variant
+    let find_a_compiler version =
+      let r = Opam_monorepo.find_compiler ~version ~platforms in
+      if Option.is_none r then
+        Current.Job.log job "WARNING: Unsupported compiler version %s (have: %a)"
+          version Fmt.(Dump.list pp_ocaml_version) platforms;
+      r
     in
-    let variants = List.filter_map find_compiler vs in
+    let variants = List.filter_map find_a_compiler vs in
     if variants = [] then failwith "No supported compilers found!";
     Lwt_result.return (`Duniverse variants)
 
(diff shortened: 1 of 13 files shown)


emillon:
cc @NathanReb

avsm:
Thanks for the PR!  Some quick comments:

- if `opam-monorepo` is a plugin, then its installation is a bit special. opam builds the plugin in the current switch but installs the binary globally (into `~/.opam/plugins/bin`) so its available to all future switches.  So you could prod `opam-tools` into installing `opam-monorepo` globally using one of its tools switches, to avoid contaminating the local project's dependencies.

- what's the upstream `opam monorepo` support for handling depexts?  With the `duniverse` binary, it tracked depexts in the `dune-get` file.  We do need something equivalent with the opam files -- why aren't the depexts from the `opam` files just propagated?

- one the above are answered, a persistent build directory is indeed necessary. The existing duniverse uses a shared _build dir and gets very nice incremental build speeds as a result.

lib/analyse.ml
```diff
@@ -93,26 +100,76 @@ module Analysis = struct
...
+    |> List.assoc (OpamPackage.Name.of_string package)
+    |> version_of_equal_constraint
+
+  let x_opam_monorepo_version opam =
+    try
+    OpamFile.OPAM.extended opam "x-opam-monorepo-version" (function
+      | String (_, s) -> s
+      | _ -> assert false
+        )
+    with Invalid_argument _ -> None
```
talex5, line 156: Minor: please indent everything with ocp-indent.

talex5 (commented):
This approach looks good to me.

I think you probably will need to persist `_build` (the same way duniverse does it), otherwise it's going to be pretty slow.

emillon:
Regarding persisting `_build`, in the past [there has been a case in RWO](https://github.com/realworldocaml/book/pull/3405) where we had to run `dune clean` locally for the fix to kick in. I could see that happen again (rarely) in the future.
- is there a way to manually clear this cache?
- does this cache have an expiration time?
- should we clear this cache when the lock file changes (for example, by adding a hash of the lock file to the cache key)?

lib/analyse.ml
emillon: sure. is it ok to enable ocamlformat on new files? (I see there's an `.ocamlformat-enable` file)

talex5:
Each builder in the cluster has its own copy of the cache, and changes are not always saved (e.g. if two builds use the cache concurrently, only one of them will save its copy at the end). Also, the cache is shared between branches, so updating it to a new format on one branch might require downgrading it again on the other one.

Including the dune version in the cache ID might be a good idea (in case one branch tries to upgrade the cache format and another one doesn't want that), but apart from that I think it's dune's responsibility to ensure the cache is correct.

Caches don't expire at the moment, but we could add that easily enough.

lib/analyse.ml
talex5: I think that enable file is left over from a previous experiment; it should probably be deleted.

I don't want to mandate any particular formatting tool (although I said "ocp-indent" above, I really just mean that the indentation should be consistent; feel free to use ocamlformat on your new file locally before submitting it).

emillon: ok! it seems that I can't write ocaml without a formatter these days. I'll add a todolist item to format things once the rest is ready.

NathanReb:
> what's the upstream opam monorepo support for handling depexts? With the duniverse binary, it tracked depexts in the dune-get file. We do need something equivalent with the opam files -- why aren't the depexts from the opam files just propagated?

My intuition was that opam depext should be able to handle it from the lockfile. If it's not then I can include all the transitive depexts in the lockfile but in the long run I think it'd be better if `opam depext` can do the right thing. @emillon what exactly is the issue with `opam depext`? It can't take an opam file as an input so you have to pin the package and that fails for what reason?

emillon:
I thought that failed, but I tried again and it works. Maybe this was done at the wrong step. In any case, that part works now.

emillon:
Update for the depext situation: actually, it's not working. `opam pin -n --locked add .` interprets `add` as the lockfile suffix, and falls back to using the normal opam file (with a warning). When passing `--locked` last, it pins recursively. At the moment this fails in cases where virtual packages are used (e.g. `uchar`). We're looking at a possible solution for that.

NathanReb:
To clarify virtual packages are okay but not for packages+versions that aren't on the main opam-repository!

emillon:
The depext situation is now fixed. We need to recursively pin for now but recursive depexts will be added in a future version of opam-monorepo. I also added `_build` directory caching (the same that was used in duniverse).

I had a look at installing opam-monorepo through opam-tools, but this means building a compiler fairly late in the build process (after the opam file is available). In contrast, the current solution selects a base image based on the version in the lock file. This guarantees that no compiler is built, but this only works if the CI has the right image. The other drawback of the current solution is that the dependencies of opam-monorepo are present in the current switch, but this is a monorepo build so no other packages are going to be installed.

I'll have another look to see if it's possible to do the build a bit earlier (and somehow cache the various tool switches?) but otherwise these two limitations might be ok for a first version. What do you think?

NathanReb:
> The other drawback of the current solution is that the dependencies of opam-monorepo are present in the current switch, but this is a monorepo build so no other packages are going to be installed.

This is true now because we use a pinned version but the released one won't have any deps but dune and ocaml, rest will be vendored in the tarball so we can consider this one fixed. For the next release we'll also make sure we have a way to tell opam and opam-monorepo which dependencies have to come via opam and which ones need to be vendored. When we reach this point, the dependencies in the opam file will be properly interpreted, be it a released or a pinned version.

emillon:
And this is now ready to go.

In terms of release plan, @NathanReb and I were thinking that we:
- merge this
- make sure that realworldocaml/book#3412 is green (it is with ocaml-ci-local but who knows)
- release the plugin on opam
- update the plugin installation so that it is done from the opam release

And as next steps for support itself:
- add lint builds
- install the plugin in a separate switch by hand or using opam-tools

talex5 (approved):
Deployed and running!

emillon:
Thanks! the realworldocaml built went fine, so I think we can release the plugin. I'll update this PR once it's done.

talex5:
Can this be merged now? It makes it more difficult to deploy other features if they have to be temporarily merged with this unmerged branch first.

emillon:
Yes, that can be merged. I just pushed a quick fix so that it's usable in opam-monorepo itself. I'll do the rest of the work (update to use 0.1.0 from opam) in a separate PR. Thanks!

talex5:
Thanks!

PR #279, opened 2020-12-07
opam-monorepo 0.1.0: use released version

emillon:
This bumps the commit to use the released version.

### DIFF
--- a/lib/analyse.ml
+++ b/lib/analyse.ml
@@ -89,11 +89,26 @@ module Analysis = struct
   let pp_ocaml_version f (_variant, vars) =
     Fmt.string f vars.Worker.Vars.ocaml_version
 
+  let ocaml_major_version vars =
+    Ocaml_version.with_just_major_and_minor
+      (Ocaml_version.of_string_exn vars.Ocaml_ci_api.Worker.Vars.ocaml_version)
+
+  let find_compiler ~platforms ~version =
+    let v =
+      Ocaml_version.with_just_major_and_minor
+        (Ocaml_version.of_string_exn version)
+    in
+    let matches_compiler (variant, vars) =
+      if Ocaml_version.equal v (ocaml_major_version vars) then Some variant
+      else None
+    in
+    List.find_map matches_compiler platforms
+
   let duniverse_selections ~job ~platforms dir =
     let vs = duniverse_get_ocaml_versions dir in
     if vs = [] then failwith "No OCaml compilers specified!";
     let find_a_compiler version =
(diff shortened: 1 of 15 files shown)


emillon:
I'd like to install the plugin from opam, but it's not available yet in the workers (and running `opam update` does not work, but I think it was not a good idea anyway).

What's a good way to deal with this?

Thanks!

talex5:
When was it released? The base image builder ran today, so it might just be a case of asking the CI to use the newer images.

talex5:
`opam update` is fine - just remember to `git fetch && git reset --hard` first.

emillon:
> When was it released? The base image builder ran today, so it might just be a case of asking the CI to use the newer images.

That was merged 5 days ago: ocaml/opam-repository#17750. How can I tell the CI to use the newer images?

talex5:
For local testing, you can just click Rebuild on the `docker pull` boxes.

emillon:
Ah, that works. Thanks! I'll try to update this so that we can use `install_project_deps` instead of calling opam install manually here.

emillon:
Just did so - it's now doing the normal stuff to opam install dune and opam-monorepo at the right versions.

talex5 (commented):
Just putting the latest opam-repository version in the selections will force a rebuild of all monorepo targets every time there is a new commit to opam-repository. For opam builds, the solver searches backwards through the Git history to find the oldest commit with the same metadata for the selected packages. That way, irrelevant changes don't affect things. It might be worth sending a request for dune and opam-monorepo to the solver, just for that.

emillon:
Thanks for the tip. I'll try that.

emillon:
Done. I extracted a `solve` function and pass that to the opam-monorepo bits. I have a couple questions about that:
- I had to build a dummy opam file containing my dependencies (ocaml, opam-monorepo, dune) - is there a way around that? (by passing the constraints to the solver directly maybe?)
- this tries to solve this for all workers. since there's a strict ocaml version, this fails on "almost all" workers. should I try to filter the platforms based on the ocaml version first?

talex5:
Yeah, should be easy to pass in just the platform you want in this case.

emillon:
That worked. I realized that it probably won't work if a monorepo project locks a version with a different patch number (such as 4.10.0), but we can see if it's a problem later.

talex5 (approved):
Thanks - looks good to me!

(I moved things around a little to minimise the diff)

emillon:
Thanks! Yes, I wasn't too sure about the `solve` interface, this looks better this way.

PR #280, opened 2020-12-10
initial ocaml-ci-github idea

patricoferris:
I thought I would just open a draft PR about the `ocaml-ci-github` idea to generate ideas and discussion. It currently isn't quite at the `dune @ci && dune promote` step (shutting down the pipeline gracefully is something I'm not sure how to do) and it copies a lot of the code from `ocaml-ci-local` but I separated it into it's own folder to make it more readable. 

It only adds one thing to the library which is the ability to call the solver using only the OS variables (rather than providing the full `Platform.t`) with the hope that you can bypass the pulling of docker images and just solve for the different platforms to get the `$DEPS` variable. I think the longest thing is clone the opam repository!

Also, [this line of code needs improving](https://github.com/ocurrent/ocaml-ci/compare/master...patricoferris:github?expand=1#diff-4be58fdaac9b93d3349691b12189d777e1115f29543269f17f09674ee7eac864R81) 😅  (a lot!) but it kinda works. The main problem (I think) is that I don't know if you can get the hashes of the docker images w/out pulling them so I ended up just using generic images like `ocaml/opam:ubuntu-18.04-ocaml-4.11`. Ideally this pipeline would have not have to use Docker in order to make it quick.

[Here is an example of it in the real world](https://github.com/patricoferris/ppx_deriving_yaml/actions/runs/412720781). It would be good to also test it on something that solves differently depending on the platform. A part of me does wonder if this really needs the OCurrent pipeline?

### DIFF
--- a/github/dune
+++ b/github/dune
@@ -1,0 +1,29 @@
+(executables
+ (names main)
+ (libraries logs.fmt
+            fmt.tty
+            current
+            current_web
+            current_docker
+            current_git
+            current_github
+            current_rpc
+            dockerfile-opam
+            ocluster-api
+            capnp-rpc-unix
+            mirage-crypto-rng.unix
+            ocaml_ci
+            ocaml-ci-api
+            prometheus
+            github-workflow)
+ (preprocess (pps ppx_deriving_yojson)))
+
+(install
+ (section bin)
+ (package ocaml-ci-service)
+ (files (main-copy.exe as ocaml-ci-github)))
+
+ (rule
+  (target main-copy.exe)
+  (deps (package ocaml-ci-solver))
+  (action (copy main.exe main-copy.exe)))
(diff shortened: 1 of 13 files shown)


talex5:
>The main problem (I think) is that I don't know if you can get the hashes of the docker images w/out pulling them

`Docker.peek`? https://github.com/ocurrent/ocurrent/blob/master/plugins/docker/s.ml#L30

maiste:
Is this PR introducing what we have in `services/pipeline`? Or am I missing the point of this PR? In case this is not required, can we close this PR?
(cc @tmcgilchrist)

tmcgilchrist:
Closing because this is stale.

Perhaps @patricoferris remembers what this was originally aiming for. It does seem to be an early version of the pipeline

PR #287, opened 2020-12-16
add hannesm, roburio, and mirleft to allow list

hannesm:
thanks for your work, this would allow me to deploy ocaml-ci for more projects.


talex5:
Done! Don't forget to enable it just for the repositories you want (it doesn't cope well with "all repositories" at the moment).

PR #288, opened 2020-12-16
Opam monorepo 0.2 support

### LINKED ISSUE #282: monorepo: fails when dependencies are locked using a newer version of opam-repository
Hi,

At the moment the following happens for monorepo builds:
- during the analysis phase, the solver is queried for ocaml, dune and opam-monorepo. The oldest opam-repository commit that knows about that solution is selected.
- during the build phase, the lock file is pinned in order to resolve depexts.

The issue is that if the lock file has been produced with an opam-repository state that's more recent than the result of the analysis, it won't be able to compute the depexts.

As an example (inspired by a true story with RWO), querying for ocaml=4.10.1, dune=2.7.1, opam-monorepo=0.1.0 resolves to ocaml/opam-repository@aaf8fed5fa134979c4552e283800798ac3f64366. If the project refers to conf-gmp, today it will get locked to version 3. But this version has been uploaded only after that opam-repository commit.

As a workaround, conflicting with conf-gmp > 2 kind of works, but this is only a short term solution since it will apply to all transitive dependencies released after.

The root issue is that we need access to the repository to determine depexts. That's a known opam-monorepo 0.1.0 limitation which is going to be lifted soon, but I'm wondering if there's something we can do in the meantime on the CI side. Maybe passing the pin-depends to the solver would work, but if done too naively it will try to install them via opam in the build phase.

emillon:
This is the same as 0.1 except that transitive depexts are present in the lock file:
    
- they can be listed directly by parsing the lock file, with no access to opam-repository
- the "pull" command of 0.1.0 is not affected by this. So it is possible to install the 0.1.0 plugin and rewrite the version in the lock file so that it does not exit with an error.

Closes #282

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -22,34 +22,34 @@ let group_opam_files =
     )
 
 let mkdir dirs =
-  if dirs = [] then
-    []
-  else
-    let dirs =
-      dirs
-      |> List.map (fun (dir, _, _) -> Filename.quote (Fpath.to_string dir))
-      |> String.concat " "
-    in
-    [Obuilder_spec.run "mkdir -p %s" dirs]
+  let dirs =
+    dirs
+    |> List.map (fun (dir, _, _) -> Filename.quote (Fpath.to_string dir))
+    |> String.concat " "
+  in
+  [Obuilder_spec.run "mkdir -p %s" dirs]
 
 (* Generate instructions to copy all the files in [items] into the
    image, creating the necessary directories first, and then pin them all. *)
 let pin_opam_files ~network groups =
-  let open Obuilder_spec in
-  mkdir groups @ (
-    groups |> List.map (fun (dir, files, _) ->
-        copy files ~dst:(Fpath.to_string dir)
-      )
-  ) @ [
-    groups |> List.concat_map (fun (dir, _, pkgs) ->
-        pkgs
-        |> List.map (fun pkg ->
-            Printf.sprintf "opam pin add -yn %s %s" pkg (Filename.quote (Fpath.to_string dir))
-          )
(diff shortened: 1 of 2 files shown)


emillon:
Keeping this WIP until the corresponding plugin is released.

emillon:
This one is now ready. It actually does not need a plugin release so can be merged when convenient.

talex5 (changes requested):
I tried this on realworldocaml and it generated an empty `RUN` line:
```
RUN sudo chown opam /src
RUN cd ~/opam-repository && (git cat-file -e 51576718e3d32386fcc2248de168f74e990b4a42 || git fetch origin master) && git reset -q --hard 51576718e3d32386fcc2248de168f74e990b4a42 && git log --no-decorate -n1 --oneline && opam update -u
RUN 
```
I guess it's from `pin_opam_files`.

emillon:
I just fixed it. This used to fail because it would generate `mkdir -p` (and no directories), so I filtered in `mkdir`, but it's cleaner to filter in `pin_opam_files` to avoid the empty `RUN` indeed.

If you want to try it on a 0.2 build, there's a branch `format-0-2` on realworldocaml/book (which builds fine locally).

talex5:
I don't see a [format-0-2](https://github.com/realworldocaml/book/tree/format-0-2) branch, but it seems happy on `master` so merged!

emillon:
Ah, I forgot to push it indeed. I cherry picked it on an existing PR, and it worked: https://ci.ocamllabs.io/github/realworldocaml/book/commit/60b370cd69cb8fbcfedd22ecf45db1ed06acdb37/variant/debian-10-4.10

PR #298, opened 2021-01-13
Use monorepo 0.2.0 instead of downgrading lockfile

emillon:
Previously the behaviour was:

- for 0.1 lock file, use plugin 0.1.0
- for 0.2 lock file, downgrade lock file version and use 0.1.0

Now that plugin 0.2.0 has been released and understands the 0.1 format we can relax the constraints a bit:

- for 0.1 lock file, use plugin `>= 0.1.0 & < 0.3.0`
- for 0.2 lock file, use plugin `>= 0.2.0 & < 0.3.0`


talex5:
(rebased to get fixed git-unix bound)

PR #299, opened 2021-01-14
Fix opam-dune-lint (port ocurrent/opam-dune-lint#20)

kit-ty-kate:
The `(lint-opam)` step of ocaml-ci will fail until this is merged.

See: ocurrent/opam-dune-lint#20

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -56,7 +56,7 @@
   ocaml-ci-api
   conf-libev
   (opam-0install (>= "0.2"))
-  git-unix
+  (git-unix (< 3.0.0))
   (capnp-rpc-unix (>= 0.7.0))))
 (package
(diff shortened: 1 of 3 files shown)


talex5 (changes requested):
ocaml-ci-solver.opam is generated by dune, edit dune-project instead.

If you use vim, you might find this useful:

```
au BufRead,BufNewFile *.opam
\ if match(getline(1), "This file is generated by dune") >= 1 |
\   set readonly |
\ endif
```

kit-ty-kate:
> au BufRead,BufNewFile *.opam
> \ if match(getline(1), "This file is generated by dune") >= 1 |
> \   set readonly |
> \ endif

Oh that's cool! Thanks!

PR #300, opened 2021-01-14
Add upper-bound on Git

talex5:


talex5:
Already fixed by #299.

PR #302, opened 2021-01-14
monorepo: support exact compiler version

### LINKED ISSUE #281: monorepo: does not support compiler versions not in base images
Hi,

We're having the following situation in RWO:
- the version for ocaml and ocaml-base-compiler is set to 4.10.2 in the lock file
- in the analysis phase, it considers only 4.10 platforms and queries for (ocaml=4.10.2, opam-monorepo=0.1.0, dune=2.7.1)
- this fails because the platform has 4.10.1 installed:

```
[INFO] Can't find all required versions.
Selected: base-bigarray.base base-threads.base base-unix.base ocaml-config.2
          ocaml-base-compiler&opam-monorepo-deps ocaml-base-compiler
          ocaml-variants
- ocaml -> ocaml.4.10.1
    ocaml-base-compiler 4.10.1 requires = 4.10.1
- ocaml-base-compiler -> ocaml-base-compiler.4.10.1
    User requested = 4.10.1
- ocaml-variants -> (problem)
    Rejected candidates:
      ocaml-variants.4.13.0+trunk+no-flat-float-array: In same conflict class (ocaml-core-compiler) as ocaml-base-compiler
      ocaml-variants.4.13.0+trunk+nnp: In same conflict class (ocaml-core-compiler) as ocaml-base-compiler
      ocaml-variants.4.13.0+trunk+fp+flambda: In same conflict class (ocaml-core-compiler) as ocaml-base-compiler
      ocaml-variants.4.13.0+trunk+fp: In same conflict class (ocaml-core-compiler) as ocaml-base-compiler
      ocaml-variants.4.13.0+trunk+flambda: In same conflict class (ocaml-core-compiler) as ocaml-base-compiler
      ...
- opam-monorepo-deps -> (problem)
    Rejected candidates:
      opam-monorepo-deps.dev: Requires ocaml = 4.10.2
```

I'm not sure what to do in that case; it would be fine to create a new switch where dune and opam-monorepo would be installed.

A similar problem would appear if the project locks a compiler variant (e.g. the application needs flambda).

emillon:
When the exact compiler version (down to the patch level) is available in a base image, it is used. However, if it is not (for example the base image has 4.10.2 but the project has 4.10.1 in its lockfile), it would fail.

This adds a fallback mechanism for this situation by using the closest base image (4.10 in the example) and creating a switch with the exact version.

Closes #281

### DIFF
--- a/lib/opam_monorepo.ml
+++ b/lib/opam_monorepo.ml
@@ -2,10 +2,17 @@ type info = string * OpamFile.OPAM.t
 
 type lock_file_version = V0_1 | V0_2 [@@deriving yojson, ord]
 
+(** The kind of switch the package will be built in. *)
+type switch_type =
+  | Base  (** When the locked version is present in a base image, use it. *)
+  | Create of { compiler_package : string }  (** Otherwise, create a switch. *)
+[@@deriving yojson, ord]
+
 type config = {
   package : string;
   selection : Selection.t;
   lock_file_version : lock_file_version;
+  switch_type : switch_type;
 }
 [@@deriving yojson, ord]
 
@@ -86,6 +93,28 @@ let opam_dep_file packages =
   in
   lines |> List.map (fun s -> s ^ "\n") |> String.concat ""
 
+let switch_type ~platforms requested_ocaml_version =
+  let has_exact_match =
+    platforms
+    |> List.exists (fun (_, (vars : Ocaml_ci_api.Worker.Vars.t)) ->
+           let platform_version =
+             Ocaml_version.of_string_exn vars.ocaml_version
+           in
+           Ocaml_version.equal platform_version requested_ocaml_version)
+  in
+  if has_exact_match then Base
+  else
+    let compiler_package = Ocaml_version.Opam.V2.name requested_ocaml_version in
+    Create { compiler_package }
+
+(** Eliminate platforms with an incompatible version and force the patch version
+    if that matches (it will be fulfilled for any [switch_type]). *)
+let adjust_ocaml_version platform ~version =
+  let p, vars = platform in
+  if Platform.compiler_matches_major_and_minor ~version vars then
+    Some (p, Platform.set_compiler_version ~version vars)
+  else None
+
 let selection ~info:(package, lock_file) ~platforms ~solve =
   let open Lwt_result.Infix in
   let ocaml_version = opam_monorepo_dep_version ~lock_file ~package:"ocaml" in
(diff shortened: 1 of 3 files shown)


emillon:
This approach can be extended to support other variants (for example if a project locks 4.10.2+flambda).
It can be tested on realworldocaml/book:
- realworldocaml/book#3420 (21de6252) uses 4.10.2, so it can use the base image
- master (2bab7768) uses 4.10.1 so it used to work but now fails. With this PR, it builds a switch.

talex5:
There was some discussion of this on Slack. @avsm suggested that the lock file should allow upgrading the micro version of the compiler, rather than forcing the CI to use the older version.

avsm:
I can see why @emillon wants this though, since there may be other reasons why an exact compiler version is required. No objections from me to this going in, but it would be nice to have an issue in opam-monorepo to track the "version relaxation by default".

The `relax_version` function could go into ocaml-version as well. A small bug in the formula generated is that it should be `<4.12~~` (the `between` function doesnt add the twiddles)

emillon:
I just opened ocamllabs/opam-monorepo#156 regarding this. Indeed, this workaround does not cache well so we need to make sure that the default solution is a good one in the general case.

(and I added some tests for the constraints function + added the `~~`)

talex5 (commented):
So if I'm understanding this correctly, it does this:

1. Get the lock file's requested OCaml version.
2. Check whether that matches a pre-existing platform. If not:
3. Find the nearest platform (same major.minor).
4. Ask the solver to solve on that platform for any compiler with that major.minor, not just the one requested.
5. At build time, `opam switch create` the version we actually want.

This seems slightly wrong. Why not adjust the platform's OCaml version to be the one it will actually have, and then ask the solver to find the exact version requested?

Otherwise, the solver could give the wrong solution (e.g. selecting a package that works only with OCaml 4.10.1 when the lock requested 4.10.0).

avsm:
@emillon any progress on @talex5's suggestion?

emillon:
I agree with that suggestion, I'll try to implement it next week.

emillon:
Done! We just implemented that suggestion: the solver always sends the exact ocaml version to an adjusted platform.

NathanReb:
Any chance we can get this merged soon?

Sorry to be so pushy it would greatly help to safely resume development of `opam-monorepo`!

avsm:
Pushed to live-engine branch, will merge after the live one's showing green :)

NathanReb:
Looking good here:
https://ci.ocamllabs.io/github/ocamllabs/opam-monorepo/commit/69d1929adc4ac0becee6ee186d174a5a713a65c5

avsm:
Thanks!

PR #309, opened 2021-01-25
lint-fmt: grant network access to opam pin

emillon:
This happens in the case where ocamlformat is vendored.
If its opam file contains `pin-depends`, it needs network access to be installed.

### DIFF
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -7,7 +7,7 @@ let install_ocamlformat =
     let opam_file = Filename.concat path "ocamlformat.opam" in
     [
       copy [ opam_file ] ~dst:opam_file;
-      run "opam pin add -k path -n ocamlformat %S" path;
+      run ~network "opam pin add -k path -n ocamlformat %S" path;
       (* Pinned to a directory containing only the .opam file *)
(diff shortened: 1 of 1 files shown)


emillon:
Here's a build where this could benefit ocaml-ci: https://ci.ocamllabs.io/github/ocaml-ppx/ocamlformat/commit/928ca6a8b47318b78d99a1faa59349d65526f0c7/variant/(lint-fmt)

avsm:
I'm ok with this, but it seems like a bug in opam that `opam pin add -n` results in network traffic. The manual page says:

```
       -n, --no-action
           Just record the new pinning status, and don't prompt for
           (re)installation or removal of affected packages.
```

And fetching the remote sources seems like a pretty significant action. Maybe this just needs a revision to the manual page entry upstream in opam.

emillon:
That surprised me as well, but I think that's due to the recursive nature of pin-depends, but I'm not sure how to fix that with the current UI. What we're doing is:

- `opam pin add -n`
- `opam depext`
- `opam install --deps-only`

If `pin add` does not access the network, then this network access would need to be "resolved" in `opam depext` and somehow written back to the switch state.

PR #311, opened 2021-02-18
Update to ocaml-git > 3.2.0

MisterDA:
Yes, apparently it's that simple.


avsm:
Or not :-) You'll need to update the opam-repo checkout hash in the Dockerfile so that the new library is found in the image

avsm:
pushed to `live-engine` to test

Issue #297, opened 2021-01-08
Test packages on the full range of supported OCaml versions


CraigFe:
To my understanding, the analysis phase for Opam projects works roughly as follows:
- determine the set of OCaml versions at which all packages are co-installable (by attempting to solve the constraints);
- for each such version, emit the set of opam packages to install (as returned by the constraint solver).

This has the consequence that only OCaml versions at which all packages are co-installable are tested in CI, even if certain packages have support for a larger set. This is problematic for repositories containing many packages, since it can severely limit the test matrix. For instance, Irmin is currently only tested on OCaml 4.11 despite almost all of the packages supporting back to at least 4.08.

It would be nice to have a mechanism that tests each package on the full range of OCaml versions that it (allegedly) supports. An obvious solution is to branch the pipeline for each package and install dependencies separately – as was previously done with `ocaml-ci-scripts` – but this is inefficient and doesn't provide any information about co-installability. Here's an initial proposal for how we might do better:

- for each OCaml version in the test matrix, determine the largest co-installable subset of opam packages `p_i`;
- for each non-empty `p_i`, emit the set of opam dependencies to install _and_ the set of Dune packages `d_i` corresponding to `p_i`. The test phase can then use `dune runtest --only-packages [... d_i]` to ensure that only these packages are tested at that version.

The magic here is in calculating the `p_i`s and `d_i`s. I suspect `p_i` can be approximated by initially assuming total co-installability and then using counter-examples to iterate downwards, but perhaps there is better built-in solver magic for this. A complication is that bizarre sets of constraints can result in _multiple_ `p_i`s, in which case it seems sufficient to just pick one deterministically. Once that's done, `d_i` might be computable from `p_i` using `opam-dune-lint`?

talex5:
This might be easier:

If a package's opam file contains an explicit constraint on the version of OCaml then:
- Exclude it from non-matching versions
- Force it to be included for matching versions (complain if not)

CraigFe:
> If a package's opam file contains an explicit constraint on the version of OCaml then:
> 
>     * Exclude it from non-matching versions
> 
>     * Force it to be included for matching versions (complain if not)

I'm not sure I like the idea of adding artificial constraints in `.opam` files purely to signal intention to the CI – it's not easy to check when / if it's possible to relax those constraints – but it's definitely easier to implement and clearly better than the prior approach of replicating this info in config files. I'd be happy with this as a solution :slightly_smiling_face:

[...]

PR #314, opened 2021-02-25
Test the opam file as well

kit-ty-kate:
The content of the opam file is never tested by ocaml-ci. Several users have had the issue in the past where `dune build` works fine but `dune build -p <pkg>` does not (e.g. merlin's testsuite)

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -97,6 +97,7 @@ let spec ~base ~opam_files ~selection =
     user ~uid:1000 ~gid:1000 ::
     install_project_deps ~opam_files ~selection @ [
       copy ["."] ~dst:"/src/";
-      run "opam exec -- dune build @install @runtest && rm -rf _build"
+      run "opam exec -- dune build @install @runtest && rm -rf _build";
+      run "opam install .";
     ]
   )


dra27:
Does this need to be done on every platform? It doubles the build work - wouldn’t it be better as a next pipeline stage after the other builds have succeeded (for example, just on Ubuntu)?

avsm:
This will also fail if not all the packages are available on that particular switch.  I think we need something more on the lines of the discussion in  #297 where an explicit subset of packages can be passed through from the solver to the platform.

talex5:
This will likely also cause problems for vendored dependencies (that will be removed before the final release to opam-repository).

Is this a big problem? I'd expect any problems here would be picked up at release time by dune-release and would be easily fixed then.

tmcgilchrist:
> Is this a big problem? I'd expect any problems here would be picked up at release time by dune-release and would be easily fixed then.

If we wanted this functionality to continuously check the "releasability" of packages, then it should be a limited linting step that applies to a single linux distro (Ubuntu) and runs after the other builds have succeeded. Potentially only running on the default branch rather than all git refs.

shonfeder:
Closing due to the age of this PR and the various concerns raised. If we move forward with this, it would be along the lines of https://github.com/ocurrent/ocaml-ci/pull/314#issuecomment-785890604 .

PR #315, opened 2021-02-25
Add MisterDA to allow list

MisterDA:


talex5:
I've also pushed the updated stack live (with `make deploy-stack`).

PR #316, opened 2021-02-25
Replace calls to find with Bos.OS.Path.fold and Fpath

MisterDA:


lib/dune
```diff
@@ -1,6 +1,6 @@
 (library
   (name ocaml_ci)
   (preprocess (pps ppx_deriving.std ppx_deriving_yojson))
-  (libraries logs current current_docker current_github current.term ocaml-ci-api str capnp-rpc-unix
+  (libraries bos logs current current_docker current_github current.term ocaml-ci-api str capnp-rpc-unix
```
avsm, line 4: you need to add this to the opam file too

lib/analyse.ml
```diff
@@ -163,44 +161,51 @@ module Analysis = struct
...
-            let full_path = Filename.concat (Fpath.to_string dir) path in
-            if is_empty_file full_path then (
-              Current.Job.log job "WARNING: ignoring empty opam file %S" path;
-              None
-            ) else if consider_opam_file path then
-              Some path
-            else None
-        )
+      let opam_files full_path =
+        let segs = Fpath.segs full_path |> List.tl in
```
avsm, line 166: Are you sure this will never be an empty list?

MisterDA: I think so, the full path returned by `Bos.OS.Path.fold` seems to always start with `./` which is the first segment that I'm dropping. I'll replace it with a check, I can never be so sure.

MisterDA:
Added fixes for your comment, thanks!
I marked this PR as draft because I cannot test it yet on Windows. I was hoping the CI's CI would validate it.

lib/analyse.ml
```diff
@@ -163,44 +161,51 @@ module Analysis = struct
...
-            if is_empty_file full_path then (
-              Current.Job.log job "WARNING: ignoring empty opam file %S" path;
-              None
-            ) else if consider_opam_file path then
-              Some path
-            else None
-        )
+      let opam_files full_path =
+        let segs = match Fpath.segs full_path with "./" :: tl -> tl | l -> l in
+        let path = Fpath.(to_string (Option.get (rem_prefix dir full_path))) in
```
avsm, line 167: Same question as before about the Option.get -- you seem to later on have an error messsage when Fpath.rem_prefix is None, but here it would just leak an error. Can you perhaps move `path` down to where it's used in the function and have the same Error return?

(the use of `dir`, `path`, `full_path` are all a little non-descriptive to me -- perhaps a comment noting that the input of the function is intended to be a relative path would be helpful)

talex5:
Note that Bos uses blocking IO, which might be a problem. We used to use blocking IO to delete the Git checkout at the end and that could sometimes lead to long latencies (more that 10 seconds; enough to trigger a warning from the monitoring). Reads might be less of a problem though; not sure. It probably depends how busy the service is.

lib/analyse.ml
MisterDA: I've fixed and simplified the code, I think.
Here the `Option.get` is valid because we start the fold from `dir` and from what I can see it's always the prefix of visited paths.
I've changed the variables: `root` is always the root from where the search is done, `path` the current path, and `test_path` the current test, so that we have `root / test_path = path`. I think it's clearer.

I may need to test this patch further, so I'm leaving it as draft for now.

MisterDA:
Thanks for the feedback, I've pushed fixes -- now I'm using Lwt_result to propagate errors occurring during the fold.
Should I look for some way to execute this in a background thread/process? The fold should be faster than a git checkout though.

talex5:
You can use `Lwt_preemptive.detach` to run blocking code in a separate thread.

MisterDA:
Maybe it would be more reliable to use something from opam (OpamRepositoryState ?) instead of a hand-crafted crawl?

dra27:
I don't think that would work - this is trawling _source_ repositories, not _opam_ repositories.

maiste:
@MisterDA is this PR still relevant or can we close it?

MisterDA:
Relevant to get ocaml-ci working locally on Windows, but it's blocked anyway until we get the solver service.

tmcgilchrist:
@MisterDA Can this be revisted now the solver changes are merged?

maiste:
We discussed with @MisterDA and I'm coing rebase or adapté the solution to the HEAD of master.

benmandrew (approved):
LGTM 👍

dune-project
```diff
@@ -51,6 +51,7 @@
   (fmt (>= 0.8.9))
   (cmdliner (>= 1.1.0))
   (ansi (>= 0.6.0))
+  bos
```
tmcgilchrist, line 54: Please pick a lower bound on this.

lib/analyse.ml
```diff
@@ -256,48 +255,59 @@ module Analysis = struct
...
+                (Fpath.to_string path)
+      in
+      let add_opam_files path acc =
+        match opam_files path with None -> acc | Some path -> S.add path acc
+      in
+      Bos.OS.Path.fold ~elements:(`Sat is_opam_ext) ~traverse:(`Sat traverse)
+        add_opam_files S.empty [ root ]
+      |> Result.map S.elements
     in
+    Lwt_preemptive.detach fold_on_opam_files () >>!= fun opam_files ->
```
tmcgilchrist, line 301: Why did you choose to use `Lwt_preemptive` here?

maiste: It was from previous code and related to this comment: https://github.com/ocurrent/ocaml-ci/pull/316#issuecomment-790468851

MisterDA:
I think this is good now, the CI and tests passes. Can we test this at scale and make sure we've not introduced buggy corner-cases?

maiste:
We can deploy it on `live-engine` for tests for one or two days. If it's ok, we merge it in master, otherwise, we revert the change. WDYT?

PR #321, opened 2021-03-04
Update ocamlformat version (no-op, no reformat)

MisterDA:
This is almost a no-op as ocamlformat is disabled, but makes all the
tests go green on OCaml 4.12 as ocamlformat.0.15 requires < 4.12. This
just statisfies the Opam solver.

(At least that's what I'm hoping for, let's see how the CI actually reacts)


avsm:
Removed the ocamlformat file instead in #322

PR #332, opened 2021-04-12
Write any _build/**/*.log file to the build log before completing.

ewanmellor:
This skips the main dune build log (_build/log) since that is verbose
and doesn't seem very useful, but this is intended to catch other logs,
such as those written by tests.

Signed-off-by: Ewan Mellor <ewan@tarides.com>

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -90,6 +90,14 @@ let install_project_deps ~opam_files ~selection =
     run ~network ~cache "opam install $DEPS"
   ]
 
+let cat_all_logs =
+  let open Obuilder_spec in
+  run "find _build -type f -a -name '*.log' -a '!' -empty -exec /bin/sh -c 'echo {}:; echo; cat {}; echo' ';' || true"
+
+let rmrf_build =
+  let open Obuilder_spec in
+  run "rm -rf _build"
+
 let spec ~base ~opam_files ~selection =
   let open Obuilder_spec in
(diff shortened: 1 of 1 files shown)


lib/opam_build.ml
```diff
@@ -90,13 +90,23 @@ let install_project_deps ~opam_files ~selection =
...
   let open Obuilder_spec in
   stage ~from:base (
     comment "%s" (Fmt.strf "%a" Variant.pp selection.Selection.variant) ::
     user ~uid:1000 ~gid:1000 ::
     install_project_deps ~opam_files ~selection @ [
       copy ["."] ~dst:"/src/";
-      run "opam exec -- dune build @install @runtest && rm -rf _build"
+      run "opam exec -- dune build @install @runtest";
+      cat_all_logs;
+      rmrf_build
```
talex5, line 110: There's no point deleting `_build` in a separate build step, since we'll still create and cache the layer with the files.

talex5 (commented):
I'm not sure this is a good idea. It's likely to spam the log with irrelevant test output. Also, test logs often deliberately trigger error cases. We don't want the log analyser picking these up.

We really just want to see the log of the failing test(s), which the test suite should do by default. If there are projects where that isn't happening, I think we should fix the upstream test library. 

Possibly a test library could detect it's running under CI and increase its verbosity (e.g. showing all failing logs rather than just the first) if that's useful.

PR #333, opened 2021-04-12
Extend the handling of opam-repository in the solver to use multiple

ewanmellor:
such repos.

This allows us to handle the case where some packages in the build are
declared in a separate repo, e.g. ocaml-multicore/multicore-opam.

This extends Opam_repository to accept a repo URL with each call,
and to infer a local directory name based on that URL. This means that
it no longer is hardcoded to use opam-repository as the local clone
name, nor to use the GitHub ocaml/opam-repository as the only repo.

This adds the ability to call oldest_commits_with across multiple repos,
with the semantics that we partition the packages based on which repo
we find them in, and return the oldest commit for each partitioned set.

This extends the solver service to handle multiple opam-repos. To do
this, the API to the service has been changed to pass a
(string * string) list in place of the old opam-repository commit. We
no longer clone opam-repository when the solver service is started, and
instead wait until the first request arrives. To support this,
Opam_repository.open_store will now clone the repo if the local clone
does not yet exist.

This adds a helper module Remote_commit, which is *very* similar to
Current_git.commit_id. It felt wrong to have solver depend on an
Ocurrent plugin, so I added this new module, but I feel like this could be
cleaned up and we could have a utility module somewhere common (maybe even
ocaml-git should be providing this).

This adds a dependency on Str from ocaml-ci-solver, to use a regexp for
sanitizing the repo URL into a local directory name.

Note that all of this doesn't change the way that Opam_build handles
opam-repository.  Our current Docker containers for multicore have
the primary opam-repository cloned locally and registered with opam
through a file: URL, but the multicore repo is registered as a
git://github.com URL. This means that Opam_build retains its
special-case code for refreshing ~/opam-repository inside the container,
and all other refreshing is handled by opam.

Signed-off-by: Ewan Mellor <ewan@tarides.com>

### DIFF
--- a/api/worker.ml
+++ b/api/worker.ml
@@ -18,14 +18,16 @@ module Selection = struct
   type t = {
     id : string;                        (** The platform ID from the request. *)
     packages : string list;             (** The selected packages ("name.version"). *)
-    commit : string;                    (** A commit in opam-repository to use. *)
+    commits : (string * string) list;   (** The commits in each opam-repository repo to use.
+                                            A pair of the repo URL and the commit hash. *)
   } [@@deriving yojson, ord]
 end
 
 (** A request to select sets of packages for the builds. *)
 module Solve_request = struct
   type t = {
-    opam_repository_commit : string;            (** Commit in opam repository to use. *)
+    opam_repository_commits : (string * string) list;
+    (** Pair of repo URL and commit hash, for each opam-repository to use. *)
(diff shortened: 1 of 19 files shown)


avsm:
Looks good to me on first review. I'm building on top of this PR to add multicore-opam support to ocaml-ci, so will report back when that works. (might as well merge this alongside the actual feature it supports).

PR #334, opened 2021-04-12
Enable verbose mode on the call to Solver.diagnostics.

ewanmellor:
Without the verbose flag, the 0install solver will truncate the
rejects list in the diagnostics output at 5 lines. These go into the job
logs, so there's no need to truncate the info.

### DIFF
--- a/solver/solver.ml
+++ b/solver/solver.ml
@@ -38,7 +38,7 @@ let solve ~packages ~pins ~root_pkgs (vars : Worker.Vars.t) =
     let pkgs = Solver.packages_of_result sels in
     Ok (List.map OpamPackage.to_string pkgs)
   | Error diagnostics ->
-    Error (Solver.diagnostics diagnostics)
+    Error (Solver.diagnostics diagnostics ~verbose:true)
 
 let main commit =
   let packages =


talex5 (commented):
Have you found a case where this is useful? The more verbose the logs are, the longer it takes to spot the problem.

@TheLortex has been complaining recently that it's too verbose already: https://github.com/ocaml-opam/opam-0install-solver/issues/27

ewanmellor:
I had a situation where the reported list of rejections did not include the packages that I was expecting to see.  This was very confusing, as I thought I had provided a completely wrong opam-repository, before I realized that the rejection list was being truncated.

In my opinion, you should never truncate build logs. By their nature, they're "after the event", so if you throw information away then you're never getting that back.  Debugging intermittent issues becomes a complete nightmare if the logs are truncated.

If we need a more readable summary that is quickly readable then we can do that in the UI layer or improve the output to include summary information before the detail.

And with all that said, this truncation is particularly bad. It just adds "..." on a new line in the log, which is practically invisible.  Saying "... and 15 others" would have been a great help.

ewanmellor:
Screen Shot 2021-05-10 at 2 02 45 PM

Here is a good example.  I'm looking for the 4.12.0+domains+effects compiler.  The diagnostics don't show that compiler, and instead list of bunch of alphas and betas at the top of the list, so the 5 slots displayed are useless, and it's not at all clear what the actual problem is.  (Showing the full list made it really obvious that the correct compiler package wasn't even in the list of rejects, i.e. in this case the problem is that opam is being given the wrong opam-repository.)

talex5:
If it had been in the list it would have shown only that version:

https://github.com/0install/0install/blob/22eebdbe51a9f46cda29eed3e9e02e37e36b2d18/src/solver/diagnostics.ml#L144-L151

So perhaps it just needs a hint after "No usable implementations" saying "(no versions match the user requested constraint)" to make this clear.

PR #340, opened 2021-05-19
Fix solver worker failed: Unexpected end of input

emillon:
In some cases, the C git client can create pack files that ocaml-git cannot read. See mirage/ocaml-git#514, mirage/ocaml-git#515.

This is fixed in `carton.0.4.2` (an internal dependency of `ocaml-git`). Since we do not directly use it in `ocaml-ci`, this upgrade is done through a `conflicts` clause.

It can be removed once a version of `ocaml-git` is released with a bound to `carton` greater than `0.4.2`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM ocaml/opam:debian-10-ocaml-4.10 AS build
 RUN sudo apt-get update && sudo apt-get install libev-dev capnproto m4 pkg-config libsqlite3-dev libgmp-dev graphviz -y --no-install-recommends
-RUN cd ~/opam-repository && git pull origin -q master && git reset --hard 1d088847ed8e4e6329c1f4a79961eb647fac0aea && opam update
+RUN cd ~/opam-repository && git pull origin -q master && git reset --hard c731a125bc8f1483494f44b5f9a85be92e022f9d && opam update
 COPY --chown=opam \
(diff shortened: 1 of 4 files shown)


emillon:
If the deployer is confused by `conflicts` we can use a versioned dependency to `carton` as well (that's a bit less semantic since we don't use `carton` directly).

talex5:
You'll need to update the two Dockerfiles to use a more recent opam-repository commit too.

emillon:
- Dockerfile.web was building but I updated it
- I'm not sure what the libffi failure is on `Dockerfile`?
- the alpine build failure seems unrelated?

talex5:
Probably because it's using `opam install` instead of `opam depext` and there's some new OS dependency needed. I think it was that way because `opam depext .` doesn't work. But now the images have opam-2.1 available (but not the default), so perhaps just changing `opam install` to `opam-2.1 install` will help?

The alpine problem will be fixed once https://github.com/ocaml/opam-repository/pull/18712, https://github.com/ocaml/opam-repository/pull/18704 or https://github.com/ocaml/opam-repository/pull/18699 is merged.

emillon:
All good now! It seems weird to mix `opam-2.1` and `opam` but this seems to be working.

talex5:
Thanks!

PR #341, opened 2021-05-24
test all packages with multicore OCaml if they are compatible

avsm:
The ocaml-multicore/multicore-opam repo is now pulled in by default for solving for all packages. This is safe to do since the overlay repository guarantees that it only introduces packages that depend on either `base-domains` or `base-effects`, and thus will not get selected on vanilla compilers.  Any package that does not do this and interferes with vanilla solutions is a bug, and will be fixed promptly in multicore-opam.
    
Any package that is known-broken with multicore can indicate this via a conflict with `base-depends`, and it will not be selected for testing by ocaml-ci

Builds over #333


avsm:
Pushed to live-engine after out-of-band discussion with @talex5

avsm:
reverted hastily due to a bad case of not-working-as-intended ;-)

avsm:
Not the right way to do this with multiple remotes; updated PR coming later.

PR #335, opened 2021-04-12
Move ocaml-version to current master, to get the recent add-multicore change.

ewanmellor:
Move ocaml-version to current master, to get the recent add-multicore change.

Signed-off-by: Ewan Mellor <ewan@tarides.com>

PR #344, opened 2021-06-02
latest ocaml-version: do not test multicore 4.10 any more

avsm:
supersedes #335


talex5:
Did we ever test multicore 4.10 with ocaml-ci? And is it supposed to be testing 4.12+domains now? It doesn't seem to be doing that.

avsm:
right, #345 supersedes all this to just get us to the latest metadata for everything. I'll do the testing multicore 4.12 separately once I merged the metadata to the opam-repository.

ABDUL3693:
> right, #345 supersedes all this to just get us to the latest metadata for everything. I'll do the testing multicore 4.12 separately once I merged the metadata to the opam-repository.

PR #345, opened 2021-06-03
update for ocaml 4.14-dev, latest distro list and bump ci-local versions

avsm:


avsm:
Pushing to live-engine to try it out

PR #347, opened 2021-06-18
Update to latest capnp-rpc

talex5:
Avoids deprecation warning.


talex5:
Also, added work-around for mirage/ocaml-conduit#388.

PR #349, opened 2021-07-02
Add ocaml-doc as user of ocaml-ci

jonludlam:


talex5:
Note: you need to run `make deploy-stack` for the change to take effect (I've done that now).

PR #351, opened 2021-07-06
Test with 4.12+domains too

talex5:


talex5:
Looks like most projects work about as well with this as before:

Screenshot_2021-07-06 Overview - Grafana

Tests using crowbar seem to fail on multicore, though.

PR #354, opened 2021-07-08
[Preview] Use ocamlformat.0.19.0

gpetiot:
This is a preview of the not-yet-released `ocamlformat.0.19.0`, please wait until the package is published in opam to merge this PR. The output is still likely to slightly change before the package is released.

For context, before publishing a new release we open preview pull requests on projects that use ocamlformat to discuss the output and make final adjustments before making the actual release.

I created a `.ocamlformat` file with an arbitrary configuration to try to minimize the diff with the codebase, it could still be tweaked to further reduce the diff.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,0 +1,7 @@
+version = 0.19.0
+profile = conventional
+doc-comments = after-when-possible
+exp-grouping = preserve
+break-cases = fit
+margin = 90
+space-around-lists = false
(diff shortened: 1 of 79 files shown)


talex5:
> For context, before publishing a new release we open preview pull requests on projects that use ocamlformat to discuss the output and make final adjustments before making the actual release.

I think this is an error; this project doesn't use ocamlformat (though it does check it for other projects that do).

gpetiot:
I just opened this PR in case you wanted to format the codebase in the future.

PR #356, opened 2021-07-21
Add grievejia to the allowlist

grievejia:
I understand that this project is experimental and breakages are possible. Wouldn't rely on it for anything critical. Just want to try it on one of my personal project.


avsm:
This is fine by me, as long as you can give us feedback on how it works for you. An issue here with any good/bad experiences would be appreciated.

PR #357, opened 2021-07-22
Make the (lint-fmt) phase depend on the minimal commit to avoid issues when projects fail because they require a newer dune version

kit-ty-kate:
Issue encountered in https://ci.ocamllabs.io/github/ocaml/merlin/commit/122369c769636890ca6b094f4d0595d2522b546b/variant/(lint-fmt)


talex5:
Thanks!

PR #359, opened 2021-07-29
Record build status as CheckRun.

tmcgilchrist:


talex5 (commented):
At the moment, using checks just makes the UI worse: instead of clicking the status link and seeing the list of builds, you click the checks link and go to the checks page, which tells you to click another link to see the list of builds.

I'd like to see it doing something useful before merging it.

tmcgilchrist:
As a solid improvement (over the current state) I'd like to see:

1. The summary landing page moved into the CheckRun tab in GH. eg https://ci.ocamllabs.io/github/ocurrent/ocaml-ci/commit/e3c30a9a2d8bbc0cbdac1deb74e0597ec3a1a22e 
2. Better access control over who can trigger rebuilds

tmcgilchrist:
Screen Shot 2021-08-10 at 10 51 03

tmcgilchrist:
Depends on https://github.com/ocurrent/ocurrent/pull/288 to be merged first and a suitable git submodule dance offered up.

kit-ty-kate:
I'm not sure to understand why this change is being proposed? To me this looks like a downgrade in terms of UI.
Could you give the reasoning behind this?

README.md
```diff
@@ -19,9 +22,9 @@ previously cached build step as long as the opam files don't change.
 To add the CI to your own project:
 
 1. Go to https://github.com/apps/ocaml-ci and install the app for your GitHub user.
-2. Configure just the repositories you want to test (start with one!).
+2. Configure just the repositories you want to test (start with one!). if you select `All Repositories` we won't build anything.
```
talex5, line 25:
```suggestion
2. Configure just the repositories you want to test (start with one!). If you select `All Repositories` we won't build anything. 
```

service/pipeline.ml
```diff
@@ -21,20 +21,41 @@ let platforms =
...
-let github_status_of_state ~head result =
+let github_status_of_state ~head result results =
   let+ head = head
-  and+ result = result in
+  and+ result = result
+  and+ results = results in
   let { Github.Repo_id.owner; name } = Github.Api.Commit.repo_id head in
   let hash = Github.Api.Commit.hash head in
   let url = url ~owner ~name ~hash in
+  let summary = List.fold_left (fun acc (variant, (build, _job_id)) ->
```
talex5, line 39: You can let `Fmt` deal with joining everything together. e.g. something like
```ocaml
let pp_status f build = Fmt.pf f ...

let summary = Fmt.str "@[<v>%a@]" (Fmt.list ~sep:Fmt.cut pp_status) results in
```

README.md
```diff
@@ -121,3 +124,5 @@ $ ocaml-ci mirage/irmin pull/867 alpine-3.10-ocaml-4.08 cancel
 [OCurrent]: https://github.com/ocurrent/ocurrent
 [pipeline.ml]: https://github.com/ocurrent/ocaml-ci/blob/master/service/pipeline.ml
 [capnp-api]: https://github.com/ocurrent/ocaml-ci/blob/master/api/schema.capnp
+
```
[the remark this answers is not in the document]
tmcgilchrist: Yeah my editor keeps adding these, I've fixed the issue.

tmcgilchrist:
Tweaked the UI slightly to 
Screen Shot 2021-08-15 at 12 34 35
Screen Shot 2021-08-15 at 12 34 20


@kit-ty-kate 
> Could you give the reasoning behind this?

I'm trying achieve 2 things here:
 1. Build out ocurrent to support GH Checks API. In future we could use this over recording build information as Commit status and comments on a PR (as Camelus used to do with it's checks). Once we start using Checks then we could use things like [AutoMerge](https://docs.github.com/en/github/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request) once builds have passed or trigger workflows based of the outcome of previous checks.
 2. Demonstrate a pipeline that uses 1, which is a work in progress here and would benefit from feedback on what works or doesn't work. I understand that the GH CI eg https://github.com/tmcgilchrist/ocaml-aws/runs/3331454823 uses Checks so it should be possible to achieve a similar UI (or better) than what they have.

service/pipeline.ml
```diff
@@ -21,20 +21,47 @@ let platforms =
...
-let github_status_of_state ~head result =
+let github_status_of_state ~head result results =
   let+ head = head
-  and+ result = result in
+  and+ result = result
+  and+ results = results in
   let { Github.Repo_id.owner; name } = Github.Api.Commit.repo_id head in
   let hash = Github.Api.Commit.hash head in
   let url = url ~owner ~name ~hash in
+  let variant_compare variant1 variant2 = match (variant1, variant2) with
```
talex5, line 39: I don't think you need this. The job label should be `(analysis)`, which will already sort first because `(` comes before the letters.

talex5 (approved):
Looks right, but maybe some left-over code in the sorting?

service/pipeline.ml
tmcgilchrist: Split this up into the suggested functions.

talex5:
Deployed for testing.

I wonder if this will create a lot of extra calls to GitHub? Before, we only updated the status once to go to pending and then once at the end to go to complete. Now we have to update every time the markdown status changes...

tmcgilchrist:
Bonus feature in this change, the GH UI updates inline when you get a build going from InProgress to Green/Red.

PR #360, opened 2021-07-30
Test s390x fix for ocaml-git.

tmcgilchrist:


talex5:
ocaml-ci doesn't support pin-depends in vendored libraries, but I tested this manually (by removing the version constraints on these packages from the generated Dockerfile) and the tests passed.

dinosaure:
Thanks, I will try to do a release next week.

PR #362, opened 2021-08-31
Report state on a Matrix server

TheLortex:
This PR adds a way for the CI to publish its status on a Matrix server using a bot account. 

The credentials are passed using command line arguments: `--matrix-user`, `--matrix-host`, `--matrix-passfile`. If one of the arguments is not available, then the Matrix publication is disabled.

For each project, a read-only room is created: for example `#ocurrent/ocaml-ci:matrix.ocamllabs.io`. So as a warning, **this bot will create many rooms on the server**, maybe we should use a separate server for that purpose.

The message is styled using a subset of HTML, and currently looks like this:
Screenshot 2021-08-31 at 12-34-56 Element TheLortex ocurrent

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,6 @@
 [submodule "ocaml-dockerfile"]
 	path = ocaml-dockerfile
 	url = https://github.com/avsm/ocaml-dockerfile.git
+[submodule "ocaml-matrix"]
+	path = ocaml-matrix
+	url = https://github.com/clecat/ocaml-matrix.git
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,12 @@ COPY --chown=opam \
 COPY --chown=opam \
 	ocaml-dockerfile/dockerfile*.opam \
 	/src/ocaml-dockerfile/
+COPY --chown=opam \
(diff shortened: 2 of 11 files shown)


avsm:
Tested manually on ocamllabs.io's matrix server, and working well! 
image

avsm:
I'll push this to `live` shortly, but would really like the vendored matrix libraries to be released so we don't have a bunch of git submodules in the longer term. What's the release status of that to opam-repo @TheLortex @clecat?

TheLortex:
For the ocurrent plugin, it's necessary to vendor it as we have the following dependency path: `ocaml-ci` -> `matrix-current` -> `current`, and `current` is vendored.

TheLortex:
Some news on the PR, on monday we tested it by changing the `live` branch and it worked well until it encountered rate limit errors from the server. So now changes include:
- support for rate limit errors, the request is re-issued if the previous response is a 429 Too Many Requests
- add organisation rooms along with single-project rooms
- rooms are not read-only any more, and a fixed list of admins is added on setup (for now it's only @avsm)

avsm:
re-pushed to live-engine

avsm:
now live!

PR #368, opened 2021-09-22
Constrain obuilder-spec to 0.3

MisterDA:
There'll be some changes to obuilder-spec with the Docker backend. As it's not vendored in this repo, add a version constraint to the current released version.


tmcgilchrist:
Sounds reasonable, can this be merged @MisterDA

PR #369, opened 2021-09-22
services: start the solver in main term

MisterDA:
If the solver starts uncoditionnally in the top-level, then

    dune exec -- ocaml-ci-local --help

would also start the solver in the background.

### DIFF
--- a/service/local.ml
+++ b/service/local.ml
@@ -1,13 +1,12 @@
 (* Utility program for testing the CI pipeline on a local repository. *)
 
-let solver = Ocaml_ci.Solver_pool.spawn_local ()
-
 let () =
   Unix.putenv "DOCKER_BUILDKIT" "1";
   Unix.putenv "PROGRESS_NO_TRUNC" "1";
   Prometheus_unix.Logging.init ()
 
 let main config mode repo : ('a, [`Msg of string]) result =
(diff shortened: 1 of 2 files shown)


service/local.ml
```diff
@@ -1,13 +1,12 @@
 (* Utility program for testing the CI pipeline on a local repository. *)
 
-let solver = Ocaml_ci.Solver_pool.spawn_local ()
-
 let () =
   Unix.putenv "DOCKER_BUILDKIT" "1";
   Unix.putenv "PROGRESS_NO_TRUNC" "1";
   Prometheus_unix.Logging.init ()
```
dra27, line 6: Can't this go into main too?

service/main.ml
```diff
@@ -40,8 +40,6 @@ module Metrics = struct
     Gauge.set (master "active") (float_of_int active)
 end
 
-let solver = Ocaml_ci.Solver_pool.spawn_local ()
-
 let () =
```
dra27, line 43: Same here - either shift it into `main` or put it into a function which is called from main?

dra27 (commented):
Looks fine - can possibly go the whole hog and move the other initialisation parts into `main` as well?

dra27:
The changes I request are part of #370, so this is good to go!

PR #379, opened 2021-10-19
Disable odoc lint "warn-error" flag

jonludlam:
Temporarily until odoc becomes a bit more helpful and lenient.


samoht:
LGTM :-) (this will avoid having non-fixable errors in the irmin and mirage CI)

PR #380, opened 2021-10-20
Monorepo: use opam monorepo depext command

emillon:
This command is available since 0.2.6 and can replace the workaround we are using.


tmcgilchrist:
Looks good.

PR #386, opened 2021-11-03
Add verbosemode as user of ocaml-ci

verbosemode:
I'd like to add my GH user to run the CI for the OCaml [syslog-message](https://github.com/verbosemode/syslog-message/) repo.


talex5:
Should be building now!

PR #399, opened 2021-12-22
Add dependency installation steps to README

shakthimaan:
For testing a local project, `opam update` and the dependency installation steps are required. The same have been added to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Note: you need to clone with `--recursive` because this project uses submodules
 If you forget, `git submodule update --init` will fetch them.
 
 To test the CI on a local Git clone, you need to first install the
-dependencies, and then use `dune exec` as shown below:
+dependencies and then use `dune exec` as shown below:
 
 ```sh
 opam update
--- a/README.md
+++ b/README.md
@@ -39,9 +39,12 @@ Note: you need to clone with `--recursive` because this project uses submodules
 (it depends on some packages that aren't released yet).
 If you forget, `git submodule update --init` will fetch them.
 
-To test the CI on a local Git clone, use:
+To test the CI on a local Git clone, you need to first install the
+dependencies, and then use `dune exec` as shown below:
 
 ```sh
+opam update
+opam install -t .
 dune exec -- ocaml-ci-local /path/to/project
 ```
 
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ dependencies and then use `dune exec` as shown below:
 
 ```sh
 opam update
-opam install -t .
+opam install . --deps-only --with-test --yes
 dune exec -- ocaml-ci-local /path/to/project
 ```
 


talex5:
Can we just link to some generic instructions, to avoid duplicating this in every project? @christinerose might know a suitable link.

i.e. something that explains:

- How to install opam.
- How to git clone a project and install its dependencies.
- What to change depending on your version of opam.

shakthimaan:
* I was looking at the README as a quick start guide to know the required steps to get from point A to point B. A short OPAM tutorial will be nice to have though.
* I usually do not run `opam update` to avoid installing latest and greatest packages. If this is required for installation, it is good to know.

christinerose:
Here's a link for "How to Install Opam": https://opam.ocaml.org/doc/Install.html and more info here: https://opam.ocaml.org/doc/Usage.html, but I don't see anything about `git clone` in there.
I'm also working on docs/tutorials for the new v3 site, so perhaps we can create something specific for Opam there. (??)

talex5:
@christinerose yes, I'm thinking of a page to cover the common situation of a user who wants to try out a program they found in a repository, but doesn't necessarily know (or want to learn) OCaml or opam.

It would show roughly the same steps you'd get on a project's web-page, but with lots of trouble-shooting hints when there are problems. e.g.

1. Make you have opam installed (show how to check, link to opam installation instructions, suggest minimum version of maybe 2.1)
2. git clone repo
3. cd repo
4. opam install .

Or, to test it locally without installing:

4. opam install --deps-only -t .
5. dune exec -- prog.exe

Explain what to do if opam is too old, opam switch is too old, dependencies need updating, etc.

Then each project can just provide a couple of lines for experienced users (like this one does) plus a link to the page for people who are stuck.

shakthimaan:
> I'm thinking of a page to cover the common situation of a user 
An OCaml FAQ will be nice to have!

README.md
```diff
@@ -39,9 +39,12 @@ Note: you need to clone with `--recursive` because this project uses submodules
 (it depends on some packages that aren't released yet).
 If you forget, `git submodule update --init` will fetch them.
 
-To test the CI on a local Git clone, use:
+To test the CI on a local Git clone, you need to first install the
+dependencies, and then use `dune exec` as shown below:
 
 ```sh
+opam update
+opam install -t .
```
tmcgilchrist, line 47:
```suggestion
opam install . --deps-only --with-test --yes
```
This is a better combination of flags.

tmcgilchrist:
@talex5 absolutely agree with the common instructions. 
@christinerose I can help with providing the short version of what is required, ping me on slack.

maiste:
As it would be moved outside of this repository, should we consider closing this PR?

README.md
```diff
@@ -39,9 +39,12 @@ Note: you need to clone with `--recursive` because this project uses submodules
 (it depends on some packages that aren't released yet).
 If you forget, `git submodule update --init` will fetch them.
 
-To test the CI on a local Git clone, use:
+To test the CI on a local Git clone, you need to first install the
+dependencies, and then use `dune exec` as shown below:
```
christinerose, line 43:
```suggestion
dependencies and then use `dune exec` as shown below:
```

christinerose (approved):
Just one unnecessary comma :)

PR #402, opened 2022-01-07
Support opam-version variable

emillon:
This adds `opam_version` to the set of worker vars. It is parsed from the opam template `%{opam-version}`. Special support is taken when creating the 0install context because the function `std_env` expands this variable to the version of the opam library we are linking against. This does seem to correspond to what we want here.


emillon:
cc @kit-ty-kate

emillon:
Deploying this invalidates all the `opam-vars` nodes. Is it ok?

kit-ty-kate:
> Deploying this invalidates all the `opam-vars` nodes. Is it ok?

I suspect that trying to fix `opam-0install` directly instead might avoid this

emillon:
Hmm, I don't think so. I think that we still need to shell out to the opam binary in the image to ask what its variables are. There's a new variable that's being queried, so the json schema output by `opam-vars` changes. This is the bit that invalidates the nodes.

There's something to fix in `opam-0install-solver` but it will require passing `opam-version` to `Dir_context.std_env` and we still have to fetch that value by calling the opam binary.

talex5 (approved):
Looks good to me. Adding an optional parameter to `std_env` would be nice too, but this seems fine for now.

emillon:
Thanks! I added a pointer to https://github.com/ocaml-opam/opam-0install-solver/pull/36

talex5:
I think the `.gitmodules` file needs updating to fix CI (replace `git:` with `https:`).

PR #404, opened 2022-01-11
Use https protocol for submodules

emillon:
The git protocol is not supported anymore. See https://github.blog/2021-09-01-improving-git-protocol-security-github/

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,7 +8,7 @@
 	branch = master
 [submodule "ocaml-version"]
 	path = ocaml-version
-	url = git://github.com/ocurrent/ocaml-version.git
+	url = https://github.com/ocurrent/ocaml-version.git
 	branch = master
 [submodule "ocaml-dockerfile"]
 	path = ocaml-dockerfile


tmcgilchrist:
Looks good. Thought we got all of these last year.

dra27:
It seems to have revealed a bug in ocaml-ci, though - it looks like the Analyse does not realise that it needs to refresh the submodule configuration?

kit-ty-kate:
github put git:// back last time after a couple of days or so. They just removed it again for some reason

emillon:
I think it's a brownout. They're planning to remove it, so pushing errors so that we get a chance to fix it now :)

emillon:
> It seems to have revealed a bug in ocaml-ci, though - it looks like the Analyse does not realise that it needs to refresh the submodule configuration?

Yes, this seems to be an issue with how current_git operates. It does `git clone --recursive "https://github.com/ocurrent/ocaml-ci.git"`, but the URL in `.gitmodules` in the default branch is invalid.

I'm not sure this is a big problem in practice (I think that this PR can be merged and it will work).

PR #405, opened 2022-01-12
Remove use of git: protocol in .gitmodules

talex5:
GitHub no longer supports it.


dra27:
Duplicates #404, but this one passes ocaml-ci!

talex5:
Oops! #404 seems to be passing at the moment, so merged that.

Issue #412, opened 2022-01-30
Support for opam 2.1


samoht:
It would be nice if ocaml-ci could test builds with opam 2.1 as this would allow testing beta releases a bit more thoroughly. Right now the recommend way to add beta package to opam-repository is to use `available: opam-version >= "2.1"` but then this leads to `ocaml-ci` complaining with `Availability condition not satisfied`.

dinosaure:
From @talex5:
> it would be better to treat available: opam-version >= "2.1" as a constraint for the solver to satisfy, rather than taking the opam version as an input and rejecting anything that doesn't satisfy it.

dinosaure:
Unfortunately, the problem seems big for certain packages: https://github.com/mirage/ocaml-vchan/pull/138

samoht:
The failing build is still using `opam-2.0` ; not sure what happened there...

kit-ty-kate:
The `(analyis)` steps correctly detected that the package is only available with opam 2.1:
```
2022-02-02 17:11.28 [INFO] debian-11-4.14: eliminated all possibilities in 0.66 s
2022-02-02 17:11.28 [INFO] debian-11-4.11: eliminated all possibilities in 0.94 s
2022-02-02 17:11.28 [INFO] debian-11-4.12: eliminated all possibilities in 0.94 s
2022-02-02 17:11.28 [INFO] debian-11-4.06: eliminated all possibilities in 0.81 s
2022-02-02 17:11.28 [INFO] debian-11-4.13+flambda: eliminated all possibilities in 0.68 s
2022-02-02 17:11.28 [INFO] debian-11-4.09: eliminated all possibilities in 1.00 s
2022-02-02 17:11.28 [INFO] debian-11-4.08: eliminated all possibilities in 1.10 s
2022-02-02 17:11.28 [INFO] debian-11-4.07: eliminated all possibilities in 0.96 s
2022-02-02 17:11.28 [INFO] debian-11-4.13: eliminated all possibilities in 1.37 s
2022-02-02 17:11.28 [INFO] debian-11-4.10: eliminated all possibilities in 1.12 s
2022-02-02 17:11.28 [INFO] debian-11-4.04: eliminated all possibilities in 1.00 s
2022-02-02 17:11.28 [INFO] debian-11-4.03: eliminated all possibilities in 1.09 s
2022-02-02 17:11.28 [INFO] debian-11-4.12+domains: eliminated all possibilities in 0.82 s
[...]

PR #413, opened 2022-01-30
Add opam-2.1 builds

samoht:
To avoid the multiplication of build targets, only build using 2.1 if
that's the only supported target, otherwise keep using 2.0.

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -59,7 +59,7 @@ let rec get_root_opam_packages = function
 
 let download_cache = "opam-archives"
 
-let install_project_deps ~opam_files ~selection =
+let install_project_deps ~opam_version ~opam_files ~selection =
   let { Selection.packages; commit; variant } = selection in
   let groups = group_opam_files opam_files in
   let root_pkgs = get_root_opam_packages groups in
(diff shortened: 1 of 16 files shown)


lib/opam_build.ml
```diff
@@ -76,26 +76,33 @@ let install_project_deps ~opam_files ~selection =
   let network = ["host"] in
   (if Variant.arch variant |> Ocaml_version.arch_is_32bit then
      [shell ["/usr/bin/linux32"; "/bin/sh"; "-c"]] else [])
-  @ distro_extras @ [
+  @ distro_extras
+  @ match opam_version with
+    | `V2_0 -> []
```
kit-ty-kate, line 81: To keep compatibility in case the base images upgrade to opam 2.1 by default it is better to do it anyway in any case:
```suggestion
    | `V2_0 -> [ run "sudo ln -f /usr/bin/opam-2.0 /usr/bin/opam" ]
```

samoht:
This is needed to use ocaml-ci on mirage repository which are depending on the MirageOS beta releases (like https://github.com/mirage/mirage-solo5/pull/84).

I have checked that `dune exec -- ocaml-ci-local ~/git/mirage-solo5` works properly.

lib/opam_version.ml
```diff
@@ -0,0 +1,13 @@
...
+let to_string_with_minor t = to_string t ^ ".0"
+
+let pp = Fmt.of_to_string to_string
+
+let default = `V2_0
+
+let of_string str = match String.sub str 0 3 with
+  | "2.0" -> Ok `V2_0
+  | "2.1" -> Ok `V2_1
+  | _ | exception Invalid_argument _ -> Error (`Msg "invalid opam version")
```
kit-ty-kate, line 13:
```suggestion
let of_string = function
  | "2.0" -> Ok `V2_0
  | "2.1" -> Ok `V2_1
  | _ -> Error (`Msg "invalid opam version")
```

kit-ty-kate (commented):
Looks good at a glance. Just a few general comments.

I’ll leave the rest to the maintainers

samoht:
That should fix #412

lib/opam_build.ml
```diff
@@ -74,28 +74,33 @@ let install_project_deps ~opam_files ~selection =
...
-    env "DEPS" (String.concat " " non_root_pkgs);
-    run ~network ~cache "opam depext --update -y %s $DEPS" (String.concat " " root_pkgs);
-    run ~network ~cache "opam install $DEPS"
-  ]
+       && opam update -u" commit commit ]
+  @ pin_opam_files ~network groups
+  @ [ env "DEPS" non_root_pkgs ]
+  @ match opam_version with
+    | `V2_0 -> [run ~network ~cache "opam depext --update -y %s $DEPS" root_pkgs]
+    | `V2_1 -> []
```
talex5, line 95: How do the depexts of the root packages get installed in the new scheme?

dra27: Automatically as part of `opam install`

dra27:
Semi-meta comment: if CLI versioning is used, this doesn't need to be about testing opam 2.1, but about testing "not opam 2.0". If the commands have `--cli=2.1` added whenever ocaml-ci knows that the binary is opam 2.1+ then it should carry on working even with the opam 2.2 binary, etc.

lib/opam_build.ml
talex5: But the `opam install` below doesn't install `root_pkgs` (only `$DEPS`).

samoht:
I've modified the depext command for 2.1 to add `--cli=2.1` and install what is required by the `root_pkgs` too.

samoht:
I've squashed the commits, should be ready to be merged and deployed!

dinosaure:
Is it possible to merge this PR as soon as we can to help us to unlock the release process of MirageOS 4.0?

talex5:
This seems pretty inefficient. As I understand it, it solves everything twice and then throws away half the results (doubling the load on the solver).

I guess it's OK for now if we're in a hurry, but it would be better to treat `available: opam-version >= "2.1"` as a constraint for the solver to satisfy, rather than taking the opam version as an input and rejecting anything that doesn't satisfy it.

PR #416, opened 2022-02-03
Attempt to fix new variant ID parsing

talex5:
The parsing code didn't recognise the format, but instead of reporting an error it just guessed. This PR is a minimal fix to test the change.

### DIFF
--- a/lib/variant.ml
+++ b/lib/variant.ml
@@ -72,7 +72,7 @@ let of_string s =
    | None -> s, `X86_64, Opam_version.default
    | Some (s, a) ->
       let arch a = Ocaml_version.of_opam_arch a |> Option.value ~default:`X86_64 in
-      match Astring.String.cut ~rev:true ~sep:"_opam-" a with
+      match Astring.String.cut ~rev:true ~sep:"opam-" a with
       | None -> s, arch a, Opam_version.default
       | Some (a, v) ->
          match Opam_version.of_string v with


samoht:
I don't understand how these strings are generated at the first. The `pp` function above is printing `_opam-` in the label, why `of_string` cannot read those?

talex5:
You first split on `_` to get `debian-11-4.13` and `opam-2.1`. Then you try to split the second part on `_opam-`. How did this ever work?

samoht:
This function is indeed completely buggy...

samoht:
It should fist split on `_opam` and then split the head ID...

lib/variant.ml
```diff
@@ -72,7 +72,7 @@ let of_string s =
    | None -> s, `X86_64, Opam_version.default
    | Some (s, a) ->
       let arch a = Ocaml_version.of_opam_arch a |> Option.value ~default:`X86_64 in
-      match Astring.String.cut ~rev:true ~sep:"_opam-" a with
+      match Astring.String.cut ~rev:true ~sep:"opam-" a with
```
kit-ty-kate, line 75: The current fix only works for x86_64 variants.

I think the whole function would be a lot clearer if `String.split_on_char ’_’` was used instead.
Something like:
```
match String.split_on_char ’_’ s with
| [s] -> s, `X86_64, Opam_version.default
| [s; v] when String.is_prefix "opam-" v -> s, x86_64, Opam_version.of_string v
| [s; a] -> s, arch a, Opam_version.default
| [s; a; v] when String.is_prefix "opam-" v -> s, arch a, Opam_version.of_string v
```

samoht:
ok let me try to fix my mess - I'll also add a few more tests 😓

samoht:
Tentative cleanup here: #417

PR #419, opened 2022-02-15
Update dependencies

MisterDA:
https://github.com/avsm/ocaml-dockerfile/pull/85 introduced a new `Dockerfile_distro.distro` type, but `Dockerfile_distro.tag_of_distro` and `Dockerfile_distro.latest_tag_of_distro` still take `Dockerfile_distro.t`. Resolving the alias later should not change the outcome.


MisterDA:
Supersedes #429.

PR #427, opened 2022-02-23
Disable opam-dune-lint.

tmcgilchrist:


talex5:
Note that before we had opam-dune-lint, this job used to just run `opam lint`. Might be worth restoring that behaviour instread of removing it completely.

tmcgilchrist:
Good idea, I'll switch over to `dune lint` until we have a working solution for `opam-dune-lint`.

tmcgilchrist:
https://github.com/ocurrent/ocaml-ci/pull/428

PR #431, opened 2022-03-10
Add hyper-systems as user for ocaml-ci

clecat:


clecat:
My apologies, I forgot to update the content of my fork, it should be good to go now

PR #433, opened 2022-03-16
Remove hxd pin, update git-unix pin

MisterDA:
git-unix merged support for cmdliner.1.1.0, so we can remove the hxd pin and pin git-unix to the current master.


MisterDA:
Superseded by #419.

PR #442, opened 2022-04-14
Implement avoid-version in ocaml-ci-solver and fix ocaml-beta support

dra27:
`ocaml-src.4.14.dev` reveals two shortcomings of ocaml-ci's solver which are fixed in this PR:
1. The solver ignores the `avoid-version` flag, which means that it selects `ocaml-src.4.14.dev` rather than `ocaml-src.4.14.0`. The fix is simple (and due to @kit-ty-kate) - simply change the priority order of the candidates so that the ones flagged `avoid-version` appear last, then they're only selected if absolutely necessary.
2. If the `ocaml-beta` package is detected in a solution, previously ocaml-ci-solver always added a fake `ocaml-beta.enabled` package to the universe, but this relies on the underlying docker image having the beta remote enabled. This is only done for the trunk package and for alpha/beta/rc releases of the compiler, so `ocaml-src.4.14.dev` being selected was breaking opam 2.0 builds. Now `ocaml-beta.enabled` is only added if we know the underlying compiler will support it. As it happens, this fix is masked by the `avoid-version` support.

### DIFF
--- a/solver/dune
+++ b/solver/dune
@@ -3,4 +3,4 @@
   (public_name ocaml-ci-solver)
   (package ocaml-ci-solver)
   (preprocess (pps ppx_deriving.std ppx_deriving_yojson))
-  (libraries lwt.unix ocaml-ci-api ppx_deriving_yojson.runtime opam-0install capnp-rpc-unix git-unix logs.cli))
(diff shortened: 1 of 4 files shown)


dra27:
This is presently on live-engine and seems to be behaving itself!

talex5 (approved):

PR #445, opened 2022-04-15
Support for upcoming opam-monorepo release 0.3

Leonidas-from-XIV:
The opam-monorepo developers are planning to release 0.3.0 soon, as such it seems reasonable to add support for it in ocaml-ci.

This PR consists of two parts:

1. Adding support for the `0.3` lockfile format and selecting the right version constraint (opam-monorepo 0.3.0 will not support the `0.2` format due to changes how dependencies are encoded - the migration is trivial and the tool notifies the user in that case)
2. Add support for installing dependencies that are supposed to be installed via `opam` which is a new feature in opam-monorepo starting at 0.3.0

### DIFF
--- a/lib/opam_monorepo.ml
+++ b/lib/opam_monorepo.ml
@@ -184,7 +184,7 @@ let install_opam_provided_packages ~network ~cache ~lock_file_path ~lock_file_ve
   | V0_1 | V0_2 -> []
   | V0_3 ->
       [
-        run ~network ~cache "opam install --yes --deps-only ./%s" lock_file_path;
+        run ~network ~cache "opam install --yes --ignore-pin-depends --deps-only ./%s" lock_file_path;
       ]
 
 let spec ~base ~repo ~config ~variant =
--- a/lib/opam_monorepo.ml
+++ b/lib/opam_monorepo.ml
@@ -187,15 +187,6 @@ let install_opam_provided_packages ~network ~cache ~lock_file_path ~lock_file_ve
         run ~network ~cache "opam install --yes --ignore-pin-depends --deps-only ./%s" lock_file_path;
       ]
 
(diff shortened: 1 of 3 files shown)


lib/opam_monorepo.ml
```diff
@@ -176,6 +178,15 @@ let initialize_switch ~network = function
       let open Obuilder_spec in
       [ run ~network "opam switch create %s" compiler_package ]
 
+let install_opam_provided_packages ~network ~cache ~lock_file_path ~lock_file_version =
+  let open Obuilder_spec in
+  match lock_file_version with
+  | V0_1 | V0_2 -> []
+  | V0_3 ->
+      [
+        run ~network ~cache "opam install --yes --deps-only ./%s" lock_file_path;
```
NathanReb, line 187: The correct invocation should contain `--ignore-pin-depends`. Might be worth adding `--no-lint` to get rid of the output polution.

Leonidas-from-XIV: Good point ­— added; my copy of opam 2.1.1 does not know `--no-lint` however so I left it out.

tmcgilchrist (commented):
Looks good. 
I can do a test deploy if you have a project or two that use this new version @Leonidas-from-XIV

Leonidas-from-XIV:
That would be great because indeed we would like to try it first with opam-monorepo where we have a branch that uses its new version.

On April 18, 2022 6:29:15 AM GMT+02:00, Tim McGilchrist ***@***.***> wrote:
***@***.*** commented on this pull request.

Leonidas-from-XIV:
Since opam-monorepo 0.3.0 was released and [merged to opam-repository](https://github.com/ocaml/opam-repository/pull/21264) I removed the pins from this PR. I think it should be ready to go.

PR #447, opened 2022-04-26
For each variant, only build compatible packages

### LINKED ISSUE #297: Test packages on the full range of supported OCaml versions

talex5:
This checks each local .opam file for an explicit constraint on the version of `ocaml`. If it doesn't match the platform, the package is skipped.

This is useful when different packages in a repository work on different versions of OCaml. For example, `prometheus` works on OCaml 4.03, but `prometheus-app` requires 4.08. Previously, this prevented testing on 4.03 at all.

The opam build uses `dune --only-packages` to build just the compatible subset.

Fixes #297. I tried testing this on the [prometheus](https://github.com/mirage/prometheus) repository, but hit https://github.com/ocaml/dune/issues/5621, so I don't know how easy it will be to support this.

It should be fairly easy to extend this to match `base-domains` too, which would allow merging eio support packages in existing libraries without preventing testing the other packages on older compilers (/cc @bikallem).

### DIFF
--- a/api/worker.ml
+++ b/api/worker.ml
@@ -18,6 +18,7 @@ end
 module Selection = struct
   type t = {
     id : string;                        (** The platform ID from the request. *)
+    compat_pkgs : string list;          (** Local root packages compatible with the platform. *)
     packages : string list;             (** The selected packages ("name.version"). *)
     commit : string;                    (** A commit in opam-repository to use. *)
   } [@@deriving yojson, ord]
--- a/lib/analyse.ml
+++ b/lib/analyse.ml
@@ -158,9 +158,20 @@ module Analysis = struct
     in
     Capnp_rpc_lwt.Capability.with_ref (job_log job) @@ fun log ->
     Ocaml_ci_api.Solver.solve solver request ~log >|= function
-    | Ok [] -> Fmt.error_msg "No solution found for any supported platform"
-    | Ok x -> Ok (List.map Selection.of_worker x)
     | Error (`Msg msg) -> Fmt.error_msg "Error from solver: %s" msg
+    | Ok [] -> Fmt.error_msg "No solution found for any supported platform"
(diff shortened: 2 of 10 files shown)


talex5:
Pushed live for testing.

PR #448, opened 2022-05-03
Add maiste to OCaml-CI user

maiste:


mtelvers:
To start building with OCaml-CI:

- Add the app to your account, selecting Install at https://github.com/apps/ocaml-ci
- Select only the repositories you want tested - if you select All Repositories we won't build anything!

maiste:
Thanks for the merge and for your explanations :+1:

PR #465, opened 2022-05-18
Make GitHub settings in dev.md match the ones actually in use

talex5:
I'm not sure where the ones in the documentation came from, but they don't match the settings for the real deployment, which seems wrong.

/cc @maiste


maiste:
@talex5, I don't have the possibility to subscribe to the events `Create` and `Push`... What about you @novemberkilo?

Screenshot:
image

talex5:
Did you select all the permissions in the previous section? e.g. the `Create` option appears after selecting `Contents: read-only`.

tmcgilchrist (approved):
Thanks @talex5 this matches what I can see for the apps and my own GH App.

maiste (approved):
I fixed my GithubApp with all your remarks. Thanks a lot for the correction!

Issue #381, opened 2021-10-20
CI is broken with non-trivial dependencies


hannesm:
Hello,

thank you very much for the ocaml-ci. We just encountered an interesting corner case where it does not work: https://ci.ocamllabs.io/github/roburio/happy-eyeballs/commit/4a91542ac035eca1590932c9acb2434875a058f9/variant/alpine-3.13-4.13

The repository happy-eyeballs consists of 3 opam packages: happy-eyeballs, happy-eyeballs-lwt, happy-eyeballs-mirage. The happy-eyeballs package provides a pure implementation of the algorithm/RFC, which is used by dns-client (to first attempt DNS-over-TLS, then DNS-over-TCP - with all nameservers from /etc/resolv.conf). The happy-eyeballs-lwt and happy-eyeballs-mirage depend on dns-client for DNS resolution.

Now the CI fails with the error `Error: Library "happy-eyeballs" not found.` (while compiling dns-client), though opam reports that `installed happy-eyeballs.dev` in its output. A manual installation works fine as well (first happy-eyeballs, then dns-client, then happy-eyeballs-lwt/happy-eyeballs-mirage).

Thanks for reading, and maybe there's something deeper to look into how package installations are scheduled?

Best,

Hannes

hannesm:
I ran into this issue again at https://github.com/mirage/metrics/pull/53 -- are there any plans to implement a fix, like the one suggested by @dra27?

hannesm:
Hi again, I'm still curious about this issue - and was wondering whether fixing this issue is on the roadmap? It looks like @dra27 has a good plan about how to fix this.

maiste:
Thanks for the quick reply! I agree with you. 
1. Indeed, this one seems to be a reasonable trade. As it depends on the _repo_ construction, it should be the responsibility of the developers to handle this case and not the CI.
2. Agree with that too! In this case, it is not only related to `happyeyeballs` but also to other packages. It's linked with this [dune issue](https://github.com/ocaml/dune/issues/5621) open by @talex5 which makes it hard to build examples without having the same issue in the CI. In the case of `happyeyeballs`, I image `dune build @all` work if you pin the right dependencies or vendors `dns-client` (as you suggested).

dra27:
[...]

PR #467, opened 2022-05-25
Fix issue ocaml-ci#381

maiste:
Hi,
with @art-w we have been working on a fix for #381. The idea is to decouple the `depext` and the `opam deps` installation by unpinning the elements, installing the opam version and finally reinstalling the version from the opam files to only rebuild the related elements.  It also relies on the `--only-packages` flag from `dune` to build one package at a time.

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -123,10 +123,13 @@ let spec ~base ~opam_version ~opam_files ~selection =
   let groups = group_opam_files opam_files in
   let root_pkgs = get_root_opam_packages groups in
   let compatible_root_pkgs = if only_packages = [] then root_pkgs else only_packages in
+  let compatible_root_opam =
+      List.map (fun x -> "./" ^ (Filename.basename x |> Filename.chop_extension) ^ ".opam") compatible_root_pkgs
+  in 
   stage ~from:base (
     comment "%s" (Fmt.str "%a" Variant.pp selection.Selection.variant) ::
     user ~uid:1000 ~gid:1000 ::
-    copy compatible_root_pkgs ~dst:"/src/" ::
+    copy compatible_root_opam ~dst:"/src/" ::
     install_project_deps ~opam_version ~opam_files ~selection @
     build_each_package ~opam_files:compatible_root_pkgs
   )
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -51,11 +51,26 @@ let pin_opam_files ~network groups =
       |> run ~network "%s"
     ]
 
+let unpin_opam_files ~network opam_files =
+    let pkgs = List.map (fun x -> Filename.basename x |> Filename.chop_extension) opam_files in
+    let pkgs = String.concat " " pkgs  in
+    Obuilder_spec.run ~network "opam pin remove -y %s" pkgs
+
 (* Get the packages directly in "." *)
 let rec get_root_opam_packages = function
   | [] -> []
   | (dir, _, pkgs) ::_ when Fpath.is_current_dir dir -> pkgs
   | _ :: rest -> get_root_opam_packages rest
+
+ let build_each_package ~opam_files =
+     let open Obuilder_spec in
+    List.map
+        (fun x ->
+            Filename.basename x
+            |> Filename.chop_extension
+            |> run "opam exec -- dune build --only-packages=%s @install @check @runtest && rm -rf _build")
+        opam_files
+
 
 let download_cache = "opam-archives"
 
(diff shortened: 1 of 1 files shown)


lib/opam_build.ml
```diff
@@ -95,15 +100,25 @@ let install_project_deps ~opam_version ~opam_files ~selection =
...
+        opam_files
+  in
   stage ~from:base (
     comment "%s" (Fmt.str "%a" Variant.pp selection.Selection.variant) ::
     user ~uid:1000 ~gid:1000 ::
-    install_project_deps ~opam_version ~opam_files ~selection @ [
-      copy ["."] ~dst:"/src/";
-      run "opam exec -- dune build @install @check @runtest && rm -rf _build"
-    ]
+    copy ["."] ~dst:"/src/" ::
```
dra27, line 121: Doing this before deps installation is going to hurt caching, I think? ocaml-ci's assumption is that dependencies change less frequently, so it's trying to install dependencies first so that _different PRs_ which change code, but not the dependency cone, share the deps.

lib/opam_build.ml
```diff
@@ -95,15 +100,25 @@ let install_project_deps ~opam_version ~opam_files ~selection =
...
 
 let spec ~base ~opam_version ~opam_files ~selection =
   let open Obuilder_spec in
+  let build_each_package =
+    List.map
+        (fun x ->
+            Filename.basename x
+            |> Filename.chop_extension
+            |> run "opam exec -- dune build --only-packages=%s @install @check @runtest && rm -rf _build")
+        opam_files
```
dra27, line 116: Doesn't this mean that we'd _always_ split multi-package repos, even when it's not necessary? That's going to slow down builds, as we'll lose parallelism?

dra27 (commented):
It's great for this to be being worked on, thanks! It's interesting to tackle this at the build stage, but I think - as in the original issue - that this needs to be tackled at the solving stage.

My original idea was - unlike here - rather than _installing_ each package individual that we'd _solve_ each package installation individually. I don't think it's necessary to know the dependency _graph_ within each solution - it should be enough to look just at the package each solution has in common (with some possible opam-lib trickery needed to get depopts, etc. - I can't remember what analysis we already have in ocaml-ci).

The original example in https://github.com/ocurrent/ocaml-ci/issues/381#issuecomment-953811239 was derived by hand looking at the solutions for those individual packages.

lib/opam_build.ml
maiste: You are entirely correct. How can we detect the cases that need to run this way and the other ones? In a perfect solution, we would be able to see the interdependency between multiple repositories and run this way only when we fall into this case.

maiste: Ok, it's more straightforward for me! Thanks for the clarification :)
I don't know if it is an issue, but how should the solver know that `happy-eyeballs-lwt` requires the local version of `happy-eyeballs` and not the one from `opam`?
I'm rewriting the solution without the caching issue order so we have the two solutions available.

art-w:
Hello!

Thanks @dra27 for the quick review and catching the misplaced `COPY .` :) I pushed @maiste not to do the analysis as it sounded much harder to get right... and I thought he could get away without:

- **install the depexts** For this we need to copy only the `*.opam` files, pin them (but no install since we don't have the sources), install their depexts, then unpin.  (installing the depext as required by the repo `*.opam` allow for the optional dependencies to trigger correctly in the next steps)

- **install all the dependencies** This is just as before, with good caching (except if the opam files changed).
  For the happyeyeball issue, this step installs the `dns-client` from opam... and also the `happyeyeball` from opam (as required by `dns-client`)

- **install the repo** This is where we must `COPY .` everything and pin install all the `*.opam` files. This triggers an upgrade of `happyeyeball` and hence a reinstallation of `dns-client` (due to their weird double-repo setup, this wouldn't reinstall anything for normal users)

- **run the tests** Finally, we need the `--only-packages=X` option for all `X.opam`, as otherwise dune tries to use the local files and this can conflict with the version used by the opam dependencies.

So... this is definitively a quick hack to unblock the issue by avoiding the dependency analysis. It does cost two packages reinstallations for happyeyeball/dns-client and a lost opportunity in parrallelism for `dune build`.

The lack of good dependency analysis has been coming up a few times in opam-related CI, so there's definitively something to do there (but I think it's much harder to get it right with optional dependencies, the different operating system, etc) @maiste will have to rebase and fix this PR anyway, so we can still ask him to explore that path? :D

talex5:
I'm not following this issue closely, but one point: if the CI has to do all these tricks to build a project, then it's going to be really hard for users to build it from Git too!

Indeed, I just tried to clone this repository and build it:

```
$ git clone https://github.com/roburio/happy-eyeballs.git
$ cd happy-eyeballs
$ dune build
Error: Conflict between the following libraries:
- "happy-eyeballs" in _build/default/src
- "happy-eyeballs" in /home/user/.opam/4.12.0+domains/lib/happy-eyeballs
  -> required by library "dns-client.lwt" in
     /home/user/.opam/4.12.0+domains/lib/dns-client/lwt
  -> required by library "happy-eyeballs-lwt" in _build/default/lwt
-> required by executable test in app/dune:2
Error: Conflict between the following libraries:
- "happy-eyeballs" in _build/default/src
- "happy-eyeballs" in /home/user/.opam/4.12.0+domains/lib/happy-eyeballs
  -> required by library "dns-client.lwt" in
     /home/user/.opam/4.12.0+domains/lib/dns-client/lwt
Error: Conflict between the following libraries:
- "happy-eyeballs" in _build/default/src
- "happy-eyeballs" in /home/user/.opam/4.12.0+domains/lib/happy-eyeballs
  -> required by library "dns-client.mirage" in
     /home/user/.opam/4.12.0+domains/lib/dns-client/mirage
```

So perhaps we should regard the CI rejecting this as a feature?

maiste:
I agree with you as there are not so many repositories following this interleaving structure.

We only have this trick of changing the dependency computation by moving the pinning stage. Maybe, the solver modification (as @dra27 suggested) will be a minor change and be more explicit as it won't modify anything apart from the dependency computation.

However, if we consider the rejection a feature, are we not sending the message that the CI rejects specific project structures? :confused:

PR #468, opened 2022-05-25
Only build multicore packages on multicore compilers

talex5:
If a package depends on base-domains or base-effects then it requires a multicore compiler. Don't try to test it on other platforms.

Many projects want to add a sub-package with support for effects on OCaml 5.0, alongside their existing support for Lwt. This PR allows them to do that without losing CI on all older compilers. e.g.

- https://github.com/mirage/ocaml-cohttp/pull/857
- https://github.com/mirage/mirage-crypto/pull/155

### DIFF
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -44,6 +44,12 @@ let doc_spec ~base ~opam_files ~selection =
   let cache = [ Obuilder_spec.Cache.v Opam_build.download_cache ~target:"/home/opam/.opam/download-cache" ] in
   let network = ["host"] in
   let open Obuilder_spec in
+  let to_name x = OpamPackage.of_string x |> OpamPackage.name_to_string in
+  let only_packages =
+    match selection.Selection.only_packages with
+    | [] -> ""
+    | pkgs -> " --only-packages=" ^ String.concat "," (List.map to_name pkgs)
+  in
   stage ~from:base @@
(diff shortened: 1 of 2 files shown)


talex5:
Seems to work, but logs:
```
2022-05-25 14:46.05 [INFO] is_multicore "4.14.0+options" failed: Failure("unknown variant: options")
```

ocaml-version does `with_variant t (Some "options")`, but `of_string` then fails to parse this.

talex5:
Also, lint-doc needs to handle this.

dra27:
Flying by, somewhat, but two questions (!!): looks like a bug in `Ocaml_version.Configure_options.is_multicore` - surely that should be mapping any exception raised in parsing to `false`? More seriously, though, aren't we heading towards a state where this can/should be predicated on `"ocaml" {> 5.0.0~~}`?

talex5:
> aren't we heading towards a state where this can/should be predicated on `"ocaml" {> 5.0.0~~}`?

Yes. This PR is only needed until we have a 5.0 compiler available in CI. Not sure when that will happen.

dra27:
There should be an alpha0 within the next couple of weeks - the idea is that this fixes magic numbers, installation layout, and so forth. There are a few other things converging at which point alpha1 is released and we have a "normal" compiler release cycle... but CI (and tooling, etc.) should start converging on released versions as soon as alpha0 is cut.

tmcgilchrist (approved):

Issue #163, opened 2020-04-28
Add a rebuild button for a whole commit


kit-ty-kate:
It would be very useful to add a rebuild button on each `github/<user>/<repo>/commit/<hash>/` pages, for the whole pipeline used on this commit to be rebuilt.

talex5:
Why? It seems unlikely that all variants will suffer the same build worker failure.

Though a more general bulk-rebuild feature would be nice (e.g. *rebuild everything with "Disk full" in the log*).

talex5:
We now have a bulk-rebuild feature on the `/query` page of the admin UI. Maybe that's enough for what you need? Please reopen if not.

kit-ty-kate:
Here is an example where this is not enough: https://github.com/ocaml/opam-repository/pull/19397#issuecomment-913163748 Could this be reopened? I don't have the rights to do it myself.

kit-ty-kate:
> We now have a bulk-rebuild feature on the `/query` page of the admin UI. Maybe that's enough for what you need?

There are no filters per commits so I cannot do anything there.

kit-ty-kate:
Having to make an empty commit for every new PRs in opam-repository when something breaks down in the cluster is really annoying and it would be nice to have an accessible option to at least rebuild every failing builds

mseri:
Yes, this would be really useful. And also an equivalent version that only rebuilds the failed jobs

novemberkilo:
Documenting the plan:

1. A commit can be rebuilt if there are jobs in state Aborted | Failed | Undefined and there are no jobs that are in state Active | NotStarted (this means that the Cancel button, that appears only when there are Active | Not Started jobs, cannot appear at the same time as a Rebuild button)
2. When a commit can be rebuilt, show two buttons - "Rebuild Failed" and "Rebuild All" so that the user has the choice of rebuilding only the failed jobs, or to rebuild the entire commit.

novemberkilo:
Can this issue be closed now? // @tmcgilchrist @kit-ty-kate

maiste:
Closing as it has been done in https://github.com/ocurrent/ocaml-ci/pull/469.

PR #469, opened 2022-06-07
Adds support for rebuilding a commit. 

novemberkilo:
Fix for https://github.com/ocurrent/ocaml-ci/issues/163

### DIFF
--- a/web-ui/github.ml
+++ b/web-ui/github.ml
@@ -224,11 +224,26 @@ let repo_handle ~meth ~owner ~name ~repo path =
         | Active | NotStarted -> true
         | Aborted | Failed _ | Passed | Undefined _ -> false) false jobs
     in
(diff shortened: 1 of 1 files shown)


web-ui/github.ml
```diff
@@ -283,15 +298,18 @@ let repo_handle ~meth ~owner ~name ~repo path =
         respond_error `Internal_server_error reason
     end
   | `POST, ["commit"; hash; "cancel"] ->
+    let job_lookup commit job_info =
+      let variant = job_info.Client.variant in
+      Capability.with_ref (Client.Commit.job_of_variant commit variant) @@ fun job ->
```
talex5, line 303: Using `with_ref` here ensures that `job` is released before `job_lookup` returns, but this isn't what you want in the case where you return the job,

tmcgilchrist (approved):

Issue #383, opened 2021-10-28
lint-fmt build fails because it uses too recent a compiler


NathanReb:
This issue might be considered a duplicate of #261 but I opened a separate issue as that one has been closed and I'm not sure how related they actually are.

I recently ran into this kind of error:
https://ci.ocamllabs.io/github/NathanReb/ppx_yojson/commit/dc90d842097cccacfa8b751abeee7356e92b3f99/variant/(lint-fmt)

If I understand correctly, the lint-fmt build fails because it's running on a 4.13 opam switch and my project uses an ocamlformat version that's incompatible with 4.13.

I'll happily upgrade ocamlformat in that case but this seems like unwanted behaviour from ocaml-ci nonetheless. I imagine it to be legitimate to stick to an old ocamlformat version for as long as the formatting is unstable between versions so, while it isn't for me, it might be a problem for some other users. It is also worth considering that it will potentially happen everytime a new compiler is released, although I think ocamlformat should become compatible with future compiler versions in the future as it will 
have its own parser (@gpetiot and @emillon should be able to confirm whether this is true).

Not sure what the fix should be but I guess having an empty switch or at least one without invariant would be nice so that one is free to use whatever ocamlformat version they want.

emillon:
>  I think ocamlformat should become compatible with future compiler versions in the future as it will
have its own parser

Just replying to this part: yes, I think that's correct. Nowadays I don't think that ocamlformat has any compat layers left so it should be buildable with the latest version of the compiler (of course, the features it's able to format are tied to the version you use)

emillon:
(This is the case from 0.19.0 onwards. There's a `< 4.14` constraint in the opam file but I'm not sure it's relevant)

maiste:
Closing at it was fixed by #472. Feel free to reopen if needed.

PR #472, opened 2022-06-13
Fix lint-fmt issues

maiste:
Hi,
This PR is the continuation of the work done on #403 to solve #398:
* First, it rebases the work over master to make it up to date. 
* Thanks to the important help from @TheLortex, it adds another fix that uses @NathanReb solution to solve #383. It uses the solution computed by the solver to work on the latest compiler version that works with `ocamlformat`.

The idea is to get the selection computed to choose the image we want to build our image on.

PS: Another solution might be added in the future to support the `with-tool` field suggested by @dra27. But this is not something this PR intends to fix.

### DIFF
--- a/lib/analyse.ml
+++ b/lib/analyse.ml
@@ -182,6 +182,28 @@ module Analysis = struct
       and+ ys = lwt_result_list_mapm ~f xs in
       y::ys
 
+  let opam_dep_file packages =
+    let lines =
+      [ {|opam-version: "2.0"|}; {|depends: [|} ]
+      @ List.map (fun (pkg, ver) -> Printf.sprintf {|  "%s" %s|} pkg ver) packages
+      @ [ {|]|} ]
+    in
+    lines |> List.map (fun s -> s ^ "\n") |> String.concat ""
+
+  let exactly v = Printf.sprintf {|{ = "%s" }|} v
+
+  let find_opam_repo_commit_for_ocamlformat ~solve ~platforms version =
+    let ( let+ ) = Lwt_result.Infix.( >|= ) in
+    let deps_for_ocamlformat =
+      ( "deps_for_ocamlformat.opam"
+      , opam_dep_file [("ocamlformat", exactly version)])
+    in
+    let+ workers =
+      solve ~root_pkgs:[deps_for_ocamlformat] ~pinned_pkgs:[] ~platforms
+    in
+    let selection = List.hd workers in
+    selection.Selection.commit
+
   let of_dir ~solver ~job ~platforms ~opam_repository_commit dir =
     let solve = solve ~opam_repository_commit ~job ~solver in
     let ty = type_of_dir dir in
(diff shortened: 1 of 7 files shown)


lib/analyse_ocamlformat.ml
```diff
@@ -66,12 +66,23 @@ let ocamlformat_version_from_file ~root job path =
...
-    |> Lwt_result.map (Option.map (fun version -> Opam { version }))
+    let* version_in_dot_ocamlformat =
+      ocamlformat_version_from_file ~root job
+        Fpath.(to_string (root / ".ocamlformat"))
+    in
+    match version_in_dot_ocamlformat with
+    | None ->
+            Lwt_result.return (None, None)
+            (* This case happens when the ocamlformat file doesn't specify a version.
+               It will try to get the latest version for the last platform. *)
```
tmcgilchrist, line 85: Being opinionated here, this should probably fail or suggest adding the ocamlformat version to the file as a warning before trying the latest ocamlformat version. https://github.com/ocaml-ppx/ocamlformat#formatting-code

If someone hasn't specified a version, are they expecting to follow the latest or have they neglected to add it and need to be gently reminded.

talex5 (commented):
It looks like this is going to double the load on the solver service, which seems unreasonable to me. Do we really have to do this for every job?

tmcgilchrist:
To move some of the load off the solver service, we could either change the solver to use the cluster in the form of the solver service or have a plugin to opam that caches the ocaml versions that ocamlformat supports. The solver service would be a larger amount of work than the plugin. What do you think?

talex5:
> To move some of the load off the solver service, we could either change the solver to use the cluster in the form of the solver service or have a plugin to opam that caches the ocaml versions that ocamlformat supports. The solver service would be a larger amount of work than the plugin. What do you think?

Can't we just use a hash table in analyse.ml?

maiste:
> Can't we just use a hash table in analyse.ml?
Wouldn't it be interesting to use OCurrent to handle caching instead of a hash table, or is that too much?

talex5:
> Wouldn't it be interesting to use OCurrent to handle caching instead of a hash table, or is that too much?

That would be fine too!

maiste:
I added the cache, as suggested with a hash table  (shorter than the OCurrent_cache in the end), so the solver can be relieved.

lib/analyse.ml
```diff
@@ -182,6 +185,35 @@ module Analysis = struct
...
+  let solver_cache = Hashtbl.create 128
+
+  let find_opam_repo_commit_for_ocamlformat ~solve ~platforms version =
+    let ( let+ ) = Lwt_result.Infix.( >|= ) in
+    let deps_for_ocamlformat =
+      ( "deps_for_ocamlformat.opam"
+      , opam_dep_file [("ocamlformat", exactly version)])
+    in
+    let+ workers =
+        match Hashtbl.find_opt solver_cache  platforms with
```
talex5, line 207: Shouldn't this be keyed on `version`?

I don't think `platforms` needs to be in the key at all, although it could be stored with the result. Then if the result is for a different set of platforms you can reset the cache, because it's out of date.

lib/analyse.ml
```diff
@@ -228,7 +263,7 @@ module Analysis = struct
             (fun l -> `Opam_monorepo l)
         | `Ocaml_repo -> opam_selections ~solve ~job ~platforms ~opam_files dir
       end >>!= fun selections ->
-      let r = { opam_files; ocamlformat_source; selections } in
+          let r = { opam_files; ocamlformat_selection; ocamlformat_source; selections } in
```
talex5, line 266: Why is this indented?

maiste: Typo thanks!

lib/analyse.ml
maiste: My mistake! However, should we really clear the cache if `platforms` mismatch? As some packages disable some of them in their opam file, wouldn't it have an impact or am I confusing two different things?

talex5: `platforms` here is the input to the analysis, which should be the complete set. The solver may filter some of them out later.

tmcgilchrist (approved):

talex5 (approved):
Thanks!

PR #473, opened 2022-06-14
Turn on OCaml 5.0 testing

dra27:
Not to be merged until Dune 3.2.1/3.3.0 is released!


tmattio:
My understanding is that we wanted to use the CI to have a good landscape of what needs fixing, so I don't think we want to enable the alpha repository.

@kit-ty-kate @dra27 what do you think?

dra27:
This also got inadvertently deployed at the weekend because of a dependency bump 🙂

kit-ty-kate:
Thanks!

PR #476, opened 2022-06-17
Improve installation instructions

talex5:
Show how to install the dependencies, including pinning matrix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ Get the code with:
 
 ```sh
 git clone --recursive https://github.com/ocurrent/ocaml-ci.git
+cd ocaml-ci
+opam install --deps-only -t .
 ```
 
 Note: you need to clone with `--recursive` because this project uses submodules


maiste:
It might be a dummy question, but there is something I don't understand... Why pin `ocaml-matrix` if it is in the git submodule?

talex5:
So that we can use opam to install the dependencies.

maiste (approved):

MisterDA:
Maybe it could be even better *not* to depend on packages we use in submodules as Etienne suggested, and not pin them either so that we don't "pollute" the global opam switch if any.
Instead, to install the dependencies of the submodules we could do something like this for each package we depend on in each submodule.

```sh
opam install --deps-only --with-test ./ocaml-matrix/matrix_commom.opam
```

talex5:
The matrix dependency has gone for now, so I've removed the pins.

PR #477, opened 2022-06-17
Update dependencies

maiste:
This PR is a temporary fix to make the CI works with `Prometheus`. It also removes `matrix-current` from the dependencies as it is a git submodule now.


MisterDA:
You shouldn't remove the dependency on `matrix_current`, if we're pinning the submodule as suggested in `https://github.com/ocurrent/ocaml-ci/pull/476`.

PR #488, opened 2022-07-06
Fix odoc selection with OCaml 5.0

talex5:
ocaml-ci installs odoc without telling the solver about it. This means we might not get the latest version, and on OCaml 5 that means it might not install at all. As a hack, when choosing the opam-repository commit to use for builds, make sure it has odoc 2.1.1 at least.


talex5:
Pushed live for testing.

PR #491, opened 2022-07-13
Remove ocaml-matrix.

novemberkilo:
`ocaml-matrix` depends on a fork of dream. This is blocking progress on the redesign work - we would like to depend on mainline dream.

It would be good to update ocaml-matrix so that it also depends on mainline dream. Once that is done, this commit can be reverted to restore ocaml-matrix to this project.


tmcgilchrist (approved):
Removing for now, we should aim to update ocaml-matrix and re-introduce when it is on a compatible dream version.

PR #494, opened 2022-07-15
Allow running on arches unsupported by the master distro and disabling arches (RISC-V)

MisterDA:


MisterDA:
Maybe we should be using `# syntax=docker/dockerfile:1.4` or `# syntax=docker/dockerfile:1` instead of the sha.

lib/buildkit_syntax.ml
```diff
@@ -12,5 +12,6 @@ let add arch =
     | `X86_64 | `I386 -> `X86_64
     | `Aarch64 | `Aarch32 -> `Aarch64
     | `Ppc64le -> `Ppc64le
-    | `S390x -> `S390x) in
+    | `S390x -> `S390x
+    | _ -> failwith "No support for this architecture in docker/dockerfile:experimental.") in
```
kit-ty-kate, line 16: It’s almost always better to be explicit about match cases to avoid some bugs later on
```suggestion
    | `Riscv64 -> failwith "No support for this architecture in docker/dockerfile:experimental.") in
```

MisterDA: I hesitated, thanks.

tmcgilchrist (commented):
Too early to enable this for CI, there is not sufficient capacity in the riscv64 pool.

MisterDA:
Of course, let's keep it for later 🐊

MisterDA:
I've changed this PR to allow running CI jobs on architectures unsupported by the master distro (Ubuntu supports RISC-V but Debian didn't (at the time of writing)) and disabling a list of architectures (say, RISC-V).

shonfeder:
Given the conflict, how stale this PR is, and the fact that the motivating issue behind is not evident, I'm going to close this up for now. Please open an issue to explain the need for this functionality and the impact of not having it, if still relevant.

tmcgilchrist:
The motivation behind this change was to allow ocaml-ci to build on RISC-V hardware, with the understanding that ocaml-ci should support all Tier-1 OS/architecture combinations. At the time this was opened we only had 2 underpowered SiFive machines, now we have 6 machines in the [RISC-V pool](http://infra.ocaml.org/by-use/linux-riscv64) it seems more viable to provide ocaml-ci builds. That would depend on the capacity available in that pool, I haven't checked what the usage of it is.

The complicating factor is how ocaml-ci derives the opam vars for a build. It assumes they can be derived from the master distribution (Debian) but for RISC-V it isn't supported and it needs to be worked around. That is what https://github.com/ocurrent/ocaml-ci/pull/494#issuecomment-1387422625 is talking about.  In general deriving opam vars for builds could use some work since MacOS and FreeBSD end up being hardcoded in https://github.com/ocurrent/ocaml-ci/blob/f7d66cc929abd21bc390ae74c4579eafeafb503e/service/conf.ml when ideally they should be derived from the containers  and ocaml-version dependencies. Happy to provide more details :-)

shonfeder:
Thank you for the context! I've opened https://github.com/ocurrent/ocaml-ci/issues/955 to track. I'd have no problems at all with this PR being reopened if someone has the free cycles to wrap it up.

PR #497, opened 2022-07-19
monorepo: remove support for 0.1 lockfiles

emillon:
These are not used anymore.

### DIFF
--- a/lib/opam_monorepo.ml
+++ b/lib/opam_monorepo.ml
@@ -1,6 +1,6 @@
 type info = string * OpamFile.OPAM.t
 
-type lock_file_version = V0_1 | V0_2 | V0_3 [@@deriving yojson, ord]
+type lock_file_version = V0_2 | V0_3 [@@deriving yojson, ord]
 
 (** The kind of switch the package will be built in. *)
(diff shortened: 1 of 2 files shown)


lib/opam_monorepo.ml
```diff
@@ -81,17 +81,17 @@ let opam_monorepo_dep_version ~lock_file ~package =
   |> version_of_equal_constraint
 
 let lock_file_version_of_string = function
-  | "0.1" -> V0_1
   | "0.2" -> V0_2
   | "0.3" -> V0_3
+  | "0.1" as s ->
+    Printf.ksprintf failwith "lockfile version %s is not supported anymore" s
```
tmcgilchrist, line 87: Is there a tooling option for upgrading from 0.1 to 0.2/0.3 lock files? 
Could that a pointer to that tool or doc be included here?

emillon: No, there's not really such a mechanism at the moment. The recommended way to handle it is to regenerate the lock file from a newer version. I'll mention this.

tmcgilchrist (approved):

Issue #410, opened 2022-01-27
`lint-doc` stage does not account for constraints on Odoc


CraigFe:
Currently Irmin's documentation does not build correctly when using Odoc 2.0.{1,2} (https://github.com/ocaml/odoc/issues/793), and as a result we have added an incompatibility with these versions to `irmin.opam` ([here](https://github.com/mirage/irmin/blob/ddf0ec90e373117701d6d689209bc746d4db92ab/irmin.opam#L37)).

Unfortunately, `ocaml-ci` does not account for the project's constraints on Odoc when [installing it](https://github.com/ocurrent/ocaml-ci/blob/master/lib/lint.ml#L50). This is causing Irmin's CI to fail consistently (e.g. [here](https://ci.ocamllabs.io/github/mirage/irmin/commit/ddf0ec90e373117701d6d689209bc746d4db92ab/variant/(lint-doc))). Perhaps Odoc should be installed along with the project's dependencies to ensure that this doesn't happen.

tmcgilchrist:
Related https://github.com/ocurrent/ocaml-ci/issues/264

PR #500, opened 2022-07-21
Solve and install doc dependencies

### LINKED ISSUE #264: lint-doc should install doc dependencies
The `lint-doc` phase does not currently seem to install the doc dependencies (packages tagged with `{with-doc}`).
This might be an issue for packages that have custom steps (e.g. building pdfs from latex, in which case they might have `"conf-texlive" {with-doc}` or something similar).

To test the documentation, ocaml-ci currently installs odoc and dune by hand in the `lint-phase`
https://github.com/ocurrent/ocaml-ci/blob/7df11b81707a3c9207a3e84ad85c04f78f351ad3/lib/lint.ml#L42
However if ocaml-ci were to install the doc dependencies this step would be redundant.

Furthermore, if odoc were to not be listed in the dependencies it would still be an issue for users down the road but would not be indicated by ocaml-ci. In my opinion ocaml-ci should only check that odoc >= 1.5.0 is installed but not install it. This incidently will also make this phase faster by removing two calls to opam depext (which itself calls opam twice)

MisterDA:
From what I gather, it's too costly to do two solves, so now the main analysis is also taking into account docs, which means that the main build stages will also install docs dependencies. The only way I can think of to avoid that would be to remove from the dependency graph starting with the `with-doc` roots all reachable packages that are not themselves reachable from non-`with-doc` roots. Overkill?

> In my opinion ocaml-ci should only check that odoc >= 1.5.0 is installed but not install it.

_from [https://github.com/ocurrent/ocaml-ci/issues/264#issue-728552519](#264)_.

I've changed the script to check the presence of `odoc >= 1.5.0` however there's still the hack that installs the latest version of odoc.

Should fix #264 and #410.


kit-ty-kate:
> I've changed the script to check the presence of `odoc >= 1.5.0` however there's still the hack that installs the latest version of odoc.

why trying to check that? Dune will fail anyway if the documentations are built and odoc is missing so that check should be redundant

kit-ty-kate:
> From what I gather, it's too costly to do two solves, so now the main analysis is also taking into account docs, which means that the main build stages will also install docs dependencies.

Is that really more time than the time spent building odoc and its dependencies (N times for each jobs) ?

MisterDA:
> > From what I gather, it's too costly to do two solves, so now the main analysis is also taking into account docs, which means that the main build stages will also install docs dependencies.
> 
> Is that really more time than the time spent building odoc and its dependencies (N times for each jobs) ?

Likely not, I'm switching to solving for docs deps separately.

PR #501, opened 2022-07-22
Setting up a dream based web-ui.

novemberkilo:
This represents a complete cut-over of `web-ui` to a `dream` based implementation. It aims to provide the same user experience with the addition of:

- support for flash messages (and use thereof when we rebuild a step)
- csrf protection (secure forms)

It is expected that we will delete the current `web-ui` directory, and rename the `dream-web` directory to `web-ui` so that it can take its place. We will then bring in the redesign of the site using the conventions being established here.


dream-web/controller/github.ml
```diff
@@ -0,0 +1,157 @@
...
+let list_orgs ci =
+  Client.CI.orgs ci >>!= fun orgs ->
+  Dream.respond @@ View.Github.list_orgs ~orgs
+
+let list_repos ~org ci =
+  let list_repos' org o =
+    Client.Org.repos o >>!= fun repos ->
+    Dream.respond @@ View.Github.list_repos ~org ~repos
+  in
+  Capability.with_ref (Client.CI.org ci org) @@ list_repos' org
```
tmcgilchrist, line 29:
```suggestion
  Capability.with_ref (Client.CI.org ci org) @@ fun o ->
  Client.Org.repos o >>!= fun repos ->
  Dream.respond @@ View.Github.list_repos ~org ~repos
```

I would make this more direct like so

dream-web/view/dune
```diff
@@ -0,0 +1,6 @@
+(library
+ (public_name dream-web.view)
+ (name view)
+ (libraries base ansi stdio unix fmt dream tyxml ocaml_ci_api)
```
tmcgilchrist, line 4:
```suggestion
 (libraries ansi unix fmt dream tyxml ocaml_ci_api)
```

dream-web/view/github.ml
```diff
@@ -0,0 +1,399 @@
...
+    in
+    Astring.String.cut ~sep:"@@@" body |> Option.get
+  in
+  let ansi = Ansi.create () in
+  let open Lwt.Infix in
+  Dream.stream
+    ~headers:[ ("Content-type", "text/html; charset=utf-8") ]
+    (fun response_stream ->
+      let%lwt () = Dream.write response_stream header in
+      let%lwt () = Dream.write response_stream (Ansi.process ansi data) in
```
tmcgilchrist, line 383:
```suggestion
      Dream.write response_stream header >>= fun () ->
      Dream.write response_stream (Ansi.process ansi data) >>= fun () ->
```
Maybe the version without ppx is better?

MisterDA: I find the version with the ppx better, it correctly propagates backtraces which is super for debugging. It's also possible to write `;%lwt` for sequences but the readability of _that_ is discutable.

dream-web/view/github.ml
```diff
@@ -0,0 +1,399 @@
...
+    | _ :: _, (Leaf _ as t) :: ts -> t :: add k x ts
+    | k :: ks, Branch (k', t) :: ts when Astring.String.equal k k' ->
+        Branch (k, add ks x t) :: ts
+    | _ :: _, (Branch _ as t) :: ts -> t :: add k x ts
+end
+
+let short_hash = Astring.String.with_range ~len:6
+
+let breadcrumbs steps page_title =
+  let open Tyxml.Html in
```
maiste, line 42: Instead of opening `Tyxml.Html` every time, wouldn't it be better to open it once? Otherwise shouldn't the function by wrapped into a module and `Tyxml.Html` open at the top of this module?

dream-web/view/github.ml
```diff
@@ -0,0 +1,399 @@
...
+  let can_cancel =
+    List.fold_left
+      (fun accum job_info ->
+        accum
+        ||
+        match job_info.Client.outcome with
+        | Active | NotStarted -> true
+        | Aborted | Failed _ | Passed | Undefined _ -> false)
+      false jobs
```
maiste, line 282: Maybe this can be written with a simplified version with:
```suggestion
     let check { job_info.Client.outcome; _ } with
        match outcome with
        | Active | NotStarted -> true
        | Aborted | Failed _ | Passed | Undefined _ -> false)
     in
    List.exists check jobs
```

dream-web/view/github.ml
```diff
@@ -0,0 +1,399 @@
...
+  in
+  let can_rebuild =
+    List.fold_left
+      (fun accum job_info ->
+        accum
+        ||
+        match job_info.Client.outcome with
+        | Active | NotStarted | Passed -> false
+        | Aborted | Failed _ | Undefined _ -> true)
+      false jobs
```
maiste, line 292: Same as above.

dream-web/view/template.ml
```diff
@@ -0,0 +1,34 @@
...
+    (html
+       (head
+          (title (txt "OCaml-CI"))
+          [
+            meta ~a:[ a_charset "UTF-8" ] ();
+            link ~rel:[ `Stylesheet ] ~href:"/css/normalize.css" ();
+            link ~rel:[ `Stylesheet ] ~href:"/css/ansi.css" ();
+            link ~rel:[ `Stylesheet ] ~href:"/css/github.css" ();
+            link ~rel:[ `Stylesheet ] ~href:"/css/style.css" ();
+            link ~rel:[ `Stylesheet ] ~href:"/css/flash-messages.css" ();
```
maiste, line 27: Can't all these files be concatenated in one file? It would avoid calling the server five times and save some bandwidth.

dream-web/main.ml
```diff
@@ -0,0 +1,106 @@
+let main port backend_cap =
+  Lwt_main.run
+    (let open Lwt.Infix in
+    Backend.Make.ci backend_cap >>= fun ci ->
```
maiste, line 4: As you use the `keyword%lwt` syntax in all the file should this be written:
```suggestion
    let%lwt ci = Backend.Make.ci backend_cap in
```

maiste (commented):
Thanks for this big piece of work, it's nice :+1: 
A general comment, IMHO, would be to choose a style for `Lwt` and try to keep the same for readability. And, it's an opinionated comment so maybe others would disagree with it :smile:

dream-web/.ocamlformat
```diff
@@ -0,0 +1,6 @@
+version = 0.23.0
+ocaml-version = 4.14.0
+profile = conventional
+break-infix = fit-or-vertical
+parse-docstrings = true
+module-item-spacing = compact
```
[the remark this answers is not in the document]
novemberkilo: Thanks @MisterDA for covering this in #503 :) ❤️

dream-web/view/github.ml
novemberkilo: Thanks all for your input - I'm switching everything to the version without ppx for consistency.

novemberkilo:
Thank you very much everyone for your review. I have folded in several of your suggestions and things are much improved.

PR #503, opened 2022-07-22
Config OCamlFormat but don't enable it ;)

MisterDA:
> Can this be added a standalone PR? It should be at the top level to cover all files.

https://github.com/ocurrent/ocaml-ci/pull/501#discussion_r927233433

There it is to help @novemberkilo a bit. As I'm not sure whether we want to enable it globally, I've disabled ocamlformat for now. It could be used just for the "standalone" dream-web module, if needed.


novemberkilo:
I think this is subsumed by #514

PR #505, opened 2022-07-28
Hack: convert ocaml-ci-web to eio

talex5:
This is just a hack to test Eio. It's using Eio versions of cohttp and capnp-rpc.

The Prometheus library is still lwt, and it uses lwt_eio for that.

Note: there are some problems in cohttp-eio; see https://github.com/mirage/ocaml-cohttp/issues/913 and https://github.com/mirage/ocaml-cohttp/issues/883. There's some extra code to work around that a bit.


avsm:
An eio version of capnp?! Hurrah!!

talex5:
> An eio version of capnp?! Hurrah!!

Yes; an experimental branch at https://github.com/talex5/capnp-rpc/tree/eio

Note that, confusingly, the package is still named capnp-rpc-lwt for now (to make the diff easier to read).

PR #506, opened 2022-08-01
Adds badge route to dream-web.

novemberkilo:


dream-web/controller/badges.ml
```diff
@@ -0,0 +1,73 @@
+type schema = {
+  schema_version : int; [@key "schemaVersion"]  (** always: 1 *)
```
maiste, line 2: Couldn't the `[@default 1]` be added to enforce the condition?

dream-web/controller/badges.ml
```diff
@@ -0,0 +1,73 @@
...
+
+let schema_of_status =
+  let v = v ~label:"ocaml-ci" in
+  function
+  | `Not_started -> v ~message:"unknown" ~color:"inactive" ()
+  | `Pending -> v ~message:"building" ~color:"yellow" ()
+  | `Failed -> v ~message:"failing" ~color:"critical" ()
+  | `Passed -> v ~message:"passing" ~color:"success" ()
+
+let ( let* ) = Lwt.Infix.( >>= )
```
maiste, line 47:
```suggestion
let ( let* ) = Lwt.Syntax.( let* )
```

maiste: And why use `(let*)` as it doesn't seem to be used anywhere? :wink:

maiste (commented):
Small typos but otherwise it LGTM :+1:

dream-web/controller/badges.ml
tmcgilchrist: Should check if the `@default value` can be used to enforce a stricter typing here.

PR #507, opened 2022-08-02
Cleanup - replace web-ui with dream-web

novemberkilo:
This PR:

- Replaces web-ui with dream-web
- Implements prometheus metrics 
- Uses ocaml-crunch and `Dream.static` to serve static files (removing the workaround of duplicating the `static` folder in the root of the project)


dream-web/dune
```diff
@@ -1,3 +1,9 @@
+(rule
+ (target static.ml)
+ (deps (source_tree static))
+ (action (with-stdout-to %{null}
+  (run ocaml-crunch --mode=plain --output=%{target} static))))
```
MisterDA: You don't need the `with-stdout-to`, and you can add the `--silent` flag to `ocaml-crunch`.

MisterDA (approved):

web-ui/backend.mli
```diff
@@ -1,7 +1,13 @@
...
 
 type t
 
 val make : Ocaml_ci_api.Raw.Client.CI.t Sturdy_ref.t -> t
 
 val ci : t -> Ocaml_ci_api.Client.CI.t Lwt.t
+
+module Make : sig
+  val ci : Uri.t -> Ocaml_ci_api.Client.CI.t Lwt.t
+end
```
maiste, line 13: Why do we need to expose `ci` and `Make.ci`? What is the difference between them apart from the Uri.t?

web-ui/controller/badges.ml
```diff
@@ -1,5 +1,5 @@
 type schema = {
-  schema_version : int; [@key "schemaVersion"]  (** always: 1 *)
+  schema_version : int; [@key "schemaVersion"] [@default 1]  (** always: 1 *)
```
maiste, line 2:
```suggestion
  schema_version : (int  [@default 1]); [@key "schemaVersion"]  (** always: 1 *)
```
In the [ppx_yojson README](https://github.com/ocaml-ppx/ppx_deriving_yojson#default) it is written this way. I don't know if it changes something, but I thought it worth pointing out...

web-ui/main.ml
```diff
@@ -1,101 +1,137 @@
...
+   See also https://github.com/aantron/dream/tree/master/example/w-one-binary *)
+  let loader root path _request =
+    match Static.read (Filename.concat root path) with
+    | None -> Dream.empty `Not_Found
+    | Some asset -> Dream.respond asset
 
-let static ~content_type ?(max_age=86400) body =
-  let headers = headers content_type max_age in
-  Server.respond_string ~status:`OK ~headers ~body ()
+  let t ci = Dream.router
```
maiste, line 11: Shouldn't this module live as a standalone file? IMHO, it feels like it will simplify the `main` module and keep the idea that it's the "late level" where we put pieces together.

web-ui/main.ml
```diff
@@ -1,101 +1,137 @@
...
-  | `GET, ("badge" :: path) ->
-     Badges.handle ~backend ~path
-  | _ ->
-    Server.respond_not_found () |> normal_response
-
-let pp_mode f mode =
-  Sexplib.Sexp.pp_hum f (Conduit_lwt_unix.sexp_of_server mode)
+let setup_logs default_level =
+  Prometheus_unix.Logging.init ?default_level ()
+  let () = Dream.initialize_log ()
```
maiste, line 97:
```suggestion
  Prometheus_unix.Logging.init ?default_level ();
  Dream.initialize_log ()
```
Maybe I missed a point here?

maiste (commented):
Few questions, otherwise, LGTM :+1: !

web-ui/backend.mli
novemberkilo: I guess the implementation enables this kind of thing and I've been playing with the notion that this is idiomatic OCaml 🤷🏾‍♂️ 

```
    Backend.Make.ci backend_cap >>= fun ci ->

```

web-ui/controller/badges.ml
novemberkilo: Thanks - I should have read that first :)

web-ui/main.ml
novemberkilo: I thought about this when I extracted the module and didn't actually do anything about it at the time. Thanks for the suggestion 👍🏾

PR #509, opened 2022-08-09
Persist refs in the build index.

novemberkilo:
In preparation for supporting history of builds for a branch, we need to add refs to the persistence layer. Then we can lookup the commits/builds associated to the branch of a repo - i.e. to an `(owner, name, ref)`

### DIFF
--- a/lib/index.ml
+++ b/lib/index.ml
@@ -1,8 +1,7 @@
 let src = Logs.Src.create "ocaml_ci.index" ~doc:"ocaml-ci indexer"
+
 module Log = (val Logs.src_log src : Logs.LOG)
-
 module Db = Current.Db
-
 module Job_map = Astring.String.Map
 
 type t = {
@@ -11,10 +10,13 @@ type t = {
   get_jobs : Sqlite3.stmt;
   get_job : Sqlite3.stmt;
   get_job_ids : Sqlite3.stmt;
+  get_commits_job_ids_for_ref : Sqlite3.stmt;
   full_hash : Sqlite3.stmt;
 }
 
-type job_state = [`Not_started | `Active | `Failed of string | `Passed | `Aborted ] [@@deriving show]
+type job_state =
+  [ `Not_started | `Active | `Failed of string | `Passed | `Aborted ]
+[@@deriving show]
 
 type build_status = [ `Not_started | `Pending | `Failed | `Passed ]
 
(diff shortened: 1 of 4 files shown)


lib/index.mli
```diff
@@ -26,6 +27,8 @@ val get_job : owner:string -> name:string -> hash:string -> variant:string -> (s
 
 val get_job_ids: owner:string -> name:string -> hash:string -> string list
 
+val get_build_history : owner:string -> name:string -> ref:string -> (string * string * string option) list
+
```
tmcgilchrist, line 31: Could you document this API.

lib/index.ml
```diff
@@ -85,104 +138,133 @@ module Status_cache = struct
...
+            variant
+            Fmt.(option ~none:(any "-") string)
+            job_id);
       match job_id with
-      | None -> Db.exec t.record_job Sqlite3.Data.[ TEXT owner; TEXT name; TEXT hash; TEXT variant; NULL ]
-      | Some id -> Db.exec t.record_job Sqlite3.Data.[ TEXT owner; TEXT name; TEXT hash; TEXT variant; TEXT id ]
+      | None ->
+          Db.exec t.record_job
+            Sqlite3.Data.
+              [ TEXT owner; TEXT name; TEXT hash; TEXT variant; TEXT ref; NULL ]
```
tmcgilchrist, line 168: Rename this from `ref` it clashes with the build in ref keyword.

lib/index.ml
```diff
@@ -27,52 +29,103 @@ let is_valid_hash hash =
...
+ALTER TABLE ci_build_index
+  ADD COLUMN ref TEXT NOT NULL DEFAULT "-"
+    |}
+       |> or_fail "alter table"
+     else Log.info (fun f -> f "ref col detected.");
+     let record_job =
+       Sqlite3.prepare db
+         "INSERT OR REPLACE INTO ci_build_index (owner, name, hash, variant, \
+          ref, job_id) VALUES (?, ?, ?, ?, ?, ?)"
+     in
```
tmcgilchrist, line 68: These could be rewritten as let and's since they are independent eg
```ocaml
let record_job = ...
and remove = ...
```

tmcgilchrist (commented):
There are some formatting changes that seem unnecessary in this PR, could you review them and just make the necessary minor changes.

PR #510, opened 2022-08-10
Tidyup

tmcgilchrist:


web-ui/router.mli
```diff
@@ -1,4 +1,3 @@
-module Router : sig
-  val t : Ocaml_ci_api.Client.CI.t -> Dream.handler
-end
+val create : Ocaml_ci_api.Client.CI.t -> Dream.handler
+
```
novemberkilo, line 2: This is nice and clean and easier to read 👍🏾 

fwiw I found this and followed the `t` naming ... https://github.com/ocaml-templates/spin-dream/blob/main/example/src/main.ml

novemberkilo (approved):
Thanks for cleaning up @tmcgilchrist -- this is indeed much tidier :)

web-ui/router.mli
tmcgilchrist: I would always reserve `t` for the main type of a module.

PR #511, opened 2022-08-11
Remove un-necessary opens

tmcgilchrist:


novemberkilo (approved):
Thanks for the cleanup @tmcgilchrist

PR #512, opened 2022-08-12
Run times and durations for steps and builds.

novemberkilo:
Inspired by the module of the same name
in current_web_pipelines.

### DIFF
--- a/lib/run_time.mli
+++ b/lib/run_time.mli
@@ -1,0 +1,53 @@
+val src : Logs.src
+module Log : Logs.LOG
+
+type timestamps =
+    Queued of float
+  | Running of { ready : float; started : float; }
+  | Finished of { ready : float; started : float option; finished : float; }
+  (** Type to encode the timestamps of a step or a build. All floats represent
+      Unix epoch time - i.e. milliseconds from 1970-01-01 00:00:00 GMT
+
+      The step or build is either:
+      - Queued - the associated timestamp indicates when it was ready to run (enqueued)
+      - Running - with timestamps for when it was ready and when it started
+      - Finished - with timestamps for when it was ready, when it started and when it finished.
+      Note that it may never run -- it may be cancelled or it may timeout. *)
+
+type run_time_info =
+    Cached
+  | Queued_for of float
+  | Running of { queued_for : float; ran_for : float; }
+  | Finished of { queued_for : float; ran_for : float option; }
+  (** Type to encode the durations of the various stages of a step or build.
+      All floats represent milliseconds.
+
+      The step or build is either:
+      - Cached - a previously run instance of the step was used and nothing was run
+      - Queued - the associated duration represents the number of milliseconds it has been queued for (relative to now)
+      - Running - has a queued_for duration, and a ran_for duration. The latter should be interpreted as the number of
+      milliseconds that it has been running for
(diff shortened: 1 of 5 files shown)


lib/run_time.mli
```diff
@@ -0,0 +1,52 @@
...
+
+      The step or build is either:
+      - Queued - the associated duration represents the number of milliseconds it has been queued for (relative to now)
+      - Running - has a queued_for duration, and a ran_for duration. The latter should be interpreted as the number of
+      milliseconds that it has been running for
+      - Finished - has a queued_for and an optional ran_for duration. The ran_for duration will be None when it did
+      not run. *)
+
+val pp_timestamps : Format.formatter -> timestamps -> unit
+(** ppx_derived Show formatter for timestamps *)
```
tmcgilchrist, line 33: Should this just be a Fmt function that takes a `timestamp` and leave the IO to the caller. It would be more flexible that way.

test/test_run_time.ml
```diff
@@ -0,0 +1,309 @@
...
+  let state f (st: Run_time.timestamps) = Fmt.pf f "%a" Run_time.pp_timestamps st in
+  let equal (st1: Run_time.timestamps) (st2: Run_time.timestamps) =
+    match (st1, st2) with
+    | Queued v1, Queued v2 -> cmp_floats v1 v2
+    | Running v1, Running v2 -> cmp_floats v1.ready v2.ready && cmp_floats v1.started v2.started
+    | Finished {ready=ready1; started=None; finished=finished1}, Finished {ready=ready2; started=None; finished=finished2} ->
+        cmp_floats ready1 ready2 && cmp_floats finished1 finished2
+    | Finished {ready=ready1; started=Some started1; finished=finished1}, Finished {ready=ready2; started=Some started2; finished=finished2} ->
+        cmp_floats ready1 ready2 && cmp_floats started1 started2 && cmp_floats finished1 finished2
+    | _ -> false
```
tmcgilchrist, line 25: Curious whether this can explicitly match without being too verbose.

test/test_run_time.ml
```diff
@@ -0,0 +1,309 @@
...
+    match (rt1, rt2) with
+    | Cached, Cached -> true
+    | Queued_for v1, Queued_for v2 -> cmp_floats v1 v2
+    | Running v1, Running v2 -> cmp_floats v1.queued_for v2.queued_for && cmp_floats v1.ran_for v2.ran_for
+    | Finished v1, Finished v2 ->
+      (match v1.ran_for, v2.ran_for with
+      | None, None -> cmp_floats v1.queued_for v2.queued_for
+      | Some r1, Some r2 -> cmp_floats v1.queued_for v2.queued_for && cmp_floats r1 r2
+      | _, _ -> false)
+    | _ -> false
```
tmcgilchrist, line 41: Same for these two.

lib/run_time.mli
novemberkilo: So this is what results from `[@@deriving show]` and it's easy to use it in tests. I don't expect this to be used further (and would ideally like it to not be part of the interface but I can't use it in tests otherwise).

> Should this just be a Fmt function that takes a timestamp 

Yep - we'd write presenters of this ilk to suit the cli and also the web front-end.

tmcgilchrist (approved):

PR #522, opened 2022-09-01
Timestamps and run-times for steps and builds.

novemberkilo:
This PR fixes a few issues that were detected when working on timestamps and durations for builds. It subsumes #517 and #516 

Notes:
- The pipelines in ocaml-ci start with an analysis step - thus a build is `created_at` the same time as the analysis step is created.
- A build corresponds to what `ci_build_index` knows about the jobs associated to a triple `(owner, name, hash)` -- this means that a build only contains the latest runs of any given variant


maiste (commented):
When running it on my own `ocaml-ci`, the time is displayed as `cached`. Is it something that we expect?

service/api_impl.ml
```diff
@@ -17,9 +18,33 @@ let make_commit ~engine ~owner ~name hash =
...
+                let started_at_t = Raw.Builder.JobInfo.started_at_init slot in
+                let module S = Raw.Builder.JobInfo.StartedAt in
+                (match started_at with
+                | None -> S.none_set started_at_t
+                | Some v -> S.ts_set started_at_t v);
+                let finished_at_t = Raw.Builder.JobInfo.finished_at_init slot in
+                let module S = Raw.Builder.JobInfo.FinishedAt in
+                (match finished_at with
+                | None -> S.none_set finished_at_t
+                | Some v -> S.ts_set finished_at_t v);
```
maiste, line 47: Maybe it is worth being split in a `let f variant outcome t = (* ... *)` function above as it is a big function?

test/dune
```diff
@@ -5,6 +5,7 @@
   alcotest
   alcotest-lwt
   ppx_deriving_yojson.runtime
+  client-utilities
```
maiste, line 8:
```suggestion
  ocaml-ci-client-utilities
```
Same: can be ignored.

web-ui/controller/dune
```diff
@@ -10,6 +10,7 @@
   ppx_deriving_yojson.runtime
   prometheus-app
   capnp-rpc-unix
-  ocaml-ci-web.view)
+  ocaml-ci-web.view
+  client-utilities)
```
maiste, line 14:
```suggestion
  ocaml-ci-client-utilities)
```
Same: can be ignored.

web-ui/controller/git_forge.ml
```diff
@@ -97,9 +98,16 @@ module Make (View : View) = struct
...
     let csrf_token = Dream.csrf_tag request in
     let flash_messages = Dream.flash_messages request in
-    Dream.respond
-    @@ View.list_steps ~org ~repo ~refs ~hash ~jobs ~csrf_token ~flash_messages
-         ()
+    let first_step_queued_at = Run_time.first_step_queued_at jobs in
+    match first_step_queued_at with
+    | Error e ->
+        Dream.log "Error - %s" e;
+        Dream.empty `Internal_Server_Error
```
maiste, line 105: I got this error when rebuilding a step. I was wondering how much it could be an issue.

web-ui/dune
```diff
@@ -16,4 +16,5 @@
   prometheus-app.unix
   ocaml-ci-api
   ocaml-ci-web.controller
-  ocaml-ci-web.view))
+  ocaml-ci-web.view
+  client-utilities))
```
maiste, line 20:
```suggestion
  ocaml-ci-client-utilities))
```
Same: can be ignored.

web-ui/view/dune
```diff
@@ -1,4 +1,4 @@
 (library
  (public_name ocaml-ci-web.view)
  (name view)
- (libraries ansi unix fmt dream tyxml ocaml_ci_api))
+ (libraries ansi unix fmt dream tyxml ocaml-ci-api client-utilities timedesc))
```
maiste, line 4:
```suggestion
 (libraries ansi unix fmt dream tyxml ocaml-ci-api ocaml-ci-client-utilities timedesc))
```
Can be ignored.

web-ui/view/git_forge.ml
```diff
@@ -55,6 +58,8 @@ module type View = sig
     job:Current_rpc.Job.t ->
     status:Current_rpc.Job.status ->
     csrf_token:string ->
+    timestamps:Run_time.timestamps option ->
+    build_created_at:float option ->
```
maiste, line 62: What is the difference between the creation time and the `queued_at`?

lib_client/run_time.ml
```diff
@@ -0,0 +1,209 @@
...
+        {
+          queued_for = started_at -. queued_at;
+          ran_for = current_time -. started_at;
+        }
+  | Finished { queued_at; started_at; finished_at } -> (
+      if finished_at < build_created_at then Cached
+      else
+        match started_at with
+        | None ->
+            Finished { queued_for = finished_at -. queued_at; ran_for = None }
```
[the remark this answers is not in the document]
novemberkilo: "Finish" is an overloaded term here -- I think that it indicates "finished successfully" to us but in fact, is modelled simply as the point at which a job stops. 

It can never start -- for example it could be cancelled or aborted. It will still have a finished_at timestamp which would be the point at which it was cancelled or aborted.

So, no, it is not an error if a job is finished but never started. It just indicates that the job never ran. (Confusing I know :/ )

lib_client/run_time.ml
```diff
@@ -0,0 +1,209 @@
...
+
+let timestamps_from_job_info (ji : Client.job_info) :
+    (timestamps, string) result =
+  match (ji.outcome : Client.State.t) with
+  | NotStarted ->
+      Option.fold ji.queued_at
+        ~none:(Error "Queued_at is None - cannot construct timestamp.")
+        ~some:(fun v -> Ok (Queued v))
+  | Active ->
+      Option.fold ji.queued_at
```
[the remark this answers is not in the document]
novemberkilo: fwiw I am not happy about this either but it is because `Client.job_info` is populated via a capnp API call. In this implementation I could not get `queued_at` to be just a `float`: 

See https://github.com/ocurrent/ocaml-ci/pull/522/files#diff-cea86c9ace189571331fc2cb600c0636162c5a4e7c32bc8d7377d67333a630ecR230 and thus https://github.com/ocurrent/ocaml-ci/pull/522/files#diff-c45befd13f73a35b60ccfd2a34324712f492068238eafb658d0219d38cc3795bR24

web-ui/controller/git_forge.ml
novemberkilo: That's interesting - I intentionally had it error out because I want to know the circumstances in which this can happen. 

I guess if everything is in the state `Not Started` it could. If it happens routinely then of course we should not crash. I'd like to "test this in prod" for a bit to better understand this. Can you recall the state of your system when you noticed the error?

web-ui/view/git_forge.ml
novemberkilo: https://github.com/ocurrent/ocaml-ci/pull/522/files#diff-95ab9cbeb2c74e16d0ab54903bca0b0a20d108d33a2336be6925fc4871dc22a1R75-R77

lib_client/run_time.ml
```diff
@@ -0,0 +1,209 @@
...
+      finished_at : float;
+    }
+[@@deriving show]
+
+type run_time_info =
+  | Cached
+    (* This indicates that the job never ran because cached results were used *)
+  | Queued_for of float (* elapsed time in seconds *)
+  | Running of { queued_for : float; ran_for : float }
+  | Finished of { queued_for : float; ran_for : float option }
```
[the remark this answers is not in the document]
novemberkilo: Indeed - and it is derived from a `timestamps` type.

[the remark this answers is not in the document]
novemberkilo: https://github.com/ocurrent/ocaml-ci/pull/522/files#diff-0766897181ef8bc55637f7dae75b471ffad37c1f6261b57a3f8c3de62eadcf14R121

lib_client/run_time.ml
```diff
@@ -0,0 +1,209 @@
...
+  | Cached -> 0.
+  | Queued_for v -> v
+  | Running v -> v.queued_for
+  | Finished v -> v.queued_for
+
+let ran_for = function
+  | Cached -> 0.
+  | Queued_for _ -> 0.
+  | Running v -> v.ran_for
+  | Finished v -> Option.value ~default:0. v.ran_for
```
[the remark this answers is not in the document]
novemberkilo: When the step does not run at all.

web-ui/controller/git_forge.ml
novemberkilo: On further reflection I am going to change my mind and not throw an error here. I don't think it's appropriate to crash if we can't get a timestamp.

lib/run_time.ml
```diff
@@ -1,73 +1,62 @@
...
-
-type run_time_info =
-  | Cached
-    (* This indicates that the job never ran because cached results were used *)
-  | Queued_for of float (* elapsed time in milliseconds *)
-  | Running of { queued_for : float; ran_for : float }
-  | Finished of { queued_for : float; ran_for : float option }
+  | Queued of float
+  | Running of { queued_at : float; started_at : float }
+  | Finished of {
```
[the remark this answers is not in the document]
novemberkilo: In summary, we could do what you are suggesting. 

In fact, we construct an instance of `timestamps` from `job_info` in `lib_client/run_time.ml` -- in my opinion this type is "tighter" than the one we would have if we just went with the suggestion and it gives us the opportunity to unpack the attributes of `job_info` and do some error checking in one spot (when we construct it). 

If we only used the `job_info` type I think we would end up doing a lot of the same error checking all over the code (concrete example: If a job is running and its `queued_at` and `started_at` timestamps are `None` then that's an error).

Hope that all makes sense? Happy to revisit this in more detail.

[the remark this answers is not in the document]
maiste: I got your point! Thanks for the explanation.

lib_client/run_time.ml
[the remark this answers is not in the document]
maiste: That's the explanation above, thanks ;)

lib_client/run_time.ml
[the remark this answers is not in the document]
maiste: That's fair.

PR #525, opened 2022-09-07
WIP: Build page with new design

novemberkilo:
This PR introduces

- the css from the new designs
- a new template so that we can have both the old and new styles coexist
- a helper module `Tyxml_helpers` for additional, non-standard attributes that are needed 
- a `Build` module that contains functions for building the components that appear on the build page
- a `Common` module that collects re-usable things like breadcrumbs, flash-messages, status-icons and other symbols

I've managed to cut across the entire build page, including the functionality of the buttons, so that we should have a feature complete build page in the new design. After discussing things with @tmcgilchrist we won't merge this until we also have the step page cut across to the new design. That work will build on this.

Screen Shot 2022-09-09 at 4 29 13 pm


web-ui/static/css/style_v1.css
```diff
@@ -0,0 +1,319 @@
+@tailwind base;
```
maiste, line 1: Is it extracted from the code Asad gave us?

web-ui/controller/git_forge.ml
```diff
@@ -192,24 +212,30 @@ module Make (View : View) = struct
...
+    (* let return_link = View.return_link ~org ~repo ~hash in
+       let csrf_token = Dream.csrf_tag request in
+       let first_step_queued_at =
+         match Run_time.first_step_queued_at jobs with
+         | Error e ->
+             Dream.log "Error - %s" e;
+             None
+         | Ok v -> Some v
+       in
+       let total_run_time = Run_time.total_of_run_times jobs in *)
```
maiste, line 233: Is there a reason this is commented and not deleted?

web-ui/controller/git_forge.ml
```diff
@@ -68,6 +68,19 @@ module Make (View : View) = struct
...
+    | Ok y -> f y
+    | v -> (
+        match v with
+        | Error (`Capnp ex) ->
+            Dream.log "Internal server error: %s"
+              (Fmt.to_to_string Capnp_rpc.Error.pp ex);
+            Dream.empty `Internal_Server_Error
+        | _ ->
+            (* FIXME: This is trying to match `Msg of variant *)
+            Dream.empty `Internal_Server_Error)
```
maiste, line 82: Why not change the `>>=` Operator to support the `Error (_)` case? Otherwise, can't this be written this way to simplify the syntax?
```suggestion
  let ( >>!!= ) x f =
    x >>= function
    | Ok y -> f y
    | Error (`Capnp ex) ->
            Dream.log "Internal server error: %s"
              (Fmt.to_to_string Capnp_rpc.Error.pp ex);
            Dream.empty `Internal_Server_Error
     | Error _ ->
            (* FIXME: This is trying to match `Msg of variant *)
            Dream.empty `Internal_Server_Error)
```

web-ui/controller/git_forge.ml
```diff
@@ -93,9 +106,16 @@ module Make (View : View) = struct
...
-    Client.Commit.refs commit >>!= fun refs ->
+    @@ fun commit_cap ->
+    Client.Commit.status commit_cap >>!!= fun status ->
+    Client.Commit.jobs commit_cap >>!= fun jobs ->
+    Client.Commit.refs commit_cap >>!= fun refs ->
+    let build_status : Client.State.t =
+      match status with
+      | `Failed -> Failed ""
+      | `Pending | `Not_started -> NotStarted
+      | `Passed -> Passed
```
maiste, line 117: Shouldn't this live in `Client.State` as we might want to change it in the future?

web-ui/view/git_forge.ml
```diff
@@ -15,6 +15,16 @@ module type View = sig
...
 
+  val cancel_success_message_v1 :
+    Client.job_info list -> ([> `Success ] * string) list
+
+  val cancel_fail_message_v1 : int -> ([> `Fail ] * string) list
+
+  val rebuild_success_message_v1 :
+    Client.job_info list -> ([> `Success ] * string) list
+
+  val rebuild_fail_message_v1 : int -> ([> `Fail ] * string) list
```
maiste, line 26: I'm not a big fan of the naming scheme.... I'll explain in the general review message.

web-ui/view/template_v1.ml
```diff
@@ -0,0 +1,69 @@
...
+        ();
+      script
+        ~a:
+          [
+            a_defer (); a_src "https://unpkg.com/alpinejs@3.x.x/dist/cdn.min.js";
+          ]
+        (txt "");
+      link ~rel:[ `Stylesheet ]
+        ~href:
+          "https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700&display=swap"
```
maiste, line 25: @tmattio I remember that you suggested downloading the font if we want to use them instead of using the CDN, but I can't remember why.  Could you give us a refresh about it?

web-ui/view/tyxml_helpers.ml
```diff
@@ -0,0 +1,8 @@
+let x_cloak = Tyxml.Html.Unsafe.string_attrib "x-cloak" ""
```
maiste, line 1: We should add the recurrent Tyxml structure we use to simplify our code flow. For example,
`a ~a:[href = ""] [txt ...]` or others that are used often. WDYT?

maiste (commented):
This is a nice PR that will bring a significant change to the CI :partying_face:  I agree with @tmcgilchrist and you that it's better to wait for the step to be ready before merging.  Some concerns I want to raise:
* The `_v1` name seems not to be the right idea as we will have to change it we the new versions, and it will break the API all the time. why not use an alias to address it? Like `let name = name_v1` hidden from the API side.
* The documentation and get started links don't point anywhere. Why not just hide them?
* Same for the history build: I can uncomment it when I will push the history page.

Thanks for this PR!

web-ui/view/template_v1.ml
tmattio: Essentially for user privacy: users might not want to have to send query to Google (which most likely run analytics on the requests)

We're vendoring everything in ocaml.org: https://github.com/ocaml/ocaml.org/tree/main/asset/vendors

MisterDA: Is there a real need for external fonts? Why not use [system-ui value for font-family](https://caniuse.com/font-family-system-ui)?

web-ui/static/css/style_v1.css
novemberkilo: Yes I have copied the generated css across.

web-ui/controller/git_forge.ml
novemberkilo: cruft - deleting

web-ui/view/tyxml_helpers.ml
novemberkilo: I wasn't sure if this might lead to a bit of confusion for readability (since the template will consist of standard HTML but also these simplifications) but on reflection I think this is a good idea - I'd like to introduce this in a consolidation step in the next PR perhaps.

web-ui/view/git_forge.ml
novemberkilo: I'm doing this to accomplish the following:

While we are cutting over to the new designs, we need to maintain the existing `Template` and also work with the new design that is introduced in `Template_v1`  My plan for the cutover is to make it so that when `Template` is no longer being used, we remove it altogether and then possibly rename `_v1` to no longer carry this suffix. 

This is the strategy for other helper functions (like `rebuild_fail_message`) as well. 

It is cumbersome in the interim, but it's following a system of naming that I expect will unambiguously allow us to have both the old and new while we need them to coexist.

Or perhaps I am missing the point of your suggestion?

maiste: Thanks for the clarification! The idea behind is pretty clear. I'm a bit concerned about the code cohabitation, but as I don't have a better solution to suggest, let's stick to your suggestion :+1:

novemberkilo:
Switching this back to draft since it is superseded by #527 and should be closed and not merged.

maiste:
Removing the assignment as this is shadowed by #527

novemberkilo:
Closing in favour of #527

PR #526, opened 2022-09-07
Support history within the web UI

maiste:
This PR introduces a building history into the web UI as a hidden link. Next PR will make it accessible from the step interface and conform it with the new UI.
- [X] Update the Capn Proto schema to support history
- [X] Create a view for GitHub
- [x] Create a view for GitLab
- [x] Replace `Either.t` with a custom type

### DIFF
--- a/api/client.ml
+++ b/api/client.ml
@@ -110,6 +110,30 @@ module Repo = struct
     let request, params = Capability.Request.create Params.init_pointer in
     Params.ref_set params gref;
     Capability.call_for_caps t method_id request Results.commit_get_pipelined
+
+  let history_of_ref t gref =
+    let open Raw.Client.Repo.HistoryOfRef in
+    let request, params = Capability.Request.create Params.init_pointer in
+    Params.ref_set params gref;
+    Capability.call_for_value t method_id request
+    |> Lwt_result.map @@ fun history ->
+       Results.refs_get_list history
+       |> List.fold_left
+            (fun acc slot ->
+              let open Build_status in
+              let hash = Raw.Reader.RefInfo.hash_get slot in
+              let state = Raw.Reader.RefInfo.state_get slot in
+              let started = Raw.Reader.RefInfo.started_get slot in
+              let time =
+                match Raw.Reader.RefInfo.Started.get started with
+                | Raw.Reader.RefInfo.Started.None | Undefined _ -> None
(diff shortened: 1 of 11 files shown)


novemberkilo (approved):
- [x] Run up branch locally
- [x] Review the code

Hey @maiste thanks for this - this is clean and nice 🎉  

A couple of questions and comments -

Do we know what order the commits are listed in? It should be ordered similarly to the git history of the branch. I think we should confirm this and tighten down the SQL query if we can to ensure that this happens. We might need to join with the cache to get to timestamps etc.?

Have you been able to confirm that the gitlab approach works by looking at the history of a gitlab project? If you don't have credentials to do this I think maybe we can ask @tmcgilchrist to please run this locally to confirm?

maiste:
> Hey @maiste thanks for this - this is clean and nice

Thanks a lot :smile: 

> Do we know what order the commits are listed in? It should be ordered similarly to the git history of the branch. I think we should confirm this and tighten down the SQL query if we can to ensure that this happens. We might need to join with the cache to get to timestamps etc.?

I was expecting the question! They are not ordered for know and I was wondering if I could use the First build Timestamp to get order the commits? Or should I check with `GitHub` and `Gitlab` to get the history?

> Have you been able to confirm that the gitlab approach works by looking at the history of a gitlab project? If you don't have credentials to do this I think maybe we can ask @tmcgilchrist to please run this locally to confirm?

No indeed, I wanted your point of view on it before. @tmcgilchrist could you try with `GitLab` to ensure the links are working?

novemberkilo:
> I was expecting the question! They are not ordered for know and I was wondering if I could use the First build Timestamp to get order the commits? Or should I check with GitHub and Gitlab to get the history?

I think we should use the timestamps in cache but this is because I think this will correctly reflect the order in which we built the commits. Since we won't build every commit, I'm wondering if we shouldn't push ahead with the timestamps-from-current-cache approach instead of getting the commit history from GitHub.

maiste:
I agree with you. The second approach has a too-high cost if we have to do it each time...

maiste:
This is a new version with a history order from the most recent to the old one. I'm not a fan of the method as it requires the cache to be up to date, and it relies on a bit of a hack by using the cache time. Maybe having some persistent data for this kind of data would be better.

novemberkilo:
Closing in favour of #527

PR #527, opened 2022-09-14
New build and step pages

novemberkilo:
Supersedes #525 by building on it.

Screen Shot 2022-09-09 at 4 29 13 pm
Screen Shot 2022-09-14 at 9 59 44 pm


web-ui/controller/git_forge.ml
```diff
@@ -192,24 +200,17 @@ module Make (View : View) = struct
     Capability.with_ref (Client.Org.repo org_cap repo) @@ fun repo_cap ->
     Capability.with_ref (Client.Repo.commit_of_hash repo_cap hash)
     @@ fun commit_cap ->
-    Client.Commit.refs commit_cap >>!= fun refs ->
+    (* Client.Commit.refs commit_cap >>!= fun refs -> *)
```
maiste, line 203: Shouldn't this comment be removed?

web-ui/router.ml
```diff
@@ -13,6 +13,7 @@ let create ~github ~gitlab =
       Dream.get "/css/ansi.css" (fun _ ->
           Dream.respond ~headers:[ ("content-type", "text/css") ] Ansi.css);
       Dream.get "/css/**" @@ Dream.static ~loader "/css";
+      Dream.get "/images/**" @@ Dream.static ~loader "/images";
```
maiste, line 16: This is more a side note for us than a request, but we might need to separate static files in a `static` sub dir we can mount with `ocaml-crunch` at some point.

web-ui/view/github.ml
```diff
@@ -178,7 +189,12 @@ let return_link ~org ~repo ~hash =
    instead of providing a return link *)
 let list_steps ~org ~repo ~refs ~hash ~jobs ~first_step_queued_at
     ~total_run_time ?(success_msg = div []) ?(fail_msg = div [])
-    ?(return_link = div []) ?(flash_messages = []) ~csrf_token () =
+    ?(return_link = div []) ?(flash_messages = [])
+    ?(build_status : Client.State.t = Passed) ~csrf_token () =
+  (*FIXME: Fix the interface for this function so that we drop the things we are now ignoring *)
+  ignore success_msg;
+  ignore fail_msg;
+  ignore return_link;
```
maiste, line 197: Could you open an issue on [OCaml-ci](https://github.com/ocurrent/ocaml-ci/issues/new) with an explanation and replace the FIXME message with a link to the issue? It will be easier to track FIXME and todos over time this way.
(This is a suggestion, feel free to ignore ;) )

web-ui/view/timestamps_durations.ml
```diff
@@ -6,6 +6,14 @@ let to_iso8601 (tt : float) =
   let ts = Timedesc.of_timestamp_float_s tt in
   Timedesc.to_iso8601 @@ Option.get ts
 
+let to_ts_string (tt : float) =
+  let ts = Timedesc.of_timestamp_float_s tt in
+  Timedesc.to_string
+    ~format:
+      "{mon:Xxx} {day:0X} {hour:0X}:{min:0X} \
+       {tzoff-sign}{tzoff-hour:0X}:{tzoff-min:0X}"
+  @@ Option.get ts
```
maiste, line 15: Are we sure to have an `optionè here by construction? (Prefer to ask for nothing than having errors at runtime)

web-ui/view/timestamps_durations.ml
```diff
@@ -94,3 +102,9 @@ let show_build ~first_step_queued_at ~total_run_time =
                (Duration.of_f total_run_time);
         ];
     ]
+
+let pp_timestamp v = Option.fold ~none:"-" ~some:to_ts_string v
+
+let pp_duration =
+  Option.fold ~none:"-" ~some:(fun v ->
+      Fmt.str "%a" Run_time.duration_pp (Duration.of_f v))
```
maiste, line 110: Side question: what is the difference between a duration and a timestamp in this context?

maiste (commented):
Thanks for this PR :+1: 

The display seems to work fine on `Linux` with:
* Chromium `105.0.5195.125`
* Mozilla Firefox `104.0.2`
* Need to be tested on Safari to ensure everything is ok

One part the redesign might lack at some point is the mobile device version as the display is currently not supported and lead to an original design :)

Few questions:
- Will the Dockerfile copy system be implemented later?
- Same for the copy & paste system?

novemberkilo:
Yes I have left the features that require JavaScript (like selecting lines etc) to a subsequent PR.

web-ui/view/github.ml
novemberkilo: Will do.

maiste (approved):

kit-ty-kate:
is there a reason why the colors in the output log have been removed? Could we also get the black on white background back by any chance? I find the new white on black background makes it extremely hard to read the log

maiste:
Hi Kate, thanks for the report!
I can't tell you about the colors but if IIRC, we are supposed to add a separation for the Dockerfile so it can be easily copy.

About the white background, this is a good idea to have a dark mode at some point as it is very common in many website.'

novemberkilo:
@kit-ty-kate thanks for the feedback. #538 should restore colours in the logs (although with the black background they might not work as well). I've asked our designer to help with a dark/light switch so that we can make the logs also present in black on white background.

Issue #529, opened 2022-09-20
 New routes to introduce commits and history in the context of refs (and PRs and MRs)


maiste:
In the current version of the ocaml-ci, to access a specific ref, we are relying on the commit to display the build.  The route looks like this:
```
/:orga/:repo/commit/:commit_hash
```
It brings two issues:
* First, the design doesn't reflect what happens behind it and the objects we are manipulating.
* Second, it is impossible for the new design to distinguish between a branch or a PR with the same HEAD commit.

The solution is to separate the pull/merge requests from the branch refs. The new URLs would look like this:
```
/:orga/:repo/commit/:hash/ref/pull/:number
/:orga/:repo/commit/:hash/ref/mr/:number 
/:orga/:repo/commit/:hash/ref/tree/:name
```

It also allows us to bring more functionalities in the future (as an extension with `/history). Also, it is a better API to describe what we are trying to do. However, it has one drawback. We need to handle by hand the `:name` as there might be some `subject/area/content` branches. That's why we must introduce the `-` in the loop. This way, we can easily match the pattern in the URL.

novemberkilo:
Note I am suggesting in #530 that we have a leading hyphen to separate the ref data. This may not be necessary but for now, it allows us to continue to support routes like 

```
/:orga/:repo/commit/:hash/[cancel,rebuild]
```

while still allowing us to use wildcard matching.

maiste:
I agree :+1: It's import to keep legacy routes.

novemberkilo:
I'm closing this as done seeing that we've had these routes in production since September, without incident.

PR #530, opened 2022-09-21
Adds refs to the urls used in the checks API.

novemberkilo:
Possibly useful for #529 // @maiste

### DIFF
--- a/service/pipeline.ml
+++ b/service/pipeline.ml
@@ -26,15 +26,15 @@ let platforms =
   Current.list_seq (List.map v v2_1)
 
 (* Link for GitHub statuses. *)
-let url ~owner ~name ~hash =
+let url ~owner ~name ~hash ~gref =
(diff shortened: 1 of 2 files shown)


novemberkilo:
This has problems. 

For the commits that appear on a branch that is also in a Pull Request, some commits will have a `gref` of `refs/pull/X/head` (these will correspond to when we deploy the service and thus pick up the head of the PR ref) and some will have a `gref` of `refs/heads/<branch-name>`

This seems to be a consequence of reading the ref from an in-memory map.(?)

novemberkilo:
As discussed we're going to use this to experiment and see if this implementation is viable or not. So as to not break anything, there is now support for these new routes, and we just drop the additional ref data and redirect to the (usual) build page.

maiste (approved):
All good for me :+1:

PR #538, opened 2022-09-22
Cleanup small things in build & step pages

novemberkilo:
- [x] "0s" instead of "0s000ms" 
- [x] Attempt to fix spurious blank lines appearing in the logs (when waiting on the stream)
- [x] Restore colour in logs via `ansi.css`

### DIFF
--- a/web-ui/view/github.ml
+++ b/web-ui/view/github.ml
@@ -388,21 +388,28 @@ let show_step ~org ~repo ~refs ~hash ~jobs ~variant ~job ~status ~csrf_token
   in
   let ansi = Ansi.create () in
   let line_number = ref 0 in
+  let last_line_blank = ref false in
   let decorate data : string =
-    let aux l log_line =
-      line_number := !line_number + 1;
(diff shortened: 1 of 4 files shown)


web-ui/view/github.ml
```diff
@@ -390,19 +390,28 @@ let show_step ~org ~repo ~refs ~hash ~jobs ~variant ~job ~status ~csrf_token
...
-              ]))
+      match log_line with
+      | "" ->
+          l
+          (* Don't include empties.
+             We seem to get these while we wait on the stream *)
+      | log_line ->
+          line_number := !line_number + 1;
+          Fmt.str "%s\n%s" l
+            (Fmt.to_to_string (pp_elt ())
```
maiste, line 401: Why not use an `%a` here as you rely on the formatter to produce the result?

tmcgilchrist: That would be more idiomatic use of Fmt

tmcgilchrist (approved):

PR #539, opened 2022-09-22
API: improve the history lookup

maiste:
This PR improves the previous PR (#526) by changing the query run behind to fetch the data. It is more accurate but it involves a small decrease in performance as we have to parse some data by hand.


novemberkilo (approved):
🚀

PR #541, opened 2022-09-26
Enable highlight-lines and copy-link feature for logs.

novemberkilo:


web-ui/view/step.ml
```diff
@@ -52,3 +52,79 @@ let title_card ~status ~card_title ~hash_link ~created_at ~finished_at
             rebuild_button;
           ];
       ])
+
+let log_highlight_js =
+  Tyxml.Html.script ~a:[]
+    (Tyxml.Html.Unsafe.data
+       "\n\
```
MisterDA, line 59: Maybe a quoted string literal would be more readable here? No backslashes and no `\n`.
https://v2.ocaml.org/manual/lex.html#sss:stringliterals
Or you could use [crunch](https://github.com/mirage/crunch) to embed the data.

novemberkilo: I couldn't get ocamlformat to leave it like that. How can I annotate a function so that ocamlformat ignores it?

MisterDA: > I couldn't get ocamlformat to leave it like that.

Even with `{|my string|}` syntax?

> How can I annotate a function so that ocamlformat ignores it?

https://ocaml.org/p/ocamlformat/0.24.1/doc/faq.html#how-to-locally-disable-ocamlformat?

novemberkilo: So the quoted string literal worked well. Thanks again @MisterDA 👍🏾

web-ui/view/step.ml
```diff
@@ -52,3 +52,87 @@ let title_card ~status ~card_title ~hash_link ~created_at ~finished_at
             rebuild_button;
           ];
       ])
+
+let log_highlight_js =
+  Tyxml.Html.script ~a:[]
+    (Tyxml.Html.Unsafe.data
+       {|"
```
MisterDA, line 59: I think you just want the `{|` delimiter and not the opening quote. For quoted string literals, the delimiters are just `{|` and `|}`, if you put a quote it'll be part of the string content.

PR #544, opened 2022-09-28
Add benmandrew as user of ocaml-ci

benmandrew:


mtelvers:
I have also run `make deploy-stack` to make these changes live.

tmcgilchrist:
Also needs a push to `live-engine` to sync the branches @benmandrew @mtelvers

PR #545, opened 2022-10-03
Gitlab docs

tmcgilchrist:
Make both pipelines optional and update the index page to not show links for disabled backends

### DIFF
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -24,11 +24,7 @@ Pull request
 Push
 ```
 
-### Running a scheduler and a worker
-
-You will need to run a scheduler and a worker to connect it to the CI. Follow the instruction from [ocurrent/ocluster](https://github.com/ocurrent/ocluster#the-scheduler-service).
-
-### Running the server locally
+### Running the GitHub pipeline locally
 
 You will need the following:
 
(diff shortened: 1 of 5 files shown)


doc/dev.md
```diff
@@ -52,14 +48,37 @@ This will generate a capability file. See the logs for `Wrote capability referen
...
+
 ### Running the web client locally
 
 Using the capability file written out by the service, run the web client as follows:
 
 ```
 dune exec -- ocaml-ci-web \
-  --backend ./capnp-secrets/ocaml-ci-admin.cap
+  --backend ./capnp-secrets/ocaml-ci-admin.cap \
+  --gitlab-backend ./capnp-secrets/ocaml-ci-gitlab-admin.cap
```
novemberkilo, line 76: Given that this PR also introduces the change to `ocaml-ci-web` so that `gitlab-backend` is optional, is it worth mentioning that here?

novemberkilo (approved):
Thanks for this @tmcgilchrist

web-ui/router.ml
```diff
@@ -8,199 +8,194 @@ let loader root path _request =
...
+              ~repo:(Dream.param request "repo")
+              ~hash:(Dream.param request "hash")
+              request gitlab
+        | _ ->
+            Dream.log "Form validation failed";
+            Dream.empty `Bad_Request);
+  ]
+
 let create ~github ~gitlab =
+  let ( ++ ) = List.append in
```
maiste, line 90: Why not use `@` instead?

maiste (commented):
I agree with @novemberkilo comment. 
Also, isn't there a way to express: "GitHub, GitLab, or both"? Maybe we should make the GitHub one optional too? This way, the user can design to run the instance they? WDYT,

web-ui/router.ml
MisterDA: Haskellers in hiding!

tmcgilchrist: Caught! I really want an ad-hoc way to introduce infix operators.

doc/dev.md
tmcgilchrist: I added a mention here https://github.com/ocurrent/ocaml-ci/pull/545/files#diff-a207a689efababf8e9874e3fa823535a3444e998809ad8d71b0b69a625ffaf58R79

PR #546, opened 2022-10-03
Create cap-secrets if not found

maiste:
When you are trying to start ocaml-ci, and the `cap-secrets` repository doesn't exist, it will raise this issue:
```
ocaml-ci-service: internal error, uncaught exception:
         Sys_error("./capnp-secrets/secret-key.pem: No such file or directory")
```
This PR intends to fix this behaviour by creating the dedicated repository only when required, and we are sure not to override anything.

### DIFF
--- a/gitlab/main.ml
+++ b/gitlab/main.ml
@@ -53,6 +53,21 @@ let setup_log default_level =
 
 let or_die = function Ok x -> x | Error (`Msg m) -> failwith m
 
+let check_dir x =
+  Lwt.catch
+    (fun () ->
+      Lwt_unix.stat x >|= function
+      | Unix.{ st_kind = S_DIR; _ } -> `Present
+      | _ -> Fmt.failwith "Exists, but is not a directory: %S" x)
+    (function
+      | Unix.Unix_error (Unix.ENOENT, _, _) -> Lwt.return `Missing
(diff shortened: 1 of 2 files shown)


maiste:
@tmcgilchrist I moved it to Lwt as you suggested :+1:

MisterDA:
There is essentially the same code in OBuilder: [`Os.ensure_dir`](https://github.com/ocurrent/obuilder/blob/671a2400528f29736dc446813f0699055d8e631b/lib/os.ml#L140-L149).
I suggest adopting the same name and code (perhaps converted to Lwt) for consistency.

```ocaml
let check_dir x =
  match Unix.lstat x with
  | Unix.{ st_kind = S_DIR; _ } -> `Present
  | _ -> Fmt.failwith "Exists, but is not a directory: %S" x
  | exception Unix.Unix_error(Unix.ENOENT, _, _) -> `Missing


let ensure_dir path =
  match check_dir path with
  | `Present -> ()
  | `Missing -> Unix.mkdir path 0o777
```

MisterDA:
…although the code is simple enough so we don't have to care too much.

maiste:
Good idea! However, I'd keep the `lwt` version.

tmcgilchrist (approved):

PR #547, opened 2022-10-04
Allow unauthenticated access to the metrics endpoint.

tmcgilchrist:
The main change here is to align the auth check between GitHub and GitLab, in particular the `/metrics` endpoint used by prometheus was unnecessarily requiring authentication. 

Second I explicitly matched on variants in match statements, both to make it clearer what the roles are and for completeness.

### DIFF
--- a/gitlab/pipeline.ml
+++ b/gitlab/pipeline.ml
@@ -4,6 +4,18 @@ module Git = Current_git
 module Gitlab = Current_gitlab
 module Docker = Current_docker.Default
 
+module Metrics = struct
+  open Prometheus
+
+  let namespace = "ocurrent"
+  let subsystem = "gitlab"
+
+  let repositories_total =
+    let help = "Total number of active repositories" in
+    Gauge.v_label ~label_name:"account" ~help ~namespace ~subsystem
+      "repositories_total"
+end
+
 let platforms =
   let schedule = Current_cache.Schedule.v ~valid_for:(Duration.of_day 30) () in
   let v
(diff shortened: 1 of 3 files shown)


gitlab/pipeline.ml
```diff
@@ -4,6 +4,18 @@ module Git = Current_git
 module Gitlab = Current_gitlab
 module Docker = Current_docker.Default
 
+module Metrics = struct
+  open Prometheus
+
+  let namespace = "ocurrent"
```
maiste, line 10:
```suggestion
  let namespace = "ocaml-ci"
```

gitlab/pipeline.ml
```diff
@@ -136,6 +148,9 @@ let set_active_repos ~(installation : Installation.t Current.t)
        (fun acc r -> Index.Repo_set.add r.Gitlab.Repo_id.name acc)
        Index.Repo_set.empty
   |> Index.set_active_repos ~owner:installation.Installation.name;
+  Prometheus.Gauge.set
+    (Metrics.repositories_total installation.Installation.name)
+    (float_of_int (List.length repos));
```
maiste, line 153: Shouldn't this setting system be hidden into the metrics, so we don't have to worry about the type? Like:
```ocaml
Metrics.set_repositories_total installation.Installation.name (float_of_int (List.length repos));
```
This way we have more control over the metrics inside the `Metrics` module. WDYT?

maiste (approved):
Two small questions but otherwise, LGMT :+1:

gitlab/pipeline.ml
tmcgilchrist: This aligns with the equivalent metrics in the GitHub plugin https://github.com/ocurrent/ocurrent/blob/master/plugins/github/installation.ml#L4-L13

In future the GitLab plugin will have a table of registered org/repos that can feed this Metric.

maiste: Oh, is it because this is considered to be `current_gitlab`? :thinking:

tmcgilchrist: Basically yes.

gitlab/pipeline.ml
tmcgilchrist: Sure, when this gets moved into the ocurrent plugin.

PR #551, opened 2022-10-12
First pass at live updates of step page.

novemberkilo:
Uses an approach of polling the server every 10s for a json representation of the step.

The polling stops when the build status is "passed" or "failed" or we time out (set at 45m right now).

Got some console.logs etc. that needs cleaning up etc. but would appreciate a look pls.

### DIFF
--- a/web-ui/controller/git_forge.ml
+++ b/web-ui/controller/git_forge.ml
@@ -2,6 +2,7 @@ module type View = View.Git_forge.View
 
 module Client = Ocaml_ci_api.Client
 module Run_time = Ocaml_ci_client_lib.Run_time
+module Timestamps_durations = View.Timestamps_durations
 
 module type Controller = sig
   val list_orgs : Backend.t -> Dream.response Lwt.t
@@ -33,6 +34,14 @@ module type Controller = sig
     Dream.client Dream.message ->
     Backend.t ->
     Dream.server Dream.message Lwt.t
+
+  val show_step_json :
+    org:string ->
+    repo:string ->
+    hash:string ->
+    variant:string ->
+    Backend.t ->
+    Dream.response Lwt.t
 
   val rebuild_step :
     org:string ->
@@ -66,6 +75,24 @@ module Make (View : View) = struct
   open Lwt.Infix
   module Client = Ocaml_ci_api.Client
   module Capability = Capnp_rpc_lwt.Capability
+
+  type step_t = {
+    status : string;
+    created_at : string;
+    finished_at : string;
+    queued_for : string;
+    ran_for : string;
+  }
+
+  let yojson_of_step_t t =
+    `Assoc
+      [
+        ("status", `String t.status);
+        ("created_at", `String t.created_at);
+        ("queued_for", `String t.queued_for);
+        ("finished_at", `String t.finished_at);
+        ("ran_for", `String t.ran_for);
+      ]
 
   let ( >>!= ) x f =
     x >>= function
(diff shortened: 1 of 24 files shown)


web-ui/view/common.ml
```diff
@@ -86,14 +86,17 @@ let form_for ~x_ref ~action ~csrf_token ~submit_button ~input_value =
           ();
       ])
 
-let form_rebuild_step ~variant ~csrf_token =
+let form_rebuild_step ~variant ~csrf_token ?(show = true) () =
+  let display_none = if show then "" else "display:none" in
```
maiste, line 90: image
The button doesn't seem to be redrawn once the step is finished.

web-ui/controller/git_forge.ml
```diff
@@ -281,4 +308,70 @@ module Make (View : View) = struct
...
+    let jobs = Client.Commit.jobs commit_cap in
+    Capability.with_ref (Client.Commit.job_of_variant commit_cap variant)
+    @@ fun job_cap ->
+    jobs >>!= fun jobs ->
+    Capability.inc_ref job_cap;
+    let build_created_at =
+      Option.value ~default:0.
+      @@ Option.join
+      @@ Result.to_option
+      @@ Run_time.build_created_at ~build:jobs
```
maiste, line 327:
```suggestion
      Run_time.build_created_at ~build:jobs
      |> Result.to_option
      |> Option.join
      |> Option.value ~default:0.
```
IMHO, The composition style with too many functions hides what happens underneath. Feel free to ignore the comment.

web-ui/controller/git_forge.ml
```diff
@@ -281,4 +308,70 @@ module Make (View : View) = struct
...
+    in
+    let step_info =
+      let filter (j : Client.job_info) = j.variant = variant in
+      List.find_opt filter jobs
+    in
+    let build_status =
+      Option.value ~default:(Undefined 1)
+        (Option.map (fun i -> i.Client.outcome) step_info)
```
maiste, line 335:
```suggestion
      Option.fold ~none:(Undefined 1) ~some:(fun i -> i.Client.outcome) step_info
```
:wink:

web-ui/controller/git_forge.ml
```diff
@@ -281,4 +308,70 @@ module Make (View : View) = struct
...
+                 (fun i ->
+                   Timestamps_durations.pp_timestamp i.Client.finished_at)
+                 step_info);
+          queued_for =
+            Timestamps_durations.pp_duration
+              (Option.map Run_time.queued_for run_time);
+          ran_for =
+            Timestamps_durations.pp_duration
+              (Option.map Run_time.ran_for run_time);
+        }
```
maiste, line 374: For clarity, this function could be extracted in another function.
```suggestion
      yojson_of_step_t
        {
          status = Fmt.str "%a" Client.State.pp build_status;
          created_at =
            Option.fold ~none:""
              ~some:(fun i -> Timestamps_durations.pp_timestamp i.Client.queued_at)
                 step_info;
          finished_at =
            Option.fold ~none:""
              ~some: (fun i ->Timestamps_durations.pp_timestamp i.Client.finished_at) step_info;
          queued_for =
            Timestamps_durations.pp_duration
              (Option.map Run_time.queued_for run_time);
          ran_for =
            Timestamps_durations.pp_duration
              (Option.map Run_time.ran_for run_time);
        }
```

maiste (commented):
Thanks for this work; it looks great! 
I have a small issue with the rebuild button printing. Otherwise, it's some code suggestions that you can ignore if they don't seem relevant to you.

web-ui/view/common.ml
novemberkilo: I think that in this particular case, since the build has passed, there won't be a Rebuild button. The Rebuild button should only be redrawn for failed builds.

web-ui/view/step.ml
```diff
@@ -135,4 +142,77 @@ document.addEventListener('alpine:init', () => {
         }
     }))
 })
+
+|})
+
+let poll =
+  Tyxml.Html.script ~a:[]
+    (Tyxml.Html.Unsafe.data
+       {|
```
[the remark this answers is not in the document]
novemberkilo: There are a few options being considered - the main one is to use jsoo. The approach with using crunch etc. is possible as well. For now, I'd like to leave this as it is and take @tmcgilchrist up on his offer to jump in and see if we can introduce jsoo machinery.

benmandrew (commented):
Looks good to me, it works on my machine and the rebuild button is redrawn on a failed build.

web-ui/controller/api.ml
```diff
@@ -0,0 +1 @@
+include Git_forge.Make_API
```
tmcgilchrist, line 1: I would avoid the `Make` word here, it suggests a Functor to me.

web-ui/controller/git_forge.ml
```diff
@@ -1,8 +1,9 @@
+open Lwt.Infix
```
tmcgilchrist, line 1: Check your usages of this module, some places you re-open it for a single function.
Ideally I would keep the scope this is open in as short as possible.

web-ui/representation/dune
```diff
@@ -0,0 +1,6 @@
+(library
+ (public_name ocaml-ci-web.representation)
```
tmcgilchrist, line 2: Not crazy on this naming what about `model` or `types` ?
Not really better names but they are shorter 🤣

web-ui/representation/step.ml
```diff
@@ -0,0 +1,17 @@
+type t = {
+  status : string;
+  created_at : string;
+  finished_at : string;
+  queued_for : string;
+  ran_for : string;
+}
```
tmcgilchrist, line 7: Maybe just use `[@@deriving yojson]` and avoid a hand written function.
Would be nice to have simple tests on this since it forms a kind of API.

web-ui/controller/git_forge.ml
```diff
@@ -69,42 +62,34 @@ module type Controller = sig
     Backend.t ->
     Dream.server Dream.message Lwt.t
 end
-
 (* Abstract controller for any Git_forge that implements `Ocaml_ci_api.Client` API *)
+
+module type API_Controller = sig
```
tmcgilchrist, line 67: You don't seem to be using this module type anywhere?

web-ui/controller/git_forge.ml
```diff
@@ -308,8 +293,16 @@ module Make (View : View) = struct
     in
     List.iter (fun (s, m) -> Dream.add_flash_message request s m) flash_messages;
     Dream.redirect request (Fmt.str "/github/%s/%s/commit/%s" org repo hash)
+end
+
+module Make_API = struct
```
tmcgilchrist, line 298: Add a top-level comment saying what this module is for and why?

novemberkilo:
Ok I am ready to ship this in its current form and continue with getting naming and structure in a better place as we do the same for the build page. // @maiste @tmcgilchrist

PR #553, opened 2022-10-12
History page

benmandrew:
The history page UI is done, but currently using placeholder commit titles.

The logic for retrieving the titles has been implemented to just before the DB query, as this would require changing the DB schema which we discussed we would leave until the migration system is done.

### DIFF
--- a/lib/index.mli
+++ b/lib/index.mli
@@ -56,7 +56,7 @@ val get_build_history_with_time :
   owner:string ->
   name:string ->
   gref:string ->
-  (string * string * float option) list
+  (string * string * string * float option) list
 (** [get_build_history_with_time ~owner ~name ~gref] is a list of builds for the
     branch gref of the repo identfied by (owner, name). The builds are
     identified by triples (variant, hash, Some timestamp) *)
--- a/lib/index.mli
+++ b/lib/index.mli
@@ -11,6 +11,13 @@ type job_state =
 
 type build_status = [ `Not_started | `Pending | `Failed | `Passed ]
 
+type ref_info = {
+  hash: string;
+  message: string;
+  gref: string;
+  started_at: float option;
+}
+
 val init : unit -> unit
 (** Ensure the database is initialised (for unit-tests). *)
 
@@ -18,6 +25,7 @@ val record :
   repo:Repo_id.t ->
   hash:string ->
   status:build_status ->
+  message:string ->
   gref:string ->
   (string * Current.job_id option) list ->
   unit
@@ -43,23 +51,16 @@ val get_job :
 
 val get_job_ids : owner:string -> name:string -> hash:string -> string list
 
+val get_message : owner:string -> name:string -> hash:string -> string
+
 val get_build_history :
   owner:string ->
   name:string ->
   gref:string ->
-  (string * string * string option) list
-(** [get_build_history ~owner ~name ~gref] is a list of builds for the branch
-    gref of the repo identfied by (owner, name) The builds are identified by
-    triples (variant, hash, Some job_id) *)
-
-val get_build_history_with_time :
-  owner:string ->
-  name:string ->
-  gref:string ->
-  (string * string * string * float option) list
-(** [get_build_history_with_time ~owner ~name ~gref] is a list of builds for the
+  ref_info list
+(** [get_build_history ~owner ~name ~gref] is a list of builds for the
     branch gref of the repo identfied by (owner, name). The builds are
-    identified by triples (variant, hash, Some timestamp) *)
+    identified by ref_info *)
 
 val get_status : owner:string -> name:string -> hash:string -> build_status
 (** [get_status ~owner ~name ~hash] is the latest status for this combination. *)
--- a/service/api_impl.ml
+++ b/service/api_impl.ml
@@ -95,11 +95,12 @@ let make_commit ~engine ~owner ~name hash =
           | `Passed -> Results.status_set results Passed);
          Service.return response
 
-        method title_impl _params release_param_caps =
-          let open Commit.Title in
+        method message_impl _params release_param_caps =
+          let open Commit.Message in
           release_param_caps ();
           let response, results = Service.Response.create Results.init_pointer in
-          Results.title_set results "Commit Title (API)";
+          let message = Index.get_message ~owner ~name ~hash in
+          Results.title_set results message;
           Service.return response
      end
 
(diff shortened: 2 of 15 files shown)


lib/index.mli
```diff
@@ -56,7 +56,7 @@ val get_build_history_with_time :
   owner:string ->
   name:string ->
   gref:string ->
-  (string * string * float option) list
+  (string * string * string * float option) list
```
maiste, line 59: Maybe we need to introduce a type to describe the `refInfo` as we get more and more string. Something like:
```ocaml
type ref_info = {
  title: string;
  hash: string;
  gref: string;
  started_at: float option
}
```

web-ui/view/gitlab.ml
```diff
@@ -116,12 +122,53 @@ let list_refs ~org ~repo ~refs =
...
+        else Fmt.str "Builds (%d)" len) ]
+  in
+  let commit_table =
+    List.fold_left
+      (fun l (u, title, status, t) ->
+        let created_at = Timestamps_durations.pp_timestamp (Some t) in
+        List.append l
+          [
+            Build.commit_row ~commit_title:(title) ~short_hash:(short_hash u) ~created_at ~status ~commit_uri:(commit_url ~org ~repo u);
+          ])
```
maiste, line 147:
```suggestion
   let f (u, commit_title, status, t) =
        let created_at = Timestamps_durations.pp_timestamp (Some t) in
        let short_hash = short_hash u in
        let commit_uri = commit_url ~org ~repo u in
        Build.commit_row ~commit_title ~short_hash ~created_at ~status ~commit_uri
   in List.map f
```
Isn't it a `List.map` instead of a `List.fold_left`?

web-ui/view/gitlab.ml
```diff
@@ -116,12 +122,53 @@ let list_refs ~org ~repo ~refs =
...
+  in
+  let commit_table_div =
+    div
+      ~a:
+        [ a_class [ "bg-gray-50 px-6 py-3 text-gray-500 text-xs font-medium" ] ]
+        (* TODO: We need to start with no stage separation - introduce Analysis/Checks and Build steps later *)
+      [ txt (
+        let len = List.length history in
+        if len = 1 then "Build (1)"
+        else Fmt.str "Builds (%d)" len) ]
```
maiste, line 138:
```suggestion
        let len = List.length history in
        Fmt.str "Builds (%d)" len) ]
```
If we follow the same conventions as `GitHub`, we don't have to bother about it.

web-ui/view/gitlab.ml
```diff
@@ -25,6 +25,21 @@ let gitlab_mr_url ~org ~repo id =
...
+  | _ -> Fmt.str "Bad ref format %S" r
+
+let ref_breadcrumb r head_hash =
+  match Astring.String.cuts ~sep:"/" r with
+  | "refs" :: "heads" :: branch ->
+      let branch = Astring.String.concat ~sep:"/" branch in
+      (branch, Fmt.str "commit/%s" head_hash)
+  | [ "refs"; "pull"; id; "head" ] ->
+      ("#" ^ id, Fmt.str "pull/%s" head_hash)
+  | _ -> (Fmt.str "Bad ref format %S" r, "")
```
maiste, line 41: GitLab doesn't use the same format for `Merge_request`, you need to check but it's something like `refs/merge_requests/<id>/head`.

maiste:
```suggestion
   let f (u, commit_title, status, t) =
        let created_at = Timestamps_durations.pp_timestamp (Some t) in
        let short_hash = short_hash u in
        let commit_uri = commit_url ~org ~repo u in
        Build.commit_row ~commit_title ~short_hash ~created_at ~status ~commit_uri
   in List.map f
```
Isn't it a `List.map` instead of a `List.fold_left`?

service/api_impl.ml
```diff
@@ -94,6 +94,13 @@ let make_commit ~engine ~owner ~name hash =
           | `Failed -> Results.status_set results Failed
           | `Passed -> Results.status_set results Passed);
          Service.return response
+
+        method title_impl _params release_param_caps =
+          let open Commit.Title in
+          release_param_caps ();
+          let response, results = Service.Response.create Results.init_pointer in
+          Results.title_set results "Commit Title (API)";
```
maiste, line 102: Can you add a `Fixme` here, so we don't forget to change it, please? :wink:

maiste (changes requested):
The PR might need to be rebased before merging the PR.

Few comments:
- The branch link leads to a `404 Error` page. 
- In the `api/client.ml`, could you add a `title` function to fetch the answer, as described in the `schema.capnp`? This way, we would only have to update the data gathered from`index.ml` to get the true title.

novemberkilo (commented):
Nicely done @benmandrew - thanks for this.

A couple of questions/comments:

- You mention that "The logic for retrieving the titles has been implemented to just before the DB query" -- I'm assuming that the title for a build is the commit message, is that right? From what I can see, [we are collecting the message](https://github.com/ocurrent/ocurrent/blob/master/plugins/github/api.mli#L28) - is the intention to wire this into `Index.record` so that we persist the commit message?

- @maiste has gone through the detail (thanks Etienne) and I've been nodding along so I am not going to double up on that.

benmandrew:
@novemberkilo for your first point, yes that is the intention. Once the schema is changed the message will be put into the DB, and then it will be retrieved by the UI when needed. The same is intended for commit statuses.

All of the logic post-retrieval is implemented, at the point where the DB query is made I insert the placeholder. I will clean up the other placeholders ('API' and 'refs') to make it clearer.

web-ui/view/github.ml
```diff
@@ -27,6 +27,21 @@ let github_pr_url ~org ~repo id =
...
+  match Astring.String.cuts ~sep:"/" r with
+  | "refs" :: "heads" :: branch -> Astring.String.concat ~sep:"/" branch
+  | [ "refs"; "pull"; id; "head" ] -> id
+  | _ -> Fmt.str "Bad ref format %S" r
+
+let ref_breadcrumb r head_hash =
+  match Astring.String.cuts ~sep:"/" r with
+  | "refs" :: "heads" :: branch ->
+      let branch = Astring.String.concat ~sep:"/" branch in
+      (branch, Fmt.str "commit/%s" head_hash)
```
MisterDA, line 40: You don't need `Fmt` here, use `Printf.sprintf` (or an alias). `Fmt` has the full power of the formatting engine but you're just creating simple strings with simple formats.

benmandrew:
Implemented changes + implemented a placeholder organisations page.

maiste:
Could you apply a  `dune fmt` to make the CI ok, please? :smile:

PR #555, opened 2022-10-18
Update submodules (routes.2.0.0, OBuilder Windows spec, Alpine)

MisterDA:
[Update ocurrent](https://github.com/ocurrent/ocaml-ci/commit/dd7219970c4fa25c919cd285e2cdeb3ce03a5eb1)

- routes.2.0.0

[Update ocaml-dockerfile](https://github.com/ocurrent/ocaml-ci/commit/d4c7d7fc79ca14f7e85cfaf5081b93a96bb2491f)

- Alpine 3.16 tier 1, 3.15 tier 2, 3.14 deprecated.

[Update ocluster](https://github.com/ocurrent/ocaml-ci/commit/52c2f3a48d4ebfc9b12c30d540df73d9258c8678)

- `Obuilder_spec.Docker.dockerfile_of_spec` now takes the os as
  parameter for escape style;
- `Obuilder_spec` now differentiates Unix and Windows users.


maiste:
Everything works fine apart from one thing: it relies on submodules, so if you don't pine `routes` to `2.0.0` the `dune build` will fail as we install the deps from `opam`. We should consider either doing a release of `ocurrent` as `6.3.0` or adding an explicit version of `routes` in `ocaml-ci`. IMHO, the first one is the best as it is required in many projects.

MisterDA:
After an offline exchange we've settled on a third option: tweaking the installation instructions to also install the submodules dependencies.

PR #558, opened 2022-10-19
Support migrations into OCaml-ci

maiste:
This PR intends to support migrations into `ocaml-ci` using `omigrate`.
Be aware that it requires more tests to ensure everything works fine.

:warning: If you plan to test it from home, create a copy of the  `var/` so you won't lose everything you have in the DB if it fails.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -22,7 +22,7 @@
    logs
    current
    current_docker
-   (omigrate (>= 0.3.0))
+   (omigrate (>= 0.3.2))
    ocaml-ci-api
    capnp-rpc-unix
    ppx_deriving_yojson
--- a/dune-project
+++ b/dune-project
@@ -22,6 +22,7 @@
    logs
    current
    current_docker
+   (omigrate (>= 0.3.0))
    ocaml-ci-api
    capnp-rpc-unix
    ppx_deriving_yojson
--- a/gitlab/main.ml
+++ b/gitlab/main.ml
@@ -125,7 +125,7 @@ module Gitlab = struct
 end
 
 let main () config mode app capnp_public_address capnp_listen_address
-    gitlab_auth submission_uri : ('a, [ `Msg of string ]) result =
+    gitlab_auth submission_uri no_migration : ('a, [ `Msg of string ]) result =
   Lwt_main.run
     (let solver = Ocaml_ci.Solver_pool.spawn_local () in
      run_capnp capnp_public_address capnp_listen_address
@@ -133,9 +133,10 @@ let main () config mode app capnp_public_address capnp_listen_address
      let ocluster =
        Option.map (Capnp_rpc_unix.Vat.import_exn vat) submission_uri
      in
+     let migration = no_migration in
      let engine =
        Current.Engine.create ~config
-         (Ocaml_ci_gitlab.Pipeline.v ?ocluster ~app ~solver)
+         (Ocaml_ci_gitlab.Pipeline.v ?ocluster ~app ~solver ~migration)
      in
      rpc_engine_resolver
      |> Option.iter (fun r ->
@@ -194,6 +195,13 @@ let submission_service =
   @@ Arg.info ~doc:"The submission.cap file for the build scheduler service"
        ~docv:"FILE" [ "submission-service" ]
 
+let disable_migrations =
+  Arg.(
+    value
+    @@ flag
+    @@ info ~doc:"Disable the system to execute the migrations."
+         [ "disable-migrations" ])
+
 let cmd =
   let doc = "Build OCaml projects on GitLab" in
   let info = Cmd.info "ocaml-ci-gitlab" ~doc in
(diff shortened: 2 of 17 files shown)


novemberkilo (commented):
@maiste thanks much for this work :) Great to see omigrate repurposed to support sqlite and being used in ocaml-ci. 

I notice that at the moment you've chosen to run migrations at startup of ocaml-ci-service. This is ... okay because sqlite is local to the app and pretty much an extension of it ... but it's a somewhat dangerous pattern to follow with migrations IMO. The danger comes from the fact that if you tried to do this with postgres or mysql or any other database that runs as a service, and you have multiple instances of ocaml-ci-service in production, then things won't work. I think that this is one reason why folks want to separate out migrations from the normal startup of the app by running migrations via a command line tool (iirc omigrate has a cli, no?) ... and possibly having the app check the migrations table at startup and report on migrations that appear not to have been run (instead of just running them). Amongst other things, this will also make it a lot easier for developers to have control of when to run migrations when they are introducing new ones.
 
I don't think that this is a pre-condition to merging this work. But I think that it would be very useful to think about this some more at this stage, and being clear on when/what you might do in response.

maiste:
Thanks!

I hear your concerns and agree with most of them regarding the kind of systems you describe. 

In our current design for `ocaml-ci`, it looks like this:
```ascii
+------------------+
|                  |
|  OCluster        |
|                  |
+---+--------^-----+
    |        |
    |        |
+---v--------+------+          +-----------+    +---------------------------+
|                   +---------->           <----+                           |
| OCaml-ci-service  |          | Database  |    | OCaml-ci-service-gitlab   |
|                   <----------+           +---->                           |
+----+------^-------+          +-----------+    +------+------------^-------+
     |      |                                          |            |
     |      |                                          |            |
     |      |                                          |            |
     |      |                                          |            |
     |      |                                          |            |
+----v------+------------------------------------------v------------+---------+
|                                                                             |
|              OCaml-ci-web                                                   |
|                                                                             |
+-----------------------------------------------------------------------------+
```
The database only depends on the `ocaml-ci-service` and `ocaml-ci-gitlab` systems. FWIU, the migrations should be something able to be applied if needed and only once. As both services rely on the same database, the migration will be applied by the first running service and only once as it is atomic (thanks to the transaction system). If we start to move to a microservice, our infrastructure should be a problem. Otherwise, I don't see how it is an issue here. Am I missing the point? :thinking:

novemberkilo:
@maiste as you point out this is not really an issue. 

I agree that migrations will be applied the way you are describing and for now this is ok. 

In general (and yep this only really becomes relevant when we move to something like postgres and have multiple instances of our services running), we'll want to have a tool that runs migrations *before* the code that uses the migrations is deployed. 

Also, when doing development work that introduces migrations, you typically want control over when the migration runs -- often you'll want to run up and down migrations on your local db as you are writing code and having migrations run as part of the initialisation of the service can get in the way.

Given that `omigrate` has a CLI ... would it make sense to use it to apply migrations?

maiste:
From an offline discussion with @tmcgilchrist, we reach the point that we can keep the automatic system as, otherwise, it will be a line in the Dockerfile, and the behaviour will be the same. The idea is to introduce a command line flag to run or disable it. This way, it will be possible if people want to run it without the migrations.

In our case, it's a little bit tricky because of the system architecture. If we didn't rely on the same database as the caching, we could run it from the outside world. However, we need the caching system to deal with it as it's done in `ocurrent` and not in the `ocaml-ci`. It means that the migrations are in charge of updating extra tables, not the database.  If we controlled the entire chain, I would agree that the migrations system should be applied by an extra program and ensure everything is correct. But here, I feel we don't have that luxury :disappointed:

maiste:
I introduce the disabling flag. With the discussion offline, we have realized that, in the end, it's not possible to make the system external as we store the `sqlite` content on a Docker volume. This PR is ready for review. It needs to be heavily test before deployment. @novemberkilo @tmcgilchrist could you try the migration system on your side to ensure it doesn't erase the past data into the database?

novemberkilo (approved):
@maiste thank you for this work :)

I can confirm that the `--disable-migrations` flag works as expected. Also, running migrations on an existing database has no effect on the data in the `ci_build_index` table. (The `schema_migrations` table is still created if its not there in the first place).

Can I suggest that we document a couple of things as we bring this to master:

- updates to the README or dev documentation so that the usage of `--disable-migrations` flag is clear.
- a section on migrations - how to write them, where they live, etc. Default behaviour and how to do dev work with outstanding migrations are present
- Documenting the design choice of not using the cli and instead using the `--disable-migrations` flag. The comments from you on the PR regarding this are valuable but a summary should be captured in our docs.

Up to you if you want to add the documentation before merging this, or in a consequent PR.

maiste:
Deployment is planned for `Thursday, November, 17th, 9:00 CET`.

PR #559, opened 2022-10-21
Organisations, Repos, and Refs pages

benmandrew:
Refs, repos, and orgs pages almost completed. Modified capnp schema to pass up aggregated data for summaries. A few modifications left to make but it's at a good point to take a look.

The capnp schema is quite heavily modified so that the relevant aggregated data is available; this may cause issues with deployment but this can be potentially discussed on Monday.

### DIFF
--- a/api/schema.capnp
+++ b/api/schema.capnp
@@ -114,7 +114,7 @@ interface Org {
 
 struct OrgInfo {
   owner        @0 :Text;
-  bio          @1 :Text;
+  description  @1 :Text;
   nRepos       @2 :UInt16;
   # Someone is unlikely to have more than 65536 repos
 }
--- a/api/schema.capnp
+++ b/api/schema.capnp
@@ -17,7 +17,7 @@ struct RefInfo {
     ts        @3 :Float64;
     none      @4 :Void;
   }
-  title       @5 :Text;
+  message     @5 :Text;
   # The state of the ref's head commit
 }
 
@@ -61,16 +61,16 @@ struct JobInfo {
 
 
 interface Commit {
-  jobs  @0 () -> (jobs :List(JobInfo));
+  jobs         @0 () -> (jobs :List(JobInfo));
 
   jobOfVariant @1 (variant :Text) -> (job :OCurrent.Job);
 
-  refs @2 (hash :Text) -> (refs :List(Text));
+  refs         @2 (hash :Text) -> (refs :List(Text));
   # Get the set of branches and PRs with this commit at their head.
 
-  status @3 () -> (status :BuildStatus);
+  status       @3 () -> (status :BuildStatus);
 
-  title @4 (hash :Text) -> (title :Text);
+  message      @4 (hash :Text) -> (title :Text);
 }
 
 interface Repo {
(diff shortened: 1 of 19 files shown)


benmandrew:
`String.starts_with` (and `ends_with`) are only implemented 4.13 onwards. What's our policy on backwards compatibility?

novemberkilo:
> What's our policy on backwards compatibility?

Well, we're pretty much in the driver's seat with this project since we're deploying it. It's in a different situation to the general case (e.g. writing libraries) ... in general I think we're pretty well motivated to maintain backward compatibility to the best extent possible. 

If you wanted to I think we could introduce a lower bound of `4.13` without any issue. WDYT @maiste and @tmcgilchrist?

tmcgilchrist:
In general I would keep wide support for OCaml versions in libraries. For executables or end applications having a more narrow OCaml version is fine. We are using 4.14 as the deploy target in Dockerfiles so a lower bound of 4.13 is not an issue.

novemberkilo (commented):
@benmandrew thanks much for this work :) 

It's pretty expansive in its scope ... and as you point out, involves changes to the API and thus the service components as well. I'm a little nervous about the surface area of the changes here. 

It's generally good practice to try and chunk PRs into small pieces - a general principle is to try and introduce a single new (conceptual) change (although this is often not practical). How do you feel about introducing a smaller PR that focuses on one page, and then adding subsequent PRs that introduce further pages etc.

Smaller PRs are much more amenable to a good quality review as well. 

Perhaps we could use this PR as a way of facilitating a discussion but then look to smaller PRs to actually introduce these changes in smaller increments. What do you think?

api/schema.capnp
```diff
@@ -60,38 +49,60 @@ struct JobInfo {
...
+  }
+  message @5 :Text;
+  # The state of the ref's head commit
 }
 
 interface Repo {
   refs         @0 () -> (refs :List(RefInfo));
   # Get the set of branches and PRs being monitored.
+  mainRef      @1 () -> (ref :RefInfo);
+  # Get the main branch, for UI display purposes
```
novemberkilo, line 80: I don't think this is correct (or at least in principle it isn't ideal). 

If I understand correctly (looking at https://capnproto.org/language.html), to maintain backward compatibility we should add new fields by appending them to the list of fields rather than inserting them.

maiste: This is a question we need to discuss. Do we want to maintain backward compatibility? Considering we maintain both ends of the pipe, is it a problem if we break the compatibility for the UI capnp? I'm not opinionated at all on this, WDYT?

benmandrew: Assuming we update `live-engine` and `live-www` at the same time, I don't believe it should be an issue.

novemberkilo: I was brought up in the school of distributed systems where making any assumptions about the clients of your API is fraught with danger :) 

You are both 💯 - We do have control over the client and the server here. I would encourage us to maintain backward compatibility on principle unless there is pain involved. In this case, does the position of the new field `mainRef` matter?

maiste: In this case it matters yes :)

maiste:
As this PR has been split into multiple PRs (#572 #562 #553 #590) I think we can consider closing it.

maiste:
Feel free to reopen.

PR #560, opened 2022-10-22
MacOS builds for pipelines

tmcgilchrist:
Adds macOS Monterey (x86_64 and ARM64) builds for ocaml-ci to all pipelines.

Currently we support 2 OCaml compilers 4.14 and 5.0 on both architectures. We could add more if there was sufficient interest and value in testing them.


service/conf.ml
```diff
@@ -63,14 +63,58 @@ type platform = {
...
+    {
+      label = "macos-homebrew";
+      builder = Builders.local;
+      pool = "macos-x86_64";
+      distro = "macos-homebrew";
+      ocaml_version = OV.Releases.v5_0;
+      arch = `X86_64;
+      opam_version = `V2_1;
+    };
+    (* Apple Silicon *)
```
tmcgilchrist, line 89: Even though we can support macOS arm64, we only have two machines available. @mtelvers

novemberkilo (approved):
👍🏾 @tmcgilchrist 

I don't know this area of the code well enough to add much to the review. I used it as an opportunity to learn ... thanks for that. ~~I'm curious to understand what the next steps are~~ (this PR makes it pretty clear that we are starting out with some hard-coded configuration, and with pretend MacOS docker containers).

Edit: I see the tracking ticket now. Sorry.

service/conf.ml
tmcgilchrist: Additional macOS arm64 machines will soon be added to the cluster.

tmcgilchrist: Additional macOS arm64 machines are available, I will enable both architectures.

PR #565, opened 2022-10-24
Refs page

benmandrew:


api/client.ml
```diff
@@ -85,6 +85,14 @@ end
 module Repo = struct
   type t = Raw.Client.Repo.t Capability.t
 
+  type ref_info = {
+    name : string;
```
maiste, line 89:
```
gref : string;
```

web-ui/view/build.ml
```diff
@@ -101,3 +107,73 @@ let tabulate_steps step_rows =
     div
       ~a:[ a_class [ "container-fluid mt-8 flex flex-col space-y-6" ] ]
       [ div ~a:[ a_class [ "table-container" ] ] step_rows ])
+
+let repo_row ~repo_title ~short_hash ~last_updated ~status ~repo_uri =
```
novemberkilo, line 111: I would suggest that this should be in a `repo.ml` module. I think that the `build` module should only contain view elements and helpers relevant to the Build page.

web-ui/view/build.ml
```diff
@@ -101,3 +107,73 @@ let tabulate_steps step_rows =
...
+                        div [ txt short_hash ];
+                        div [ txt "-" ];
+                        div [ txt last_updated ];
+                      ];
+                  ];
+              ];
+          ];
+      ])
+
+let ref_row ~ref_title ~short_hash ~last_updated ~status ~ref_uri ~message =
```
novemberkilo, line 141: As above - this should probably be in a `ref.ml` module.

web-ui/view/build.ml
```diff
@@ -54,6 +54,12 @@ let title_card ~status ~card_title ~hash_link ~ref_links ~first_created_at
           ];
       ])
 
+let tabulate rows =
+  Tyxml.Html.(
+    div
+      ~a:[ a_class [ "container-fluid mt-8 flex flex-col space-y-6" ] ]
+      [ div ~a:[ a_class [ "table-container" ] ] rows ])
```
novemberkilo, line 61: I don't think this is correct -- I do not see a `table-container` class in the designer's sample of the Refs page. Also, this function is not relevant to the Build page so I suggest it should not appear here ... if it is a common function to various views it could go into `common.ml`

web-ui/view/build.ml
```diff
@@ -101,3 +107,73 @@ let tabulate_steps step_rows =
...
+                  [
+                    div
+                      ~a:[ a_class [ "text-gray-900 text-sm font-medium" ] ]
+                      [ txt message ];
+                    div
+                      ~a:[ a_class [ "flex text-sm space-x-2" ] ]
+                      [
+                        div [ txt short_hash ];
+                        div [ txt "-" ];
+                        div [ txt last_updated ];
```
novemberkilo, line 174: I couldn't get this to render correctly on my local. This is what I got 
Screen Shot 2022-10-26 at 3 51 05 pm

I have attempted to "fix" these in this PR #571 and I suggest that this be merged so that `master` can be deployed.

web-ui/view/build.ml
novemberkilo: Also, I think that the added `table-container` class is throwing off the css. Please can we review the markup that the designer gave us - I think we are missing a div with classes 

```
border border-gray-200 rounded-lg w-full overflow-hidden shadow-sm  divide-y divide-gray-200
```

novemberkilo (commented):
Hey thanks @benmandrew - the refs page works correctly in that it links to the right builds etc. 

I've made some suggestions for fixes to the refs page that I think should have gone in prior to it appearing on master. I understand that the early merge is likely due to a miscommunication ... apologies for contributing to that :( 

Please also see #571 which undoes some of the changes here in an attempt to make it so that if master does get deployed, the refs page will be ok.

PR #568, opened 2022-10-25
API route and serialisation of build.

novemberkilo:
~~This is a WIP. I've got a test that suggests something wrong with the way the build's run_time is being calculated ...~~

This PR introduces an `api` route to fetch a json `build` object. 

In a consequent PR we will introduce some javascript for the build page that will use this route to implement live-updates of the build page.


test/test_build_representation.ml
```diff
@@ -0,0 +1,62 @@
...
+module Step = Representation.Step
+module Client = Ocaml_ci_api.Client
+module Run_time = Ocaml_ci_client_lib.Run_time
+
+let test_to_json (jobs, build_status, build_created_at, expected) =
+  let result =
+    Build.to_json
+    @@ Build.from_jobs_status ~jobs ~build_status ~build_created_at
+  in
+  Alcotest.(check string) "to_json" expected result
```
tmcgilchrist, line 11: Going via a `string` could be problematic in the future. 
Ideally this would have `Alcotest.(check Build.t)` and be able to show a nice diff of the data structure. However that requires some changes to alcotest https://github.com/mirage/alcotest/issues/194

tmcgilchrist (approved):

test/test_build_representation.ml
novemberkilo: C’est noté 👍🏾

PR #572, opened 2022-10-26
Refs page: the sequel

benmandrew:
Refs quality-of-life stuff.

Some data is not yet available to the UI. The title of the ref is derived from the gref; this works for branches (`refs/heads/[branch-name]`) but not for PRs (`refs/pull/[pull-id]/head`). It would be more robust to simply store this in the DB (along with commit messages, timestamps, etc).

Timestamps especially have some non-trivial UI logic associated with them according to the design spec, which is not yet implemented.


novemberkilo (commented):
Thanks @benmandrew - I can confirm that this renders well locally. 

I'm wondering if we can add the `started_at` and `ran_for` data in a manner that is similar to what we do for the build page.

At the moment, this would have to be done inefficiently - something like this:

```
for each branch
let sha = commit-sha of head of the branch
let jobs = capnp call to get list-of-jobs that correspond to sha
let started_at = Run_time.first_step_queued_at jobs 
let total_run_time = Run_time.total_of_run_times jobs  
```

This is what is done in for a single build in `list_steps` at the moment - see https://github.com/ocurrent/ocaml-ci/blob/master/web-ui/controller/git_forge.ml#L125-L132 

I wonder how many refs you need to have before this naive implementation really begins to struggle. Is it worth investigating this for a half day or so ... we might be able to add the timestamps and durations now and then replace them with a more sensible implementation when we have additional support from the persistence layer. WDYT?

tmcgilchrist:
@benmandrew could the UI display the PR title along with the number on this page? https://ci.ocamllabs.io/github/ocurrent/ocaml-ci

api/client.ml
```diff
@@ -213,4 +215,10 @@ module Commit = struct
          | Pending -> Ok `Pending
          | Undefined i ->
              Error (`Msg (Fmt.str "client.states: undefined state %d" i)))
+
+  let message t =
+    let open Raw.Client.Commit.Message in
+    let request = Capability.Request.create_no_args () in
+    Capability.call_for_value t method_id request
+    >>= (Results.message_get >> fun x -> Ok x)
```
maiste, line 223:
```suggestion
    >>= (Results.message_get >> Result.ok)
```
Only a suggestion; feel free to ignore it.

lib/index.mli
```diff
@@ -93,11 +93,12 @@ val get_active_repos : owner:string -> Repo_set.t
...
 
-val set_active_refs : repo:Repo_id.t -> string Ref_map.t -> unit
+val set_active_refs :
+  repo:Repo_id.t -> (string * string * string) Ref_map.t -> unit
 (** [set_active_refs ~repo refs] records that [refs] is the current set of Git
     references that the CI is watching. There is one entry for each branch and
     PR. Each entry maps the Git reference name to the head commit's hash. *)
 
-val get_active_refs : Repo_id.t -> string Ref_map.t
+val get_active_refs : Repo_id.t -> (string * string * string) Ref_map.t
```
maiste, line 102: Could you update the comment to clarify each field of the tuple?

service/api_impl.ml
```diff
@@ -94,6 +94,40 @@ let make_commit ~engine ~owner ~name hash =
...
+         let active =
+           Index.get_active_refs { Ocaml_ci.Repo_id.owner; name }
+           |> Index.Ref_map.bindings
+           |> List.filter_map (fun (_, (h, message, _)) ->
+                  if h = hash then Some message else None)
+         in
+         (match active with
+         | [] ->
+             ()
+             (* [benmandrew]: Theoretically this case should never happen O_O *)
```
maiste, line 111: If it's not supposed to happen, shouldn't we use `List.hd`? @novemberkilo  what do you thing?

api/client.ml
benmandrew: Good suggestion, didn't know that existed:)

service/api_impl.ml
novemberkilo: Yes I think that's a good suggestion - At the moment we are swallowing the error. 

Either we should handle it by throwing an exception via `List.hd` or we use this branching approach to log an error and then manually throw an exception.

maiste: Logging the error seems to be reasonable to me :+1:

maiste (approved):
The new version looks good to me! :+1: Everything works fine! Thanks for this PR :)
It can be merged at any time as the submodule has been updated.

PR #575, opened 2022-10-31
Point to fork of dream for OCaml 5.

novemberkilo:
Points to my fork of dream that builds on OCaml 5. Once https://github.com/aantron/dream/pull/241 is merged, we can revert this (modulo the restriction on OCaml 5).

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -121,7 +121,7 @@
   (ansi (>= 0.5.0))
   (capnp-rpc-unix (>= 1.2))
   (crunch (and (>= 3.2.0) :build))
-  (dream (= 1.0.0~alpha4))
+  dream
   (fmt (>= 0.8.9))
   (logs (>= 0.7.0))
   (lwt (>= 5.6.1))
@@ -131,7 +131,6 @@
   (prometheus-app (>= 1.2))
   tyxml
   (timedesc (>= 0.9.0)))
-  (conflicts (ocaml (>= 5.0)))
 )
 (package
 (name ocaml-ci-client-lib)
--- a/ocaml-ci-web.opam
+++ b/ocaml-ci-web.opam
@@ -10,7 +10,7 @@ depends: [
   "ansi" {>= "0.5.0"}
   "capnp-rpc-unix" {>= "1.2"}
   "crunch" {>= "3.2.0" & build}
-  "dream" {= "1.0.0~alpha4"}
+  "dream"
   "fmt" {>= "0.8.9"}
   "logs" {>= "0.7.0"}
   "lwt" {>= "5.6.1"}
@@ -21,9 +21,6 @@ depends: [
   "tyxml"
   "timedesc" {>= "0.9.0"}
   "odoc" {with-doc}
-]
-conflicts: [
-  "ocaml" {>= "5.0"}
 ]
 build: [
   ["dune" "subst"] {dev}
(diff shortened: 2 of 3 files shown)


dune-project
```diff
@@ -8,7 +8,7 @@
  (name ocaml-ci-api)
  (synopsis "Cap'n Proto API for ocaml-ci")
  (depends
-  (ocaml (>= 4.13))
```
tmcgilchrist, line 11: These changes are unnecessary. Using 4.13 should work fine.

ocaml-ci-web.opam
```diff
@@ -40,3 +37,10 @@ build: [
   ]
 ]
 dev-repo: "git+https://github.com/ocurrent/ocaml-ci.git"
+
+pin-depends: [
+  ["dream.dev" "git+https://github.com/novemberkilo/dream#4976439384c9da03300abdbc9ffb8528ac917982"]
+  ["dream-httpaf.dev" "git+https://github.com/novemberkilo/dream#4976439384c9da03300abdbc9ffb8528ac917982"]
+  ["dream-mirage.dev" "git+https://github.com/novemberkilo/dream#4976439384c9da03300abdbc9ffb8528ac917982"]
```
tmcgilchrist, line 44: Do you need `dream-mirage.dev` here? I wouldn't expect we install that in the opam switch.

ocaml-ci-web.opam
```diff
@@ -10,21 +10,18 @@ depends: [
   "ansi" {>= "0.5.0"}
   "capnp-rpc-unix" {>= "1.2"}
   "crunch" {>= "3.2.0" & build}
-  "dream" {= "1.0.0~alpha4"}
+  "dream"
```
tmcgilchrist, line 13: Make this `"dream" {>= "dev"}`

dune-project
```diff
@@ -121,7 +121,7 @@
   (ansi (>= 0.5.0))
   (capnp-rpc-unix (>= 1.2))
   (crunch (and (>= 3.2.0) :build))
-  (dream (= 1.0.0~alpha4))
+  dream
```
tmcgilchrist, line 124: This should be `(dream (>= :dev))` otherwise the opam file will get overwritten

tmcgilchrist (approved):

maiste:
Considering that the true `ocaml-h2` (not aanton fork) seems to have a lot of activity here to make it compatible with `ocaml.5.0.0`. Shouldn't the PR on `dream` point to this version: https://github.com/anmonteiro/ocaml-h2

novemberkilo:
> Considering that we need some tweag of dune for various ocurrent projects? Shouldn't we consider having a fork in the ocurrent organisation? We could maintain one fork over multiple ones. WDYT?

Yes definitely - this is something that @tmcgilchrist and I have also talked about. We're starting here but if it makes sense, we should maintain a fork in the `tarides` organisation.

novemberkilo:
> Considering that the true ocaml-h2 (not aanton fork) seems to have a lot of activity here to make it compatible with ocaml.5.0.0. Shouldn't the PR on dream point to this version: https://github.com/anmonteiro/ocaml-h2

This is where I started as well. It turns out that `aanton` has maintained a fork of `ocaml-h2` that has diverged from `anmonteiro`'s so I have ported over the work done on the original project to make it compatible with OCaml 5 and added that as a single commit.

maiste:
Does the fork diverge too much in comparison to the original version? In terms of maintainability, it would be easier to have the `ocaml-h2` included in `dream` instead of having multiple forks no?

PR #582, opened 2022-11-07
Fixed bug preventing GitLab routes from being added

benmandrew:
I don't have a GitLab account so I haven't been able to test this fully myself, but it should work.

The bug is that GitLab routes are only added if there are no GitHub routes, almost certainly not the intended behaviour.


maiste:
LGTM :+1: 
Already deploy on `live-www` and work perfectly. Thanks!

maiste (approved):

novemberkilo (commented):
Thanks for covering this @benmandrew 🚀

Issue #580, opened 2022-11-07
Transient server error for new branches and PRs? 


novemberkilo:
When a new branch or PR is created and it appears on ocaml-ci, the server can return a 500 when that branch is clicked on in the UI. This requires investigation - is this a transient issue (has this been the case with the previous UI as well)? In any case, the UX needs to be improved (more specs to follow).

maiste:
Fixes by #583.

PR #583, opened 2022-11-07
Fix transient server error

benmandrew:
Possible fix for issue #580.

In computing the timestamp for `first_queued_at`, the case where all jobs have a `None` timestamp is not accounted for, which results in the default-case `max_float` being returned, which is too large for Timedesc to cast to a 64-bit integer, tripping an assert.

The transient error then happens at the beginning of the build when only the analysis step exists, and there is a period of time where it has no `queued_at` timestamp.

I'm having issues running ocaml-ci locally so haven't been able to test fully.

### DIFF
--- a/lib_client/run_time.ml
+++ b/lib_client/run_time.ml
@@ -191,7 +191,11 @@ let total_of_run_times (build : Client.job_info list) =
   |> List.fold_left (fun subtotal rt -> subtotal +. total_time rt) 0.
 
 let first_step_queued_at (jil : Client.job_info list) =
-  if jil = [] then Error "Empty build"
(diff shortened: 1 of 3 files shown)


novemberkilo (commented):
Thanks @benmandrew this seems pretty compelling. Could you please add a test covering this case (of a non-empty job list but no jobs having a queued_at timestamp). See [here](https://github.com/ocurrent/ocaml-ci/blob/master/test/test_run_time_client.ml#L403) to get started.

novemberkilo (approved):
Thanks @benmandrew 🚀

maiste (approved):
Thanks @benmandrew! :+1:

PR #584, opened 2022-11-08
Spring clean web-ui (refactor and prune)

novemberkilo:
This was motivated by wanting to make the api controllers and views consistent with the way the `Git_forge` machinery implements non-api controllers and views. This lead to general cleaning out of unnecessary dependencies etc.


benmandrew (approved):
Looks good to me, I don't think it'll clash with the web views and gives my some ideas as to how I can structure them!

web-ui/controller/api/dune
```diff
@@ -0,0 +1,16 @@
+(library
+ (public_name ocaml-ci-web.controller.api)
+ (name api_controller)
+ (libraries
+  lwt
+  dream
+  unix
+  yojson
+  ppx_deriving_yojson.runtime
+  prometheus-app
```
tmcgilchrist, line 10:
```suggestion
```
This doesn't seem necessary?

web-ui/controller/api/git_forge.ml
```diff
@@ -0,0 +1,82 @@
+module type Api = Api_view.Git_forge.Api
+
+module Client = Ocaml_ci_api.Client
```
tmcgilchrist, line 3:
```suggestion
```
This can be removed as you bind the name inside `module Make` later on. You could even move the `module Run_time` into that module.

web-ui/controller/api/git_forge.ml
```diff
@@ -0,0 +1,82 @@
...
+  val show_step :
+    org:string ->
+    repo:string ->
+    hash:string ->
+    variant:string ->
+    Controller.Backend.t ->
+    Dream.response Lwt.t
+end
+
+let ( >>!= ) = Controller.Git_forge.( >>!= )
```
tmcgilchrist, line 23: This could be scoped to the `module Make`

web-ui/view/api/github.ml
```diff
@@ -1,21 +1,5 @@
...
-    Run_time.build_created_at ~build:jobs
-    |> Result.to_option
-    |> Option.join
-    |> Option.value ~default:0.
-  in
-  Dream.json
-  @@ Build.to_json
-  @@ Build.from_jobs_status ~jobs ~build_status ~build_created_at
-       ~step_route_prefix
+include Git_forge.Make (M_Github)
```
tmcgilchrist, line 5:
```suggestion
include Git_forge.Make (
            struct
              let prefix = "github"
            end)
```

This is also valid if you don't want to give a proper name to `M_Github`

web-ui/view/api/git_forge.ml
```diff
@@ -0,0 +1,50 @@
...
+    repo:string ->
+    hash:string ->
+    jobs:Client.job_info list ->
+    build_status:Client.State.t ->
+    Dream.response Lwt.t
+end
+
+module type M_Git_forge = sig
+  val prefix : string
+end
```
novemberkilo, line 22: Following an IRL chat this module is very thin at the moment but has the potential to change when we properly implement the GitLab views. If the module does not change and is just a wrapper on a string, maybe we back away from this approach and just used shared code between the different git-forges with the prefix injected into this shared code.

PR #596, opened 2022-11-09
Add the getting-started page.

### LINKED ISSUE #594: Documentation - Getting Started
https://github.com/apps/ocaml-ci is a starting point

This will change as we change the sign-up process.

novemberkilo:
Starting with the copy from https://github.com/apps/ocaml-ci

Closes #594


web-ui/view/documentation.ml
```diff
@@ -0,0 +1,90 @@
...
+              h1 [ txt "Getting Started" ];
+              p [ txt "OCaml-CI is a CI service for OCaml projects." ];
+              hr ();
+              p
+                [
+                  txt
+                    "It uses metadata from the project's opam and dune files \
+                     to work out what to build, and uses caching to make \
+                     builds fast. It automatically tests projects against \
+                     multiple OCaml versions and OS platforms.";
```
maiste, line 28: Can't we add here, that these versions are the ones allowed by the opam file?

web-ui/view/documentation.ml
```diff
@@ -0,0 +1,90 @@
...
+                      a
+                        ~a:[ a_href "https://github.com/apps/ocaml-ci" ]
+                        [ txt "https://github.com/apps/ocaml-ci" ];
+                    ];
+                  li
+                    [
+                      txt
+                        "Select only the repositories you want tested - if you \
+                         select ";
+                      code [ txt "All Repositories" ];
```
maiste, line 47: Can't we use the strong markup here instead of code?

web-ui/view/documentation.ml
```diff
@@ -0,0 +1,90 @@
...
+                  li
+                    [
+                      txt "Github documentation on how to install ";
+                      a
+                        ~a:
+                          [
+                            a_href
+                              "https://docs.github.com/en/developers/apps/managing-github-apps/installing-github-apps#installing-your-private-github-app-on-your-repository";
+                          ]
+                        [ txt "an app on your repository" ];
```
maiste, line 78: Instead, shouldn't we point to the dev.md on ocurrent.org? or the one on GitHub?

maiste:
Thanks for this PR! Few things raised just to be sure :+1:

web-ui/view/documentation.ml
novemberkilo: I don't understand this suggestion - could you please provide a link?

web-ui/view/documentation.ml
novemberkilo: We sure can. As mentioned, this PR starts by bringing over the words that are on https://github.com/apps/ocaml-ci -- I expect that it will change over time, especially as we introduce a different sign up process.

I will add some more words as suggested.

maiste: Sure! Thanks for the clarification :smile:

web-ui/view/documentation.ml
maiste: I was thinking about this [page](https://github.com/ocurrent/ocaml-ci/blob/master/doc/dev.md) and in a near future, it's [ocurrent.org version](https://www.ocurrent.org/docs/ocurrent/ocurrent/)

maiste (approved):
Appart from the `dune fmt` required by the CI, LGTM :+1:

novemberkilo:
@maiste @benmandrew after talking to @tmcgilchrist about the signup process, I've decided to start by making this getting-started more detailed. To do this I have added some images. Eventually we will build a process that does not involve the user adding anything via a PR to ocaml-ci and then we can edit this to reflect that.

web-ui/view/documentation.ml
novemberkilo: @maiste I don't think this page is relevant to the users of ocaml-ci - rather the developers of ocaml-ci. Perhaps I am missing something in your suggestion?

maiste: The part I misunderstood is the fact that the document explained how to install the GitHub App, not create your own. All good about it. Closing :)

tmcgilchrist (approved):
This is a good start.

PR #597, opened 2022-11-10
Flash messages for cancel/rebuild - better UX.

### LINKED ISSUE #579: Improve flash messages UX
Currently when we rebuild or cancel a build we get individual flash messages for each step being rebuilt or cancelled. Each flash message has to be closed by clicking on the close button. This is awkward when there are many such steps.

Improved UX:

- When we know we are rebuilding or cancelling a subset of the steps of a build, show a single flash message containing all the information we need to relay
- When we are rebuilding the entire build, simply say, "Rebuilding all."

There is also a suggestion that the flash message be ephemeral. That it disappear automatically after n seconds (where n is maybe ~ 3?)

novemberkilo:
We are issuing a flash message for every job that is cancelled or rebuilt. This can result in a lot of flash messages.

This PR collects the labels of all the jobs being rebuilt/cancelled and adds the first five to a single flash message. Depending on the number of jobs being rebuild/cancelled, the flash message is:

```
Rebuilding: (lint-doc), (lint-fmt), foo, bar

or

Rebuilding many: (lint-doc), (lint-fmt), foo, bar, baz, ...
``` 

Closes #579

### DIFF
--- a/web-ui/view/github.ml
+++ b/web-ui/view/github.ml
@@ -132,13 +132,26 @@ let cancel_success_message success =
   | [] -> div [ span [ txt "No jobs were cancelled." ] ]
   | success -> ul (List.map format_job_info success)
 
-let cancel_success_message_v1 success =
-  let format_job_info ji =
-    (`Success, Printf.sprintf "Cancelling job: %s" ji.Client.variant)
+let success_message_v1 action jis =
+  let empty_message, prefix =
+    match action with
+    | `Cancel -> ("No jobs were cancelled.", "Cancelling")
+    | `Rebuild -> ("No jobs were rebuilt.", "Rebuilding")
   in
-  match success with
-  | [] -> [ (`Success, "No jobs were cancelled.") ]
-  | success -> List.map format_job_info success
+  match jis with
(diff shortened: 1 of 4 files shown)


web-ui/view/github.ml
```diff
@@ -228,13 +228,26 @@ let cancel_success_message success =
...
   in
-  match success with
-  | [] -> [ (`Success, Fmt.str "No jobs were cancelled.") ]
-  | success -> List.map format_job_info success
+  match jis with
+  | [] -> [ (`Success, empty_message) ]
+  | jis ->
+      let trimmed = List.filteri (fun i _ -> i < 5) jis in
+      let message =
+        Astring.String.concat ~sep:"\x0C, "
```
benmandrew, line 242: Any reason for using \x0C in the separator? Seems like it's just a control character for printers (form feed).

benmandrew (commented):
Apart from the above, looks good to me and works correctly on my machine.

web-ui/view/github.ml
novemberkilo: Oh hmm cruft left over from trying to format the text. Thanks for catching.

PR #603, opened 2022-11-15
Fix formatting of refs page -- Truncate ref message. Remove started_at.

novemberkilo:
Addressing some of the formatting issues here by truncating messages to 72 characters and removing `started_at` related strings for now. 

Screen Shot 2022-11-15 at 2 41 45 pm

### DIFF
--- a/web-ui/view/ref.ml
+++ b/web-ui/view/ref.ml
@@ -12,6 +12,9 @@ module Make (M : M_Git_forge) = struct
     | _ -> raise Not_found
 
   let row ~ref ~short_hash ~started_at ~ran_for ~status ~ref_uri ~message =
+    ignore started_at; (*See FIXME - revert this when started_at is implemented *)
+    (* messages are of arbitrary length - let's truncate them *)
+    let message = Astring.String.with_range ~len:72 message in
     let ref_title =
       match ref with Branch title -> title | PR { title; _ } -> title
     in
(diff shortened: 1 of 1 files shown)


web-ui/view/ref.ml
```diff
@@ -93,7 +99,7 @@ module Make (M : M_Git_forge) = struct
       let started_at = Timestamps_durations.pp_timestamp started_at in
       let ran_for = Timestamps_durations.pp_timestamp ran_for in
       row ~ref:(ref gref name) ~short_hash ~started_at ~ran_for ~status
-        ~ref_uri:(commit_url ~org ~repo short_hash)
+        ~ref_uri:(commit_url ~org ~repo hash)
```
novemberkilo, line 102: This breaks our convention of using long hashes in URLs. This fixes it and a subsequent PR seeks to harden this against future drift back to short hashes.

novemberkilo:
I'm merging this so I can pick it up with the deploy I'm going to do.

web-ui/view/ref.ml
novemberkilo: I think that prior to introducing the new refs page, the old page used the full hash https://github.com/ocurrent/ocaml-ci/pull/565/files#diff-15d79b23f61af65eb663efd069be96a638f6c4ac6073879ddfd42a5bdd4f0437L50

PR #604, opened 2022-11-15
Refactor and harden url helpers.

novemberkilo:
We've had an issue where we've drifted into using short hashes in urls. So far we've been using long hashes and in an effort to maintain consistency, this PR collects all url helpers in one place, and asserts that hashes are not short.

This PR builds on #603

### DIFF
--- a/web-ui/view/github.ml
+++ b/web-ui/view/github.ml
@@ -11,15 +11,6 @@ module Ref = Ref.Make (struct
   let prefix = prefix
 end)
 
-let org_url org = Printf.sprintf "/%s/%s" prefix org
-let repo_url org repo = Printf.sprintf "/%s/%s/%s" prefix org repo
-
-let commit_url ~org ~repo hash =
-  Printf.sprintf "/%s/%s/%s/commit/%s" prefix org repo hash
-
-let job_url ~org ~repo ~hash variant =
-  Printf.sprintf "/%s/%s/%s/commit/%s/variant/%s" prefix org repo hash variant
-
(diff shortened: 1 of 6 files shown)


web-ui/view/github.ml
```diff
@@ -54,7 +46,11 @@ let history_v ~org ~repo ~history =
...
          ~a:[ a_class [ Build_status.class_name status ] ]
-         [ a ~a:[ a_href (commit_url ~org ~repo commit) ] [ txt commit ] ])
+         [
+           a
+             ~a:[ a_href (Url.commit_url prefix ~org ~repo ~hash:commit) ]
+             [ txt commit ];
+         ])
 
 let link_github_commit ~org ~repo ~hash =
   a ~a:[ a_href (github_commit_url ~org ~repo ~hash) ] [ txt hash ]
```
benmandrew, line 56:
```suggestion
  a ~a:[ a_href (github_commit_url ~org ~repo ~hash) ] [ txt (short_hash hash) ]
```

benmandrew: There are a couple of places on the build page and steps page where we still use short hashes. I have line numbers below as I can't seem to make code suggestions outside of already edited blocks:

- Build page, github.ml line 245
- Steps page, github.ml line 317

Change `~hash:(short_hash hash)` argument to `~hash` in both cases.

benmandrew (changes requested):
Thanks for catching this, LGTM apart from a couple of missed instances.

benmandrew (approved):

PR #605, opened 2022-11-15
Machinery to generate tailwind css from the views

novemberkilo:
Following the patterns in `ocaml/ocaml.org` and with some excellent and much appreciated help from @tmattio

This should allow us to introduce our own css easily - in particular, it will allow us to introduce text variants to support responsive classes. See the example introduced here.


benmandrew (approved):
LGTM. Installing was a little convoluted, I was asked to create a new opam package called tailwindcss, is this normal?

Issue #593, opened 2022-11-09
Extend the information model of ocaml-ci


novemberkilo:
This is needed to improve upon the current approach of cobbling together information from the `ci_build_index` and `current_cache` tables and other in-memory maps.

It is a pre-requisite to support history of builds and the metrics that we want to display for the default branch of each repo. #588 #591

novemberkilo:
Associated https://github.com/ocurrent/ocaml-ci/issues/532

novemberkilo:
Duplicate ...

PR #606, opened 2022-11-15
Ref and repo data aggregation

benmandrew:
Initial work on data aggregation. Currently working on aggregating ref-level data (derived from the head commit of the ref) and repo-level data (derived from its default branch).

To support history, a second data structure is required - already partially implemented by the `Status_cache` - which needs to provide commit-level granularity. The aggregation of ref and repo data will probably rely on this cache for the head commits.

### DIFF
--- a/lib/index.mli
+++ b/lib/index.mli
@@ -65,7 +65,7 @@ val get_build_history_with_time :
     branch gref of the repo identfied by (owner, name). The builds are
     identified by triples (variant, hash, Some timestamp) *)
 
-val get_status : owner:string -> name:string -> hash:string -> build_status
+(* val get_status : owner:string -> name:string -> hash:string -> build_status *)
 (** [get_status ~owner ~name ~hash] is the latest status for this combination. *)
 
 val get_full_hash :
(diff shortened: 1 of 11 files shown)


lib/index.mli
```diff
@@ -111,3 +111,23 @@ val get_active_refs : Repo_id.t -> ref_info Ref_map.t
...
+  val get_ref_started_at : ref_state -> float
+  val get_ref_ran_for : ref_state -> float
+  val get_repo_status : repo_state -> status
+  val get_repo_started_at : repo_state -> float
+
+  val set_ref_state :
+    repo:Repo_id.t -> ref:string -> status -> float -> float -> unit
+
+  val get_ref_state : repo:Repo_id.t -> ref:string -> ref_state
+  val aggregate_repo : repo:Repo_id.t -> unit
```
novemberkilo, line 132: Seeing that this function updates the repo's state in the `Repo_map` what do you think of naming this as `update` or `set` or similar?

benmandrew:
Partial solution to #593; work on the DB schema is still necessary.

novemberkilo (approved):
Thanks @benmandrew - I've skimmed the PR and things look good. I haven't had time to test it locally or look through it in great detail but I don't think this should hold you up from merging it when you are ready to.  I will revisit it and look through things in more detail in the next day or so.

benmandrew:
This PR has been deployed.
(Clarification due to potentially confusing commit history.)

PR #610, opened 2022-11-17
Better build run-times.

novemberkilo:
The run-time for a build is described as the total of the run-times of all the steps of the build. While this is useful, it does not tell a very good story of how long a build takes to run and this is mainly because after the analysis step, the rest of the steps run in parallel.

This PR introduces a `build-run-time` which is:

```
(run-time of the analysis step) + max ((run-time of step) | step in build)
``` 

This `build_run_time` is displayed prominently (in the same spot that the run-time for the build has been displayed). The total-run-time continues to be displayed, but now less prominently.

Screen Shot 2022-11-17 at 9 40 39 pm

### DIFF
--- a/lib_client/run_time.mli
+++ b/lib_client/run_time.mli
@@ -78,6 +78,17 @@ val build_created_at :
 val total_time : run_time_info -> float
 (** The total time taken -- the sum of the queued and running times *)
 
+val max_of_step_run_times :
+  build_created_at:float -> Ocaml_ci_api.Client.job_info list -> float
+(** Takes a list of steps and returns the maximum of the run-times of the steps.
+    This is done by calculating the run-times of the steps in the list, sorting
+    the resulting list of run-times, and returning the head of the list
+    (defaulting to 0) *)
+
 val total_of_run_times : Ocaml_ci_api.Client.job_info list -> float
 (** Takes a list of job_info which is meant to be the steps in a build. Returns
     total of run times of the steps in the build.*)
+
+val build_run_time : Ocaml_ci_api.Client.job_info list -> float
+(** Takes a list of job_info which is meant to be the steps in a build. Returns
+    the (run-time of the analysis step) + max(run-times of the steps in the build) *)
--- a/lib_client/run_time.mli
+++ b/lib_client/run_time.mli
@@ -91,4 +91,5 @@ val total_of_run_times : Ocaml_ci_api.Client.job_info list -> float
 
 val build_run_time : Ocaml_ci_api.Client.job_info list -> float
 (** Takes a list of job_info which is meant to be the steps in a build. Returns
-    the (run-time of the analysis step) + max(run-times of the steps in the build) *)
+    the (run-time of the analysis step) + max(run-times of the steps in the
+    build) *)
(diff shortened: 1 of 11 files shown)


maiste:
Thanks for this PR! What about `Cumulative build time` instead of `Total`? Maybe, it comes from my poor English vocabulary but with the too time on the screen, the `total` feel a bit confusing. WDYT?

benmandrew (commented):
The changes you've made look good.

I agree with @maiste, 'total' is ambiguous. If we do change to 'cumulative' then all instances of `total_run_time` and similar throughout the code should be changed as well.

novemberkilo:
Thanks @maiste and @benmandrew for the feedback. I've thought about this and looked at examples from other CIs (GitHub Actions and also Buildkite) and both of them use the word 'total.' I'm not convinced that using 'cumulative' is better than using 'total' ... I prefer the simpler term because of its common use. Unless there are objections, I'd like to continue with the terminology as it stands. If you think I am missing something, or would like to discuss this further, please feel free to do that here.

maiste:
I understand your point of view, and I would keep **total** if we don't have any other time. If you take the case of GitHub Actions, they only display the total time. As it's monetized, you know how much it will cost. 

The more I think about it, the more I wonder if we shouldn't invert time values instead, keep the Total on the right side, and put the max time in the middle.  Indeed, the max time reflects the maximum time you have run, while the total time reflects how much we have occupied CPUs, which is a value easier to visualise...

MisterDA:
I find the current wording somewhat confusing.
Unless I'm mistaken, GHA doesn't show the cumulated time of all parallel build, but shows your definition of _build-run-time_ that it calls "_Total duration_".
Wikipedia speaks of _total CPU time_ which is for parallel programs the sum of the durations of each independent run (your _total-run-time_).
The less I could say is that the notion of _total_ is not well defined ;)

As a user I think I'm more interested by the _build-run-time_ than the _total-run-time_. If we're to display both numbers, I'd favor the wording _cumulative_ (or even _parallel_ or something else) instead of _total_.
I don't have a strong opinion on which should be placed where.

tmcgilchrist (commented):
This new reporting of the build-run-time as the sum of analysis plus the maximum runtime of a step is an improvement over the currently reported build time. It shows how long between initiating a build until you know the results of the build, which is the more important metric for the build page. The simplification that is being made here that Analysis runs by itself and then forks into many parallel builds is not obvious to a general user of ocaml-ci. We should add a section on the Documentation page explaining this in simplest terms.

The total build run time is a slightly less important statistic about how much CPU time was used for the build. Having it in a less prominent place with other details like branch and sha is the right option IMHO. Regarding Total vs Cumulative, I would favour the simplier wording along with having an explanation of what it means. That explanation could be a tooltip (?) and / or a link to the Documentation pages which would explain with examples how that calculation works.

lib_client/run_time.mli
```diff
@@ -78,6 +78,15 @@ val build_created_at :
 val total_time : run_time_info -> float
 (** The total time taken -- the sum of the queued and running times *)
 
+val max_of_step_run_times :
+  build_created_at:float -> Ocaml_ci_api.Client.job_info list -> float
+(** Takes a list of steps and returns the maximum of the run-times of the steps *)
```
tmcgilchrist, line 83: Add a brief description of how it does this calculation?

tmcgilchrist (approved):

maiste (approved):
Your call :)

novemberkilo:
Thanks everyone for your input. I hope that once we have documentation for this page, and a tooltip, the different run-times will be clearer.

PR #609, opened 2022-11-16
Add @maiste/ocaml-ci-experiment as GitLab user

maiste:
This is a toy project to test ocaml-ci over GitLab. The goal is to have an infrastructure to work on #586.

Issue #586, opened 2022-11-08
Gitlab views to the new template


novemberkilo:

maiste:
This is completed :partying_face:

PR #611, opened 2022-11-17
Step page for GitLab

maiste:
This PR is part of #586. This is a duplication of what has been made for the GitHub view. At some point, it will require refactoring because many areas of the code are shared between GitHub and GitLab. As we want to move quickly, this can be done next quarter.

This PR needs #609 and #604 to be merged before going further.

### DIFF
--- a/web-ui/view/gitlab.ml
+++ b/web-ui/view/gitlab.ml
@@ -310,65 +310,6 @@ let list_steps ~org ~repo ~message ~refs ~hash ~jobs ~first_step_queued_at
       div buttons;
     ]
 
-let _show_step ~org ~repo ~refs ~hash ~jobs ~variant ~job ~status ~csrf_token
-    ~timestamps ~build_created_at ?(flash_messages = []) (data, next) =
-  let header, footer =
-    let can_rebuild = status.Current_rpc.Job.can_rebuild in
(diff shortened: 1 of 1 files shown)


novemberkilo (approved):
@maiste thank you - it's great to see progress on the Gitlab views.

 I'm wondering if it makes sense though to start the Gitlab work by reducing duplication straight away. We'd add a `Make` functor to the `Git_forge` in `web-ui/view`  following the same pattern as here: https://github.com/ocurrent/ocaml-ci/blob/master/web-ui/view/api/git_forge.ml#L30 

If you think this will slow things down and you would rather cut across several GitLab views before removing the duplication, please ignore my suggestion. I think though, that we should at least commit to doing the refactor and removal of duplication this quarter. I don't like the prospect of visiting this separately after a break in the next year :)

What do you think?

maiste:
I agree to refactor this quarter. About the web to do it, it seems better to do the refactoring when we have the duplicated code than before, IMHO. Indeed, it will give us a better overview of what can be shared and what can not.  I will add and merge the duplicated views if that's good for you. Once it's done, I'll factorize everything. Is that ok for you?

novemberkilo:
Your call - ok by me :)

PR #620, opened 2022-11-21
Link to the git forge within Ref Branches list

MisterDA:
I think it's nice to have a link to the Git forge here considering we're showing the forge log and the PR/MR id as well.

### DIFF
--- a/web-ui/view/git_forge.ml
+++ b/web-ui/view/git_forge.ml
@@ -1,5 +1,6 @@
 module type M_Git_forge = sig
   val prefix : string
+  val request_url : org:string -> repo:string -> string -> string
 end
 
 module type View = sig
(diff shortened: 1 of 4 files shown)


maiste:
This is something that will require more important refactoring. There is a lot of duplication in the code for both forge, but the plan is to do the refactoring once both versions run (GitHub and Gitlab) well in the copy system.

MisterDA:
Well if we're going to refactor everything where's the harm in merging this?

maiste:
Two reasons here why I'm reluctant to merge the PR: it is interfering with conforming the `gitlab` views to the new design, and this work will be erased when we merge `gitlab` and `github` views.

MisterDA:
Ok, I trust you on this :)

maiste:
Thanks! :pray: Here is the issue to track the progression of the refactoring task https://github.com/ocurrent/ocaml-ci/issues/645

PR #621, opened 2022-11-21
Fix migrations system.

maiste:
This PR intends to fix an issue in GitLab. The migrations for GitLab were never deployed, as the flag `no_migration` was not negated here. It didn't cause any issue and shouldn't when deployed, as the table in the migration directory already exists.

Also, it fixes the issue where the migrations couldn't find the `migrations` repository as it was not mounted in the docker file.


tmcgilchrist (approved):
Same as GitHub pipeline 👍🏻

maiste:
@tmcgilchrist still good for you with the Dockerfile changes?

tmcgilchrist (approved):

PR #623, opened 2022-11-22
New index page -- all organisations.

novemberkilo:
Screen Shot 2022-11-22 at 2 54 57 pm

Uses a placeholder gitlab image for profile pictures of gitlab orgs

Updated with a selector so that the user can filter within GitHub or GitLab. Also, start with focus on the search field so that the user can immediately type to filter on their organisation (something stolen from Buildkite's UX).

Screen Shot 2022-11-23 at 4 00 39 pm


web-ui/controller/index.ml
```diff
@@ -1 +1,15 @@
...
+
+let ( >>!= ) = Git_forge.( >>!= )
 let render = View.Index.render
+
+let list_orgs prefix ci =
+  let open Lwt.Infix in
+  Backend.ci ci >>= Client.CI.orgs >>!= fun orgs ->
+  Dream.respond @@ View.Index.list_orgs' prefix orgs
+
+let list_orgs' ~github ~gitlab =
```
tmcgilchrist, line 11: Maybe rename these without using prime. `list_orgs` and `list_all_orgs` ?

web-ui/view/organisation.ml
```diff
@@ -0,0 +1,52 @@
+open Git_forge
+
+module Make (M : M_Git_forge) = struct
+  let profile_picture_url org =
+    (* FIXME [benmandrew]: How can we get the GitLab profile pictures? *)
```
tmcgilchrist, line 5: Gitlab serves avatars from gravatar, you can query them via the REST API
eg
```
curl -X GET https://gitlab.com/api/v4/users\?username\=tmcgilchrist | jq   
[
  {
    "id": 490393,
    "username": "tmcgilchrist",
    "name": "Tim McGilchrist",
    "state": "active",
    "avatar_url": "https://secure.gravatar.com/avatar/67afd2b4c98c9befd18c19f0ee9d94dc?s=80&d=identicon",
    "web_url": "https://gitlab.com/tmcgilchrist"
  }
]
```

tmcgilchrist (approved):

web-ui/view/organisation.ml
novemberkilo: Ah cool thanks - paging @benmandrew :)

web-ui/view/organisation.ml
```diff
@@ -0,0 +1,52 @@
...
+    match M.prefix with
+    | "github" -> Printf.sprintf "https://github.com/%s.png?size=88" org
```
benmandrew, line 7: Requesting a png of 88px results in a highly compressed-looking image. We can request a larger image (I chose 200px before) and the rendered size can be set in the HTML, it looks far better.

web-ui/view/index.ml
```diff
@@ -32,3 +32,89 @@ let render github gitlab =
...
+          div
+            ~a:[ a_id "table"; a_class [ "flex flex-col space-y-6" ] ]
+            (rows prefix orgs);
+        ];
+    ]
+
+let list_all_orgs ~github_orgs ~gitlab_orgs =
+  let github_org_rows = rows "github" github_orgs in
+  let gitlab_org_rows = rows "gitlab" gitlab_orgs in
+  let org_rows = github_org_rows @ gitlab_org_rows in
```
benmandrew, line 84: Just to clarify we don't want (currently) to interleave the orgs, perhaps in alphabetical order?
Also, were we planning to add sorting? If so then this might have to change in the future.

benmandrew (approved):
Should be good to go once the larger Github profile image is requested.

web-ui/view/index.ml
novemberkilo: Thanks for raising this. I've updated the PR so that the organisations are sorted by name, but at page load via javascript.

benmandrew (approved):
👍

web-ui/view/index.ml
```diff
@@ -32,3 +32,114 @@ let render github gitlab =
           else li [ txt "No GitLab organisations" ]);
         ];
     ]
+
+let rows prefix orgs =
+  let open Organisation.Make (struct
+    let prefix = prefix
+  end) in
+  let orgs = List.sort String.compare orgs in
+  List.fold_left (fun l org -> List.append l [ row ~org ]) [] orgs
```
maiste, line 41:
```suggestion
  List.map (fun org -> [ row ~org ]) orgs
```
Isn't it a map here?

maiste (approved):
LTGM :+1:  Thanks for the PR!

web-ui/view/index.ml
novemberkilo: I think that your suggestion ends up giving us a list of lists and so we'd need to flatten that

PR #629, opened 2022-11-23
Grant admin privileges to some users

maiste:
Privileges are required to handle some actions in `ocaml-ci`.


MisterDA:
:fearful:

maiste:
@MisterDA, do you want to join to? :smile:

maiste:
In fact, it's not needed.

PR #630, opened 2022-11-23
Conform error page to new design

maiste:
This PR intends to conform the error page to the new design.

Here is what it looks like:
image

### DIFF
--- a/web-ui/view/client_error.ml
+++ b/web-ui/view/client_error.ml
@@ -1,12 +1,89 @@
 open Tyxml.Html
 
+let head =
+  head
+    (title (txt "OCaml-CI - Error"))
+    [
+      meta ~a:[ a_charset "UTF-8" ] ();
+      meta ~a:[ a_http_equiv "X-UA-Compatible"; a_content "IE=edge" ] ();
+      meta
+        ~a:
+          [
+            a_name "viewport"; a_content "width=device-width, initial-scale=1.0";
+          ]
+        ();
+      script ~a:[ a_defer (); a_src "/js/alpine.js" ] (txt "");
+      script ~a:[ a_defer (); a_src "/js/alpine-clipboard.js" ] (txt "");
+      link ~rel:[ `Stylesheet ] ~href:"/fonts/inter.css" ();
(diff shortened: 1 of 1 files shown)


web-ui/view/client_error.ml
```diff
@@ -1,12 +1,89 @@
...
+                  [
+                    "flex justify-center flex-col items-center fixed left-0 \
+                     top-0 w-full h-full px-10";
+                  ];
+              ]
+            [
+              h1 ~a:[ a_class [ "text-stroke" ] ] [ txt (Fmt.str "%d" code) ];
+              div
+                ~a:[ a_class [ "text-2xl font-medium my-6" ] ]
+                [ txt (Fmt.str "Opps! %s" reason) ];
```
benmandrew, line 67: Spelling: 'Oops!'

benmandrew (approved):
Looks good to me apart from one spelling error.

web-ui/view/client_error.ml
maiste: Copy & past of the mock-up haha: updated.

web-ui/view/client_error.ml
```diff
@@ -1,12 +1,89 @@
...
+            ~a:
+              [
+                a_class
+                  [
+                    "flex justify-center flex-col items-center fixed left-0 \
+                     top-0 w-full h-full px-10";
+                  ];
+              ]
+            [
+              h1 ~a:[ a_class [ "text-stroke" ] ] [ txt (Fmt.str "%d" code) ];
```
novemberkilo, line 64: Oh cool so this just renders whatever code is passed in. Nice. Did you get a chance to check that this is all ok for 500s as well?

novemberkilo (approved):
Looks good thanks for covering this @maiste 

Figuring this is the case but could you please confirm that this is what is used when `Dream.error` is triggered? And that there are no more instances of the "old" error page?

Please :ship: when ready

web-ui/view/client_error.ml
maiste: Couldn't reproduce one on my side... That's why I wanted to merge it: to make sure it apply.
And yes, this is the error handler given to `Dream` :)

Issue #638, opened 2022-11-25
Github profile pictures rate limit


benmandrew:
On the organisations page, to get Github profile pictures we simply use the Github URL (`github.com/[username].png`). If you refresh the page a few times the images will fail to load with a 429 rate-limit error. (Best to try on https://ci.ocamllabs.io rather than locally because of the number of orgs.)

In the long term we could think about some way of caching the initial results with a timeout, but for now I will implement a fallback Github image like what we have by default for Gitlab.

novemberkilo:
#639 should address this.

PR #639, opened 2022-11-25
Add Github profile picture fallback

benmandrew:
Partial fix for #638.
For testing I recommend breaking the Github URL instead of refreshing the page repeatedly like I spent far too long doing.

In addition, I: 
- Cleaned out old orgs routes that were specific to Github and Gitlab, which were replaced by the new index page routes.
- Made the org sorting on the OCaml side to be case-insensitive, to match the JS behaviour.


novemberkilo (commented):
Thanks for picking this up @benmandrew 

I think we should back out of using the GitHub url entirely so that we never hit rate limiting due to the index page.

So I suggest updating this to just have the GitHub logo for the profile picture.

Perhaps separately we can implement a simple command line tool to fetch profile pictures and store them in `static/images` -- maybe this is something that can be run automatically once a week. What do you think?

novemberkilo:
@benmandrew I've taken the liberty to add a commit that follows through with my comment above. It adds profile pictures for all our known users. I will stitch this into the user-guide for onboarding an organisation ...

maiste (commented):
I understand the approach, but I'm unsure about adding the pictures to the GitHub repository... It makes the repository heavier, and it can keep growing with time. Shouldn't we store them in a docker volume and execute the script from time to time to fill the volume instead?

novemberkilo:
@maiste I am not a fan of having these in git either. From what I remember, a good pattern s to upload these to a content delivery network (CDN) and access them from there. Do you mind if we start here so that we can get past the rate limiting issue?

tmcgilchrist:
> I think we should back out of using the GitHub url entirely so that we never hit rate limiting due to the index page.
As a temporary fix this is correct. There is nothing pressing that says we need to have profile pictures right now.

I think we should avoid storing the GitHub pictures in git, instead they should be fetched from disk. It is possible to have a url to serve them from and populate that cache of profile pictures. We could either populate at startup using a shell script or when we  query GitHub to learn about the Organisations that have installed the GitHub app. The first option is simpler and I would suggest doing that. Given GitHub profile pictures won't change that often, we can live with a stale cache. 

Additionally this works better for GitLab profile pictures which don't see to have a good directly linkable url. So we would need to cache them in a similar way.

novemberkilo:
Thanks all for your input. I've changed things so that we are no longer carrying the pictures in this repo.

Instead, in `Dockerfile.web` we download profile pictures from both GitHub and GitLab.

I've tested that the GitHub side of things works by building and running `Dockerfile.web` (against just the GitHub service). 

I've also updated `repo.ml` to use this machinery.

benmandrew:
Looks good to me. Feel free to merge when you want @novemberkilo.

web-ui/view/organisation.ml
```diff
@@ -1,24 +1,33 @@
...
+    (* /profile-pictures is where images are downloaded -- see Dockerfile.web *)
+    let local_image =
+      Printf.sprintf "/profile-pictures/%s/%s.png" M.prefix org
+    in
+    let fallback_image = Printf.sprintf "/images/%s-logo-500.png" M.prefix in
+    let local_image_exists =
+      try
+        let _ = Unix.stat local_image in
+        true
+      with _ -> false
```
maiste, line 14:
```suggestion
    let local_image_exists = Bos.OS.File.exists (Fpath.v local_image) in
```
As we use `Bos` in `ocurrent`, this is a simple suggestion to avoid dealing with the exception. Feel free to ignore it.

web-ui/view/repo.ml
```diff
@@ -2,17 +2,31 @@ open Tyxml.Html
...
+    let local_image =
+      Printf.sprintf "/profile-pictures/%s/%s.png" M.prefix org
+    in
+    let fallback_image = Printf.sprintf "/images/%s-logo-500.png" M.prefix in
+    let local_image_exists =
+      try
+        let _ = Unix.stat local_image in
+        true
+      with _ -> false
+    in
```
maiste, line 16: Same as for organizations.

maiste (approved):
LGTM :+1:

tmcgilchrist:
@novemberkilo Could you include a note about adding the organisations to these files when people are signing up to `ocaml-ci`?

PR #647, opened 2022-11-30
Instrumenting build history - experiment.

novemberkilo:
DB layer, api layer and a minimal view to
 check if we are collecting history correctly.


benmandrew (approved):
Looks good👍

PR #648, opened 2022-11-30
Dockerfiles arm64 support

benmandrew:
This PR allows `Dockerfile` and `Dockerfile.gitlab` to be built on arm64 machines (`Dockerfile.web` already works).

While `docker build` works on my machine I currently do not have a working local deployment process. I will leave as draft until I or someone else verifies deployment on arm64.

After a while trying to explicitly target the architecture with `uname -m` and variable expansion in the Dockerfile, I realised that by simply removing the `[arch=...]` requirement it targets arm64 on my machine automatically. Could people on other architectures (principally x86_64) please verify that both building and deployment still work on their machines?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,7 @@ RUN opam-2.1 exec -- dune build ./_build/install/default/bin/ocaml-ci-service
 FROM debian:11
 RUN apt-get update && apt-get install libev4 openssh-client curl gnupg2 dumb-init git graphviz libsqlite3-dev ca-certificates netbase -y --no-install-recommends
 RUN curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add -
-RUN echo 'deb [arch=amd64] https://download.docker.com/linux/debian buster stable' >> /etc/apt/sources.list
+RUN echo 'deb https://download.docker.com/linux/debian buster stable' >> /etc/apt/sources.list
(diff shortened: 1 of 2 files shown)


tmcgilchrist:
@benmandrew Can you remove this one too? https://github.com/ocurrent/ocaml-ci/blob/master/Dockerfile.gitlab#L46
All three Dockerfile builds work for me on macOS x86.

benmandrew:
@tmcgilchrist that change was already made in the original commit.

maiste (approved):
The building work on my Lenovo Thinkpad with `amd64` :+1:

kit-ty-kate:
How is that supposed to work if the base image is x86_64? https://github.com/ocurrent/ocaml-ci/pull/648/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R1

benmandrew:
@kit-ty-kate on my M1 laptop pulling the base images appears to retrieve the correct architecture from Docker Hub. This works for both `opam` and `debian`.

Have I got this right? (I'm very new with Docker.)

```bash
$ docker image inspect ocaml/opam:debian-11-ocaml-4.14@sha256:03668731d460043acc763d35e1d5dfc6e6fe68a02f987849ac74f855e3e42c10
...
"Architecture": "arm64",
"Variant": "v8",
"Os": "linux",
...
```

PR #651, opened 2022-12-01
Refactoring the forge: factorize everything.

### LINKED ISSUE #645: Factorize code for views

maiste:
This fixes #645.

This is a big PR, but it's just a merge of the modules from `view/gitlab`.ml and `view/github.ml`. It includes a new `Git_forge.Make(...)` functor to build the view based on a small quantity of information (`module type Forge`). Thanks to this PR, we won't have to duplicate the code when updating a forge view, and it's easy to add a new forge. Moreover, it ensures that the design stays consistent and makes it easier to include new features in the view.

The `history.ml` just extracts the legacy code, which should be removed when working on the new design for the history page.

IMHO, it can be worse not to wait too much before merging this PR; otherwise, it will be a nightmare to backport the code as it moves many files.


benmandrew (approved):
Looks good to me, some lovely refactoring👍

PR #658, opened 2022-12-06
History page to new design.

novemberkilo:
Small changes:

- [x] Fix 500s that result from a missing commit message (I think these are only available for active refs and we have to persist them). We now fallback to the hash itself.
- [x] New designs for history page. 
- [x] Fix incorrectly persisted build status.

Screenshot 2022-12-06 at 4 29 55 pm

### DIFF
--- a/lib/index.ml
+++ b/lib/index.ml
@@ -396,10 +396,7 @@ let _record_build_summary t ~owner ~name ~hash ~gref ~status
           FLOAT current_time;
         ]
   in
-  match status with
-  | `Not_started | `Pending -> ()
-  | `Failed -> v 0
-  | `Passed -> v 1
+  v (status_to_int status)
 
 let record ~repo ~hash ~status ~gref jobs =
   let { Repo_id.owner; name } = repo in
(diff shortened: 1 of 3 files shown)


benmandrew:
Link to branch in breadcrumbs on history page results in 404. For example, in the image clicking on 'main' leads to `/Otorus/main`, when it should be the head commit `/Otorus/commit/xxxxxx` instead (or at least redirect there).

Screenshot 2022-12-06 at 11 21 50

benmandrew (changes requested):
Link to branch in breadcrumbs on history page results in 404. For example, in the image clicking on 'main' leads to `/Otorus/main`, when it should be the head commit `/Otorus/commit/xxxxxx` instead (or at least redirect there).

Screenshot 2022-12-06 at 11 21 50

Also, having 'Build history for [branch]' with 'Build history for branch [branch] on [repo]' seems like unnecessary duplication of information, could we rewrite it so that the text underneath only specifies the repo somehow?

novemberkilo:
Closing in favour of #659

PR #659, opened 2022-12-06
Persist commit message in build summary.

novemberkilo:
Persist the commit message in the `ci_build_summary` table with a fallback.

This builds on #658 and **includes a migration**

Screenshot 2022-12-06 at 5 12 16 pm


novemberkilo:
@benmandrew I'm not sure what you are asking exactly so I'm not sure if this answers your question. This PR persists commit messages if they are available (in `Repo_map`) and falls back to the hash. The `-` you see in the screen shot is the default value of the new `message` column and will appear for commits that pre-date this PR.

novemberkilo:
This PR subsumes #658

novemberkilo:
Please note that this PR also introduces the "Build History" Button on the build page.

Screenshot 2022-12-07 at 8 01 37 pm

web-ui/view/build.ml
```diff
@@ -35,13 +36,13 @@ let title_card ~status ~card_title ~hash_link ~ref_links ~first_created_at
...
                       ~a:[ a_class [ "text-xl" ] ]
                       [ txt (Common.truncate ~len:80 card_title) ];
-                    (* a
-                       ~a:
-                         [
-                           a_class [ "btn btn-secondary btn-sm ml-4" ];
-                           a_href "#";
-                         ]
-                       [ txt "View Build History" ]; *)
+                    a
```
benmandrew, line 39: Comment refers to line 24 above:

Change `~a:[ a_class [ "justify-between items-center flex" ] ]` to `~a:[ a_class [ "justify-between items-center flex space-x-4" ] ]`; currently there is no enforced spacing between the build history button and the 'ran-for' text. This matters only for very long commit messages, as in the screenshot below:

Screenshot showing build history button next to the 'ran-for' text with no spacing

web-ui/view/step.ml
```diff
@@ -116,9 +116,11 @@ module Make (M : Git_forge_intf.Forge) = struct
       :: Common.rebuild_button ~hash ~csrf_token ~show:show_rebuild ()
     in
     let title_card =
+      let ref_path = Result.get_ok (M.ref_path (List.hd refs)) in
```
benmandrew, line 119: `refs` is the list of refs that have this commit as their *head*, for non-head commits clicking on them in the history page causes a 500 error due to the `List.hd`.

This is due to the step page having the 'build history' button. We could either discriminate between head commits *with* the button and historical commits *without*, or somehow pass the correct refs to historical commits.

benmandrew (changes requested):

web-ui/view/step.ml
maiste: Considering our time, I suggest removing the `history` button for the `history` commit. Based on the suggestion, I tried [this ugly piece](https://github.com/maiste/ocaml-ci/commit/95bc6749f7c5c1ad3efa10dc63ac2acc0c7f6a20) of code, and it's working as expected. IMHO, this is the best solution. Also, we should remove the rebuild button for this solution for now.

maiste (commented):
Same issue as Ben with the history button that makes the step page crash when using the history. Otherwise looks good to me. This for the work done :+1:

web-ui/view/step.ml
novemberkilo: Thanks. That piece of code also came to me in the middle of the night ... must be the right solution ;)

novemberkilo:
Thanks for the extensive reviews @benmandrew and @maiste -- the history button and routes are not available for GitLab for now and I've cleaned up the logic for when we show the history button as best as I can.

Screenshot 2022-12-08 at 9 24 50 am

web-ui/controller/git_forge.ml
```diff
@@ -134,7 +134,12 @@ module Make (View : View) = struct
       | `Pull n -> Fmt.str "refs/pull/%s/head" n
```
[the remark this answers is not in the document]
novemberkilo: Or perhaps are you referring to https://github.com/ocurrent/ocaml-ci/pull/659/files#diff-1a6e4157f0f2a9919d30dd1beb373497023e547350cdd3b0921423812ca04074R27

PR #657, opened 2022-12-05
UX/UI: support more features for users

maiste:
This PR intends to fix some missing features in the ocaml-ci web UI. Here is a list of the new added:
- [ ] Fix #656:
  - [X] Organisation
  - [x] Repo
  - [x] Refs
  - [ ] Build
  - [ ] Step 
- Introduce dark mode:
  - [ ] Organisation
  - [ ] Repo
  - [ ] Refs
  - [ ] Build
  - [ ] Step

PR #660, opened 2022-12-06
Responsive text truncation for commit messages

benmandrew:
The existing truncation of commit messages is server-side and to a constant number of characters, which is both unresponsive and ignores differing character widths in the font. This change affects the refs page and the build page, moving the truncation to pure CSS.

Related to the mobile UI in #657.


maiste:
Done in various PRs.

PR #661, opened 2022-12-06
Reinit opam (image default is 2.0, but we're using 2.1)

MisterDA:
In 0e785ca822b5860bf01e8af56e598e1fe0cf9260 I've changed the analyze opam-vars to use the current opam variant instead of the default, however this requires also to reinit the opam root, otherwise false information may be returned by `opam config expand …`.

@kit-ty-kate explained that if we change the default opam (which is also the case in ocaml-ci specs) we ought to reinit opam too.

The rest are updates.


kit-ty-kate (approved):

tmcgilchrist (approved):

maiste (approved):

MisterDA:
Alea jacta est.

Issue #656, opened 2022-12-05
Mobile version


maiste:
- [X] Organisation (#662)
  - [X] Repo (#663)
  - [X] Refs (#664)
  - [X] Build #674 
  - [ ] Step #676 
  - [ ] History #673

PR #662, opened 2022-12-07
Mobile version: organisation

maiste:
This is part of #656. It supersedes a part of #657.

It fixes an issue when it can't find the organisation page in the page chosen and implements a version for mobile of the organisation page.

### DIFF
--- a/web-ui/view/organisation.ml
+++ b/web-ui/view/organisation.ml
@@ -8,9 +8,13 @@ module Make (M : Forge_prefix) = struct
     in
     let fallback_image = Printf.sprintf "/images/%s-logo-500.png" M.prefix in
     let local_image_exists =
-      Result.is_ok @@ Bos.OS.File.exists (Fpath.v local_image)
+      match Bos.OS.File.exists (Fpath.v local_image) with
+      | Ok b -> b
+      | Error _ -> false
     in
+    Logs.warn (fun l -> l "Looking for %s@." local_image);
     let url = if local_image_exists then local_image else fallback_image in
+    Logs.warn (fun l -> l "Image is %S" url);
     Tyxml.Html.(
       img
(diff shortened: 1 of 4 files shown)


maiste:
organisation

benmandrew:
On Safari mobile the text does not wrap correctly. For some reason the flex-col is not contained within the container, as can be seen in the image below. I tried messing around with the CSS but couldn't find a fix for it.

Git forge URLs overflow outside of their containing item-cards

This is not just for unrealistically small screen sizes: the problem persists for iPhone 8 and 8 Plus.

benmandrew (changes requested):
Above

web-ui/view/organisation.ml
```diff
@@ -8,9 +8,13 @@ module Make (M : Forge_prefix) = struct
...
     let fallback_image = Printf.sprintf "/images/%s-logo-500.png" M.prefix in
     let local_image_exists =
-      Result.is_ok @@ Bos.OS.File.exists (Fpath.v local_image)
+      match Bos.OS.File.exists (Fpath.v local_image) with
+      | Ok b -> b
+      | Error _ -> false
     in
+    Logs.warn (fun l -> l "Looking for %s@." local_image);
     let url = if local_image_exists then local_image else fallback_image in
+    Logs.warn (fun l -> l "Image is %S" url);
```
novemberkilo, line 17: Maybe drop the logs if you don't need them anymore?

novemberkilo (approved):
Thanks @maiste - you can add CSS to your list of expertise now :)

novemberkilo:
Regarding the issue raised by @benmandrew - I suggest shipping this and opening an issue. I'm not sure we're going to be able to support a perfect experience for portrait mode on phones on all browsers. Let's collect the issues and see if there are any deal breakers?

PR #663, opened 2022-12-07
Mobile version: repo page

maiste:
This is part of #656. It supersedes a part of #657.

It factorizes some parts of the code and adds support for the repo page in the mobile version.

It requires #662 to be merged first, as it uses some CSS changes made on the general value.

### DIFF
--- a/web-ui/static/css/style_v1.css
+++ b/web-ui/static/css/style_v1.css
@@ -44,7 +44,7 @@ table {
 .custom-table tr th,
 .custom-table tr td {
   padding: 5px;
-  @apply px-6 py-4 border-b border-gray-200;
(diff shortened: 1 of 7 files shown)


maiste:
repo

benmandrew:
On Safari mobile, the repos table extends off of the right side of the screen for small enough screen-widths. This can be fixed by removing `min-width: 350px` from `table` in `main.css`.

Screenshot showing repos table extending off of the right side of the screen

novemberkilo (approved):
Thanks @maiste - I can confirm that things look ok in Firefox but less so in Safari. I am ok with shipping this as with the organisations work.

PR #664, opened 2022-12-07
Mobile version: ref page

maiste:
This is part of https://github.com/ocurrent/ocaml-ci/issues/656. It supersedes a part of https://github.com/ocurrent/ocaml-ci/pull/657.

It needs #663 to be merged first.

In the mobile version, I removed much information to make it more transparent for the users when reading on a small screen. The details information will be available on the build page. It uses parts of  Ben's CSS trick to truncate the code (#660).

### DIFF
--- a/web-ui/static/css/style_v1.css
+++ b/web-ui/static/css/style_v1.css
@@ -44,6 +44,7 @@ table {
 .custom-table tr th,
 .custom-table tr td {
   padding: 5px;
+  min-height: 80px;
   @apply md:px-6 md:py-4 border-b border-gray-200;
 }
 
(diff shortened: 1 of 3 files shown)


maiste:
ref

benmandrew:
Commit messages are not truncating correctly, it seems the containing div is slightly too long and so the ellipses are hidden by the `truncate` class.

Screenshot 2022-12-07 at 12 59 15

benmandrew (changes requested):
Above

novemberkilo (approved):
Thanks @maiste - as with the other 'mobile version' PRs ... still some issues with Safari on mobile but overall much improved so happy to ship.

PR #666, opened 2022-12-07
Replace usage of `socketpair` for IPC

MisterDA:
On Windows, using a socket as standard input file descriptor is daring. Even if it may be possible, Lwt will currently outright reject it. We can, instead of passing a file descriptor to the spawned process, pass the address of the UNIX domain socket as an argument so that the sub process may open a socket itself and connect to that address.

UNIX domain sockets are supported on Windows 10 and OCaml 4.14.

This is a "backport" from code intended to go in solver-service, but I'm also trying it here in the meantime. I've just realized that there's still a minor problem on Windows.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -99,7 +99,7 @@
   (opam-0install (>= "0.4.3"))
   (git-unix (>= 3.9.0))
   (timedesc (>= 0.9.0))
-  (ocaml (>= 4.13))
+  (ocaml (>= 4.14.0))
   (capnp-rpc-unix (>= 1.2)))
  (conflicts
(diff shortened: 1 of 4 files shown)


MisterDA:
The problem _seems_ to be some kind of race condition with the filesystem when a file was created with `let path = Filename.temp_file ...` and binding an Unix domain socket with the same path as this address (even though the path had been unlinked).
We can't either really test that the socket exists with `Sys.file_exists` (`_wstati64` below).
I've kept the code simple and added basic error checking if the solver process doesn't connect within a timeout of 1sec, or fails to start and an interrupt signal is send to the parent.
I've tested the code under Linux and Windows, and I'm quite confident I've found a sweet spot.

PR #668, opened 2022-12-07
Repos Page v2

benmandrew:
This PR introduces statistics into the repositories page. Build frequency still needs to be implemented, and has a placeholder value for the moment.

Screenshot 2022-12-07 at 21 33 49


novemberkilo:
I've taken a look and spun this up locally @benmandrew - looks like you've got most of the parts in place and that this is still a WIP? I'm wondering this because I see that your `onload` javascript fetches data from an api controller but that you are not loading this js ... and instead are loading up data directly as part of the server generated page. Maybe this is left over from previous work and you have decided not to fetch information as json via an api route and instead just load the chart data in as part of the page?


btw I see some `nan` output locally 

Screenshot 2022-12-08 at 12 45 52 pm

I wonder if it's possible to deploy this under a separate route that is not hooked up to the router so that we can "test in production." What do you think of that approach?

benmandrew:
The new testing route is `/:org/new`. Please check the old repos page still works as it should as I slapped it together from the current master branch. I will now work on implementing the 'build frequency' column.

novemberkilo:
Thanks @benmandrew 

I haven't had time to look at the code but I did run up the branch locally and can confirm that the new page shows up under the new route and that it does not have the `nan` issue anymore. I see `N/A` where I was seeing `nan`and I am seeing sensible numbers for build frequency.

PR #669, opened 2022-12-08
History buttons per ref.

novemberkilo:
After deploying the history button to production, I've noticed that we've got a problem for commits that are associated to multiple refs.

Screenshot 2022-12-08 at 3 16 08 pm

Here it is not clear what the single build history button should do. It goes to one of the two refs and if you want to see the history of the other ref, you can't.

When we originally designed this we talked about collecting information about which ref you've arrived from in the url of the build page. On reflection this is complicated and is error prone and is introducing state into an inherently stateless resource (the build page). 

So, I propose showing a smaller button next to the ref:

Screenshot 2022-12-08 at 4 46 23 pm
Screenshot 2022-12-08 at 4 47 31 pm

This is simpler, and now removes the need to find ways of encoding state information into URLs, thus maintaining their RESTfulness. :)

### DIFF
--- a/web-ui/controller/git_forge.ml
+++ b/web-ui/controller/git_forge.ml
@@ -12,7 +12,7 @@ module type Controller = sig
   val list_history :
     org:string ->
     repo:string ->
-    gref:[ `Branch of string | `Pull of string ] ->
+    gref:[ `Branch of string | `Request of int ] ->
     Backend.t ->
     Dream.response Lwt.t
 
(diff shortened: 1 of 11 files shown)


novemberkilo:
@maiste I have included a commit to support history URLs for GitLab too. I think this should work but perhaps you can add a commit to fix things if it does not.

maiste (approved):
I added a commit with new stuff. It uses the `[Branch of string | Request of int]` instead of string to remove unnecessary type construction and destruction. This makes the code more explicit about what is manipulated and allows us to fix link issues.
The rest is here to make the code works with `Gitlab` by changing the URLs and using the previous fix.

benmandrew (approved):
LGTM 👍

PR #670, opened 2022-12-08
Hotfix for web UI

maiste:
Fix the issue with the truncation of `ref` in the `ref.ml` page.

### DIFF
--- a/web-ui/static/css/style_v1.css
+++ b/web-ui/static/css/style_v1.css
@@ -78,6 +78,7 @@ table {
 }
 
 .icon-status {
+  min-width: 2rem;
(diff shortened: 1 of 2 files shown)


novemberkilo (approved):
Thanks @maiste -- could you also please add a fix (or fix separately to this PR if that makes more sense) to the issue I have mentioned here?

novemberkilo:
@maiste I'm not sure if this is an issue from this hotfix or if it's something else but it looks like the text in all the buttons on the refs page is getting truncated

Screenshot 2022-12-09 at 5 35 06 pm

maiste:
This is what this fix is intended to fix 😄

PR #672, opened 2022-12-08
Pin version of git-unix.

novemberkilo:
Build [fails for 3.10.1](https://ci.ocamllabs.io/github/ocurrent/ocaml-ci/commit/b4dea9200ea9781ca495627167fb2946617ceccd/variant/alpine-3.16-4.14_opam-2.1#L1217-1217) 

So pinning the version until we know more.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -97,7 +97,7 @@
   ocaml-ci-api
   (conf-libev (<> :os "win32"))
   (opam-0install (>= "0.4.3"))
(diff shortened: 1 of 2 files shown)


benmandrew:
This was due to a breaking change in one of `git-unix`'s dependencies, PR fix has been merged but not yet released. (https://github.com/mirage/ocaml-git/pull/607)

PR #680, opened 2022-12-12
Aggregation bugfix

benmandrew:
The correct default ref was not being stored for each repository, leading to the aggregated state being taken from a seemingly random ref.


benmandrew:
(This PR has been tested in production.)

novemberkilo (approved):
Seems sensible to me - thank you @benmandrew

PR #681, opened 2022-12-13
Adding build-repro section to the step page.

novemberkilo:
Extract the build-reproduction section of the logs into a collapsible section. This happens at the end of a build - i.e. when the logs have stopped streaming. The `Copy code` button then copies the text to the clipboard.

Screenshot 2022-12-13 at 1 06 49 pm


maiste (commented):
image

I have an issue on Firefox.107.0.1 and Chromium.108.0.5359.98. It doesn't seem to be able to display the code in the right place. It stays into the logs.

benmandrew:
The JS relies on the newly added `END-REPRO-BLOCK` inserted into the logs, but for builds done prior to this change the JS fails as the tag doesn't exist. In this case I changed the JS to hide the whole container, as it looks confusing to have the header with no code.

benmandrew (approved):

maiste (approved):
Good to go :+1:

PR #686, opened 2022-12-13
Repo chart tooltips

benmandrew:
Tooltips in Chart.JS are more complex than anticipated. They are typically rendered on the canvas but as our canvas is too small they get cut off at the edges, so we have to write a custom handler that generates HTML instead.

(Also included is a fix to correct the order of commits in the chart, and one allowing speeds to be rendered in hours.)


benmandrew:
Left until later due to complexity.

PR #687, opened 2022-12-14
Fix build rebuild button.

novemberkilo:
The Rebuild button (on the build page) does not respond with the usual All/Failed options. Looks like this broke when I added `truncate` classes to various things on the build page. This PR fixes that.


novemberkilo:
@maiste I see this on a 500 - as you say, I did not expect to see the text 'the page you requested could not be found.' Perhaps I am missing something but I only see one text block for error pages, hence the fix.
Screenshot 2022-12-14 at 6 59 16 pm

web-ui/view/step.ml
```diff
@@ -583,7 +584,7 @@ module Make (M : Git_forge_intf.Forge) = struct
                   ]))
       in
       Printf.sprintf "%s%s"
-        (List.fold_left aux "<table id='steps-table'><tbody>" data)
+        (List.fold_left aux "<table class='steps-table flex'><tbody>" data)
```
maiste, line 587: Can we keep the id here along with the class? We want it to be unique, no?

maiste (approved):
LGTM! I merge the dark PR as soon as the CI goes green. Sorry for the misunderstanding on the error page; I misread the code. That's a good idea to not say it is an error about finding the page yes :smile:

web-ui/view/step.ml
novemberkilo: This is actually why I changed it - there are multiple tables that appear according to the streaming mechanism and so `id` is not ok anymore.

novemberkilo: The break in the column in the logs screenshot above corresponds to a new table being formed.

PR #688, opened 2022-12-14
Repos page UI fixes

benmandrew:
- Reverse the order of the build history chart so that oldest-to-newest is left-to-right.
- Allow the average speed to be rendered in hours when it is above 60 minutes.


novemberkilo (approved):
🚀 thanks @benmandrew

benmandrew:
@maiste no it should not (tested on my end).

PR #689, opened 2022-12-14
Step: cancel mechanism.

maiste:
This PR introduces the capacity for a step to be cancelled.  The cancelling button is available only if the step is running. Once it's done, the step will appear as `Job cancelled`. It takes a bit of time before the step is visible as cancelled in the `backend` machinery, so the propagation might take a bit of time before being visible.

### DIFF
--- a/web-ui/router.ml
+++ b/web-ui/router.ml
@@ -63,6 +63,19 @@ let gitlab_routes gitlab =
           ~org:(Dream.param request "org")
           ~repo:(Dream.param request "repo")
           ~gref gitlab);
+    Dream.post "/gitlab/:org/:repo/commit/:hash/variant/:variant/cancel"
+      (fun request ->
+        Dream.form request >>= function
+        | `Ok _ ->
+            Controller.Gitlab.cancel_step
+              ~org:(Dream.param request "org")
+              ~repo:(Dream.param request "repo")
+              ~hash:(Dream.param request "hash")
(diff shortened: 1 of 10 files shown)


novemberkilo (approved):
I've walked through the code and this all seems pretty sensible to me. I have not run it up locally to check though. Thanks for covering this @maiste

MisterDA:
The codebase already uses `~cancellable` throughout (mostly in ocurrent). I suggest renaming `can_cancel` to `~cancellable`.

web-ui/router.ml
```diff
@@ -182,6 +195,19 @@ let github_routes github =
...
+      (fun request ->
+        Dream.form request >>= function
+        | `Ok _ ->
+            Controller.Github.cancel_step
+              ~org:(Dream.param request "org")
+              ~repo:(Dream.param request "repo")
+              ~hash:(Dream.param request "hash")
+              ~variant:(Dream.param request "variant")
+              request github
+        | _ ->
```
MisterDA, line 208: Isn't there a more explicit error message we could use, or is that on purpose?

maiste:
The `web-ui` use `can_cancel` everywhere, so I'll stick to that.

web-ui/router.ml
maiste: Just a duplication of the build error message for the form.

PR #698, opened 2022-12-19
Fix pasting logs into GitHub code blocks.

### LINKED ISSUE #654: Copying logs ends up pulling in line numbers (bug)
To track the issue raised here https://github.com/ocurrent/ocaml-ci/issues/635#issuecomment-1326724311

novemberkilo:
Fixes #654 

GitHub support pointed to https://github.com/github/paste-markdown#excluding-tables so we're adding the magic data attribute to the logs table as directed.

### DIFF
--- a/web-ui/view/step.ml
+++ b/web-ui/view/step.ml
@@ -597,7 +597,9 @@ module Make (M : Git_forge_intf.Forge) = struct
       in
       Printf.sprintf "%s%s"
         (List.fold_left aux
-           "<table class='flex steps-table fg-default bg-default'><tbody>" data)
+           "<table data-paste-markdown-skip class='flex steps-table fg-default \
+            bg-default'><tbody>"
+           data)
         "</tbody></table>"
     in
     let open Lwt.Infix in


novemberkilo:
Passes local testing. Makes the line number issues from other browsers disappear as well.

benmandrew:
I am still getting line numbers being pasted on Safari 15.2, see below. According to the [`paste-markdown` repo](https://github.com/github/paste-markdown) we need to `npm install` the package and then import the function into our JS file. I tried doing this myself but ran into issues where it requires an ES6-style import or something along those lines.

```
  - install dune 3.6.1
176	
177	
178	
-> retrieved dune.3.6.1  (cached)
179	
-> installed dune.3.6.1
```

novemberkilo:
Ah hmm yes indeed Safari picks up the line numbers doesn't it. And Firefox is introducing newlines where there aren't any. Still, this is an improvement over the complete mangling that Firefox was doing.

I don't know about installing `paste-markdown` though. If I understand correctly this library is only relevant for applications _into_ which you are pasting markdown. The magic attribute is what we are required to use to tell GitHub's paste-markdown machinery to not try and paste the table as markdown.

novemberkilo:
I've asked a follow up question to GitHub support to see if they can provide any further guidance. Depending on what they come back with I might have to open an issue on `paste-markdown`

novemberkilo:
This seems as close as we're going to get without redoing the way we do line numbers entirely (using CSS instead of putting them in the table). (See comments on the parent issue)

PR #700, opened 2022-12-19
Various UI Fixes

benmandrew:


novemberkilo (approved):
Thanks @benmandrew 

FYI I have reverted the change regarding the extra line in the steps-to-repro block. And I've applied a coat of polish to your work on the line number issue by switching from using ids to data attributes.

PR #701, opened 2022-12-19
Repo chart commit links

benmandrew:


benmandrew:
Right now, if you click on a bar in the repo bar chart it should take you to the specific commit page. Due to it overlapping with the parent row that itself is a link, there's some trickery involving cancelling using `evt.preventDefault()`.

Currently not user-friendly as there is no feedback that these links can be followed, but tooltips will help with this. Also, very small bars can be hard or even impossible to click.

novemberkilo (approved):
Thanks much for seeing this through to completion @benmandrew - stellar work 👍🏾 

I've added a couple of commits to iron out some wrinkles and add some small polish. Will deploy.

PR #704, opened 2022-12-23
Add johnyob to allowlist

johnyob:


novemberkilo:
@johnyob you should be all setup to use ocaml-ci. Please try adding a repository and it should show up on https://ci.ocamllabs.io

PR #721, opened 2023-01-06
Bug fix proposition for 703

### LINKED ISSUE #703: Build history shows incorrect build statuses
The build statuses of historical builds are wrong. Here is an example:

Build history page for `master` of ocaml-ci on 2022-12-23 - Note that `fc937b` is showing as pending

Screenshot 2022-12-23 at 12 39 04 pm

However, this is not correct. Following the link to the build of `fc937b` shows the build has completed and should be in state `Failed` -- Further, the build page of this commit shows a default (or unknown) build status.

Screenshot 2022-12-23 at 12 39 59 pm

maiste:
## Context

This PR intends to fix #703 and avoid losing the information when the job is auto cancelled. To summarise a bit of the issue here, if the job is cancelled by `ocurrent`, we don't trigger `Index.record`, and we take the last entry (`Pending`).

## Step to reproduce the bug

On a repository tracked by `ocaml-ci`:

1. Push a commit to a specific branch.
2. Wait for the job to start and before it's finished, push another commit.
3.  The aggregate state and the history will stay in the `Pending` state.

## Solution

The solution adds a hook to the cancel mechanism and inserts a new entry to the `ci_build_summary` database when triggered by `ocurrent`. It is executed when the job is not reported as `Job complete`.

This solution is, IMO, _ugly_. However, I didn't find a better one. The other possibility was to clean the state when this information is required by `ocaml-ci-web` and update the database, but we would have lost a lot of information about the commit at this point.

I would be happy to replace it with another implementation if there is a better one :owl:

### DIFF
--- a/lib/cluster_build.mli
+++ b/lib/cluster_build.mli
@@ -9,6 +9,7 @@ val config :
 
 val v :
   t ->
+  ?on_cancel:(string -> unit) ->
   platforms:Platform.t list Current.t ->
   repo:Repo_id.t Current.t ->
   spec:Spec.t Current.t ->
@@ -19,4 +20,5 @@ val v :
     platform. [~repo] is the ID of the repository-under-test on a Git Forge
     (e.g. GitHub or GitLab).
 
+    @param on_cancel The callback function to call if the job is cancelled.
     @param repo The ID of the repository-under-test on GitHub. *)
--- a/lib/index.ml
+++ b/lib/index.ml
@@ -442,7 +442,7 @@ module Commit_cache = struct
     { s; started_at; ran_for }
 end
 
-let _record_build_summary t ~owner ~name ~hash ~gref ~status ~commit_message
+let record_build_summary t ~owner ~name ~hash ~gref ~status ~commit_message
     ~(variants_timestamps : (string * Run_time.timestamps option) list) =
   let ts = List.filter_map snd variants_timestamps in
   let first_queued_at = Run_time.first_step_queued_at ts in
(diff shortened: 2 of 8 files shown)


lib/index.ml
```diff
@@ -477,6 +477,91 @@ let _record_build_summary t ~owner ~name ~hash ~gref ~status ~commit_message
...
+             total_ran_for,
+             ran_for,
+             message )
+       | row ->
+           Fmt.failwith "record_summary_on_cancel: invalid row %a" Db.dump_row
+             row
+  in
+  match last with
+  | [] -> record_empty_summary t ~repo ~gref ~hash
+  | xs -> (
```
benmandrew, line 535:
```ocaml
| (build_number, status, started_at, total_ran_for, ran_for, message)::_ -> (
```
I'm not a fan of `List.hd`, I prefer explicit pattern-matching.

benmandrew (commented):
An `on_cancel` hook is a fairly natural solution IMO, I'm happy with it.

benmandrew (approved):

lib/cluster_build.mli
```diff
@@ -16,4 +17,5 @@ val v :
 (** Build and test all the opam packages in a given build context on the given
     platform.
 
+    @param on_cancel The callback function to call if the job is cancelled.
```
tmcgilchrist, line 20:
```suggestion
    @param on_cancel The callback function executed if the job is cancelled.
```

tmcgilchrist (approved):

novemberkilo (approved):
Seems all very sensible to me. Thanks @maiste

PR #722, opened 2023-01-09
Fix for horizontal scrolling on step page

### LINKED ISSUE #716: Logs background is not consistent when scrolling horizontally
This is on mobile but it happens on desktop too. You might have to resize the browser window to trigger it.

benmandrew:
Fixes #716.


maiste (approved):
LGTM :+1: The fix works on `Firefox.108.0.1`.

PR #725, opened 2023-01-11
Introduce issue templates

maiste:
Being through the various PRs and issues, it was hard to sort out the difference between the requests, discussions and bugs. This is a proposal for the new coming issues that will be raised in the future. Having a bug report and feature request model can make things a bit clearer for us by already assigning the right label and having a common pattern to understand the issue or the request.

My hope is that it will force reporters (us included) to describe the problem in a more structured way. The `blank issue` is here to keep the possibility of opening a specific issue.

WDYT?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,32 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: type/bug
+assignees: ''
+
+---
+
+## Context
+
+<!-- A clear and concise description of what the bug is. -->
+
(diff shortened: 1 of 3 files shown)


benmandrew (approved):

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: type/bug
+assignees: ''
```
MisterDA, line 6: I think you should add @benmandrew here 🥇

tmcgilchrist (commented):
Sounds good.

PR #727, opened 2023-01-13
Add lower-bounds builds

benmandrew:
We do a second call to `solver-service` to get the lower-bounds selection of packages. This PR depends on https://github.com/ocurrent/solver-service/pull/34.

Currently all variants are built twice, once normally and once with lower bounds. However this is expensive and it may be more practical to only build lower-bounds variants with the earliest listed version of OCaml.

### DIFF
--- a/lib/analyse.ml
+++ b/lib/analyse.ml
@@ -228,21 +228,33 @@ module Analysis = struct
     let selection = List.hd selections in
     (selection.Selection.commit, selection)
 
-  let get_lower_bound_platforms platforms =
+  let filter_lower_bound_distros platforms =
+    (* [distro] is a subtype of [t] but the type checker needs some help for polymorphic variants *)
+    let t_of_distro x = Dockerfile_opam.Distro.((x : distro :> t)) in
+    let latest_debian =
+      Dockerfile_opam.Distro.(
+        tag_of_distro @@ t_of_distro @@ resolve_alias (`Debian `Stable))
+    in
     List.filter
-      (fun (v, _) -> String.equal (Variant.distro v) "debian-11")
+      (fun (v, _) -> String.equal (Variant.distro v) latest_debian)
       platforms
-    |> List.fold_left
-         (fun v v' ->
-           if
-             Ocaml_version.compare
-               (Variant.ocaml_version (fst v))
-               (Variant.ocaml_version (fst v'))
-             >= 0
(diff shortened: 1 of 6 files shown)


MisterDA:
> Currently all variants are built twice, once normally and once with lower bounds. However this is expensive and it may be more practical to only build lower-bounds variants with the earliest listed version of OCaml.

I agree, only earliest version of OCaml and just on the master distro seems enough.

For the CI to succeed, you can temporarily change the branch the solver-service submodule is pointing to.

tmcgilchrist:
I would view this as a linting step rather than a build step, so only run on a single distro (debian ?) with one OCaml version. opam-repo-ci uses debian-11 plus a check on each supported OCaml version, I think ocaml-ci should start with the oldest OCaml version.

service/pipeline.ml
```diff
@@ -141,19 +141,21 @@ let build_with_docker ?ocluster ~repo ~analysis ~platforms source =
             Spec.opam ~label:"(lint-fmt)" ~selection:lint_selection ~analysis
               (`Lint `Fmt)
             :: Spec.opam_monorepo builds
-        | `Opam_build selections ->
+        | `Opam_build (`Default selections, `Lower_bound selections') ->
```
tmcgilchrist, line 144:
```suggestion
        | `Opam_build (`Default selections, `Lower_bound lower_bound_selections) ->
```
Maybe use a name without the prime.

lib/analyse.ml
```diff
@@ -226,6 +228,22 @@ module Analysis = struct
     let selection = List.hd selections in
     (selection.Selection.commit, selection)
 
+  let get_lower_bound_platforms platforms =
+    List.filter
+      (fun (v, _) -> String.equal (Variant.distro v) "debian-11")
```
tmcgilchrist, line 233: This debian-11 should use the distro information from `ocaml-dockerfile`
```
 `Debian `Latest
```
That way it will roll forward when new versions are released.

tmcgilchrist (commented):
Please review how this lower-bounds check displays in GitHub and on the UI. 
I feel like it should be `(lint lower-bounds)` rather than the current distro-ocaml-version-opam-version-lower-bounds format.

You should also include it in the experimental variants so it doesn't fail builds until we are confident with it.

PR #729, opened 2023-01-16
Fix the bottleneck on the analysis

moyodiallo:
I think we missed something, the pool was for cloning (`Current_git.with_checkout`).

### DIFF
--- a/lib/analyse.ml
+++ b/lib/analyse.ml
@@ -2,7 +2,7 @@ open Lwt.Infix
 open Current.Syntax
 module Worker = Ocaml_ci_api.Worker
 
-let pool = Current.Pool.create ~label:"analyse" 2
+let pool = Current.Pool.create ~label:"analyse" 20
 
 let is_empty_file x =
   match Unix.lstat x with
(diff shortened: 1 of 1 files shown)


maiste:
Hi,
Thanks for the PR! Could you please provide a bit more context about why you chose these values and moved the `pool` from start to checkout? :smile:

moyodiallo:
> Hi, Thanks for the PR! Could you please provide a bit more context about why you chose these values and moved the `pool` from start to checkout? :smile:

:smile:  Yes, if you look in to this [Commit](https://github.com/ocurrent/ocaml-ci/commit/e24fc63f5245b364c1c2ea26582abf3815d5bc40) which added the pool, the goal is to prevent having huge number of clones at same time. But when pool is used for analysis with the current state we prevent sending more request to the solver service.

When a job finish cloning (`Current_git.with_checkout` ), preventing another job for starting the analysis is no longer necessary. This is why, it is convenient to use the pool in `Current_git.with_checkout` and the definition of [Current_git.with_checkout](https://github.com/ocurrent/ocurrent/blob/009e5d00645efd44a21cd6c0ad5264bbe0fd1c18/plugins/git/current_git.ml#L81-L109)(good indication Line-100) do exactly what was needed.

tmcgilchrist (approved):
Looks good. Please monitor this in production in case we hit git clone rate limits.

Issue #728, opened 2023-01-15
Steps to reproduce code block is incorrect


tmcgilchrist:
## Context

Steps to reproduce code block is incorrect. 

## Step(s) to reproduce

Use either the `Copy Code` button or manually select the full section of `Steps to Reproduce`. 
The PR section of the git clone is incorrect `git fetch origin "refs/pull//head" ` missing the PR number.
Also the `Copy Code` button doesn't give any visual feedback that anything has been copied. 

The reproductions steps are correct in the log file. I expect the code extracting it from the logs is to blame.

## Expected behaviour

I should be able to use either the button or select a section to copy the reproduction steps.
## Technical details

Using macOS 12.6.2  / Safari 16.2

## Additional context

The bug is present on this build. https://ci.ocamllabs.io/github/realworldocaml/book/commit/fc64a884f84226047d0031984b7058ea75353ffc/variant/debian-11-4.14_opam-2.1

maiste:
Fixed in #731

PR #731, opened 2023-01-17
Fix 'steps to reproduce' code block

benmandrew:
Fix for #728.

With the introduction of CSS line numbers, the JS trying to strip out real line numbers lead to the removal of numbers in the logs themselves.


benmandrew:
My mistake, I have fixed the numbers issue but missed the visual feedback on copying that was also raised in the issue. I will add this to the PR.

maiste (approved):
Thanks! Just add the extra space at the end to make it more readable. 
LGTM :+1:

PR #742, opened 2023-01-26
Switch back to local solver service.

tmcgilchrist:
Temporarily switch to a local solver service while we sort out the stability issues with the solver pool.


maiste (commented):
I'm wondering if this will solve the issue. As it will use the local version of the solver-service, aren't we going to have the same leaks issues? My knowledge of this part of the solver is too narrow to know its exact behaviour in this situation.

maiste (approved):
From the offline discussion, this fix works for now. Good to go as a temporary fix.

Issue #607, opened 2022-11-16
Feedback relating to the design and UX of ocaml-ci


novemberkilo:
As we roll out the new designs, this is an overarching issue to collect feedback from the users of ocaml-ci. Please add a comment describing your feedback - please be specific about the problem (or benefit) you are experiencing and make a suggestion for an improvement. We will triage these and work on them as appropriate.

novemberkilo:
I really like the new CI look, but I was wondering if we could try to to optimise a bit more for using the screen efficiently - the font for logs is quite large for instance (by @samoht)

novemberkilo:
From @tmcgilchrist via slack:

Feedback via other channels:

- Sorting / filtering on this page would be nice https://ci.ocamllabs.io/github/ocaml-gospel/cameleer/commit/11048c
- Show failure / error directly? Scroll to the end?
- Analysis displaying results better (nice to have) https://ci.ocamllabs.io/github/ocaml-gospel/cameleer/commit/11048c/variant/(analysis)#L777-777

novemberkilo:
From @emillon via slack:

> nice! you might want to tweak the CSS though, the contrast is very hard in some cases

(Context - colours in logs)

emillon:
The piece of logs in particular was:

![image](https://user-images.githubusercontent.com/496345/202660015-6b7e4a53-5516-437c-8cf7-1a77d9b9a2c3.png)

According to [contrastchecker](https://webaim.org/resources/contrastchecker/) this is contrast 1.88:1 whereas 4.5:1 is recommended for normal text.

benmandrew:
ANSI colours for code in the logs is still light for some text, despite the light background (for example see the `opam install` part of [any build](https://ci.ocamllabs.io/github/ocurrent/ocurrent/commit/2a710c0fe7448632adbff9aeed1323203d61d60f/variant/alpine-3.15-4.14_opam-2.1)).

novemberkilo:
> ANSI colours for code in the logs is still light for some text, despite the light background (for example see the `opam install` part of [any build](https://ci.ocamllabs.io/github/ocurrent/ocurrent/commit/2a710c0fe7448632adbff9aeed1323203d61d60f/variant/alpine-3.15-4.14_opam-2.1)).

Anyone familiar with `ansi` able to help out with this? I don't understand how it's selecting colours - the logs in opam-health-check do not get these decorations and I don't know why they do here or how we can change them. We are simply doing 

[...]

PR #744, opened 2023-01-27
Add log highlight for dark-mode

benmandrew:
Make it easier to read highlighted text in darkmode. I have switched to standard Tailwind colours that look about right to me, I'm open to suggestions for colour changes.

Feedback from https://github.com/ocurrent/ocaml-ci/issues/607#issuecomment-1405419235.


maiste (approved):
Thanks! :+1: 
Looks good on Firefox: 
image

Issue #746, opened 2023-01-31
Meta: track refactoring in the UI code.


maiste:
This issue is here to track the part of the code we can refactor in the `web-ui` part to make it easier to maintain.

-- Added 2023-01-31 --
- [x] Refactor organisations in `index.ml` to only use one function instead of relying on two `list_orgs` and `list_all_orgs`. Done in #750.
- [ ] Refactor `router.ml` to use the same system to build routes and have only one module to maintain instead of two.
- [ ] #752

PR #747, opened 2023-01-31
Web-ui: remanent bugs from the past

### LINKED ISSUE #626: UI instability on page load
It seems there is some CSS changed directly after a page loads in all pages in the new UI. This is seen primarily in the font changing from a standard to a bold font, but we also see in the build page that the table header is removed. Both of these are visually seen by the user on every page load, even on a refresh of the same page.

This is perhaps related to the tailwind CSS generation introduced in #605?

maiste:
This PR fixes two bugs:
- Partial fix for #626. There were two bugs in one. The `Build` header for the table disappeared as the JS script did not rebuild it. This PR completely removes the header, as it's not needed in the current version (correction 1). This bug didn't appear in the GitLab version, as the links to the API weren't available in the GitLab routes. This PR introduces these links in the GitLab router (correction2). @benmandrew mentioned that we could merge the router as the behaviour is the same, and I agree with his suggestion as it will make the code easier to maintain. 
- It also fixes a bug in the `Organisation` page by conforming it with the "two organisations" pages. Indeed, the design for one organisation differs from the one from the mixed organisation page. These two functions should be merged at some point and display the organisation we ask.
- Fixes #626 for good. It improves the rendering by caching the fonts for a month (for now, as we still move fast), blocking the rendering until the font is loaded and specifying the type of screen for render optimisation.

I opened an issue to track the refactoring we could do: #746

### DIFF
--- a/web-ui/router.ml
+++ b/web-ui/router.ml
@@ -122,6 +122,19 @@ let gitlab_routes gitlab =
         | _ ->
             Dream.log "Form validation failed";
             Dream.empty `Bad_Request);
+    Dream.get "/api/gitlab/:org/:repo/commit/:hash/variant/:variant"
+      (fun request ->
+        Api_controller.Gitlab.show_step
+          ~org:(Dream.param request "org")
+          ~repo:(Dream.param request "repo")
+          ~hash:(Dream.param request "hash")
+          ~variant:(Dream.param request "variant")
+          gitlab);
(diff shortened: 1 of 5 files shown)


benmandrew (approved):
Looks good to me, could you run a `dune fmt`?

One note, this does not fix #626 which mainly mentioned the text being unstyled very briefly after page-load before becoming bold. This issue still exists.
(I think if you merge this PR it will automatically close that issue.)

benmandrew:
👍👍👍

PR #749, opened 2023-02-02
Add new Nomadic Labs projects.

tmcgilchrist:


maiste (approved):
Change LGTM :+1: 
There seems to be an issue with `live-engine`. It's like we can't access the NL repositories.

tmcgilchrist:
Yes thanks, I'm sorting out that issue with NL.

PR #750, opened 2023-02-02
Factorize organisation code

maiste:
This is a partial fix for #746. It factorizes the code for the organisation page to avoid having `list_orgs` and `list_all_orgs` separated functions. The two functions are replaced by a single function, `list_orgs`, that takes a list of `(prefix, name, backend)`.

### DIFF
--- a/web-ui/controller/index.ml
+++ b/web-ui/controller/index.ml
@@ -1,21 +1,21 @@
 module Client = Ocaml_ci_api.Client
 
-let ( >>!= ) = Git_forge.( >>!= )
+let ( >>!!= ) = Git_forge.( >>!= )
+let ( >|!= ) a f = Lwt_result.map f a
 
-let list_orgs prefix ci =
+let list_orgs ~orgs =
   let open Lwt.Infix in
-  Backend.ci ci >>= Client.CI.orgs_detailed >>!= fun orgs ->
-  let orgs = List.filter (fun o -> o.Client.CI.number_repos > 0) orgs in
-  Dream.respond @@ View.Index.list_orgs prefix orgs
-
-let list_all_orgs ~github ~gitlab =
-  let open Lwt.Infix in
-  Backend.ci gitlab >>= Client.CI.orgs_detailed >>!= fun gitlab_orgs ->
-  Backend.ci github >>= Client.CI.orgs_detailed >>!= fun github_orgs ->
-  let github_orgs =
-    List.filter (fun o -> o.Client.CI.number_repos > 0) github_orgs
+  let extract ((prefix, name, backend) : string * string * Backend.t) =
+    Backend.ci backend
+    >>= Client.CI.orgs_detailed
+    >|!= List.filter (fun o -> o.Client.CI.number_repos > 0)
+    >|!= fun orgs -> (prefix, name, orgs)
   in
-  let gitlab_orgs =
-    List.filter (fun o -> o.Client.CI.number_repos > 0) gitlab_orgs
+  let to_result_list orgs org =
+    match org with
+    | Ok org -> Result.map (fun orgs -> org :: orgs) orgs
+    | Error _ -> orgs
   in
-  Dream.respond @@ View.Index.list_all_orgs ~github_orgs ~gitlab_orgs
+  Lwt_list.map_p extract orgs
+  |> Lwt.map (fun orgs -> List.fold_left to_result_list (Ok []) orgs)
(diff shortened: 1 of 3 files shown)


web-ui/controller/index.ml
```diff
@@ -1,21 +1,21 @@
 module Client = Ocaml_ci_api.Client
 
-let ( >>!= ) = Git_forge.( >>!= )
+let ( >>!!= ) = Git_forge.( >>!= )
+let ( >|!= ) a f = Lwt_result.map f a
```
MisterDA, line 4: This is too much. I'd suggest switching to the `Syntax` module instead.
Are you trying to ascii-art spaceships?

web-ui/controller/index.ml
```diff
@@ -1,21 +1,21 @@
...
+    >>= Client.CI.orgs_detailed
+    >|!= List.filter (fun o -> o.Client.CI.number_repos > 0)
+    >|!= fun orgs -> (prefix, name, orgs)
   in
-  let gitlab_orgs =
-    List.filter (fun o -> o.Client.CI.number_repos > 0) gitlab_orgs
+  let to_result_list orgs org =
+    match org with
+    | Ok org -> Result.map (fun orgs -> org :: orgs) orgs
+    | Error _ -> orgs
```
MisterDA, line 17: It should be possible to also use `Result.bind` on `org` if we're starting to use `Result` combinators.

web-ui/view/index.ml
```diff
@@ -13,56 +13,41 @@ let rows prefix orgs =
...
+                  a_onchange "filter(this.value)";
+                ]
+              build_option;
+          ];
+      ]
 
+let list_orgs ~orgs =
+  let org_rows =
+    let generate (prefix, _, orgs) = rows prefix orgs in
+    List.map generate orgs |> List.flatten
```
MisterDA, line 49: You want `List.concat_map`.

MisterDA (changes requested):

web-ui/controller/index.ml
maiste: Kind of :smile:

web-ui/view/index.ml
```diff
@@ -13,56 +13,41 @@ let rows prefix orgs =
...
-  Template.instance
-    [
-      Tyxml.Html.script ~a:[ a_src "/js/index-page-org-search.js" ] (txt "");
-      div
-        ~a:
+let search_bar_options = function
+  | [] | [ _ ] -> []
+  | orgs ->
+      let build_option =
+        option ~a:[ a_value "all" ] (txt "all")
```
benmandrew, line 20: Visual consistency with the capitalised 'GitHub' and 'GitLab'.
```suggestion
        option ~a:[ a_value "all" ] (txt "All")
```

web-ui/controller/index.ml
```diff
@@ -2,20 +2,20 @@ module Client = Ocaml_ci_api.Client
...
-  let open Lwt.Infix in
-  Backend.ci gitlab >>= Client.CI.orgs_detailed >>!= fun gitlab_orgs ->
-  Backend.ci github >>= Client.CI.orgs_detailed >>!= fun github_orgs ->
-  let github_orgs =
-    List.filter (fun o -> o.Client.CI.number_repos > 0) github_orgs
+let list_orgs ~orgs =
+  let open Lwt.Syntax in
+  let extract (prefix, name, backend) =
+    let* ci = Backend.ci backend in
+    let+ detailled_orgs = Client.CI.orgs_detailed ci in
```
benmandrew, line 9: Spelling
```suggestion
    let+ detailed_orgs = Client.CI.orgs_detailed ci in
```

web-ui/view/index.ml
```diff
@@ -13,56 +13,41 @@ let rows prefix orgs =
...
+        :: List.map
+             (fun (prefix, name, _) -> option ~a:[ a_value prefix ] (txt name))
```
benmandrew, line 22: This isn't right, it generates an `option` from every single org. If you have multiple GitHub users for example there will be a GitHub option for every single one.

benmandrew (changes requested):

web-ui/view/index.ml
benmandrew: Yep you're right, my mistake

benmandrew (approved):

novemberkilo (approved):
Thanks for this simplification @maiste 🎉

PR #753, opened 2023-02-03
Allow packages to choose to accept failures on experimental distros

benmandrew:
This PR supports projects including the `x-ci-accept-failures` field in their `.opam` files, for example:
```
x-ci-accept-failures: [
  "macos-homebrew"
]
```
This allows them to specify distros that the package itself is experimental on so that the CI doesn't fail for distros that the maintainers don't want to support. (It's a little jarring to see a big green tick mark by something that's actually failed, but we currently don't have a way of saying "this failed but no big deal".)

The inspiration for the implementation comes from https://github.com/ocurrent/opam-repo-ci/pull/62. Writing the logic in bash is fairly ugly but I'm unsure of another way of testing whether the build fails or not.

Currently we only allow users to specify distros that they want to ignore the results of (following `opam-repo-ci`'s example), but this could be extended to include the architecture for example (`amd64`, `arm64`, etc.).

All feedback welcomed.

### DIFF
--- a/lib/opam_build.ml
+++ b/lib/opam_build.ml
@@ -80,7 +80,9 @@ let rec get_root_opam_packages = function
 let download_cache = "opam-archives"
 
 let install_project_deps ~opam_version ~opam_files ~selection =
-  let { Selection.packages; commit; variant; only_packages } = selection in
+  let { Selection.packages; commit; variant; compatible_root_pkgs } =
+    selection
+  in
   let prefix =
     match Variant.os variant with `macOS -> "~/local" | `linux -> "/usr"
   in
@@ -91,9 +93,6 @@ let install_project_deps ~opam_version ~opam_files ~selection =
   let root_pkgs = get_root_opam_packages groups in
   let non_root_pkgs =
     packages |> List.filter (fun pkg -> not (List.mem pkg root_pkgs))
-  in
-  let compatible_root_pkgs =
-    if only_packages = [] then root_pkgs else only_packages
   in
   let open Obuilder_spec in
   let cache =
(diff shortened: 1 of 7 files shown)


MisterDA:
Impressive feature nonetheless! If it's too cumbersome to read embedded bash, you can always have the script standalone in its own file, and embed it with `ocaml-crunch`.

tmcgilchrist:
The `x-ci-accept-failures` field seems to follow the distro formatting from `ocaml-dockerfile` eg https://github.com/ocaml/opam-repository/blob/master/packages/conf-nlopt/conf-nlopt.1/opam#L24-L28 To be really useful we would like to extend it to cover Windows and use it alongside the `available` attribute to exclude certain arch / OS variants, while still attempting a build on others.

What is the policy of extra meta-data being added to opam files? cc @kit-ty-kate @dra27

maiste:
I have three concerns related to this PR:
- The first one is the introduction of a shell script directly in the `OCaml` code. As we can't rely on the type checker, it could be great to enforce the validity of the shell script, either by having test regression or by moving it to an external file and testing it with something like `shellcheck`.
- Secondly, isn't it duplicating the `available` field of opam? If the package is supposed to work on the arch, it is supposed to work on the arch, no?
- Finally, this PR introduces some "configuration" within the `opam` file, which, IIRC, was what `ocaml-ci` was supposed to avoid by being "opinionated". This was the reason not to use the `install` field but to write our command line to test programs. Isn't it changing the direction the CI was taking? I have personally no problem with that, but we must be clear about the path we want to go :)

WDYT?

lib/opam_build.ml
```diff
@@ -176,31 +175,53 @@ let install_project_deps ~opam_version ~opam_files ~selection =
...
+res=$?;
+test "$res" != 1 && exit "$res";
+ret=1;
+for pkg in %s; do
+  if opam show -f x-ci-accept-failures: "$pkg" | grep -qF "\"%s\""; then
+    echo -e "The package \033[1m$pkg\033[0m failed, but has been disabled for CI on \033[1m%s\033[0m using the \033[1mx-ci-accept-failures\033[0m field in its opam file.";
+    ret=0;
+  fi;
+done;
+exit "$ret";
```
kit-ty-kate, line 204: Instead of faking a job success, I think using the log matcher to display the skip is a much better way to go about it. See https://github.com/ocurrent/opam-repo-ci/blob/bca0f264ad1112310afef12f320b6291de08331d/service/main.ml#L61 for the log matcher rule to add.
```suggestion
for pkg in %s; do
  if opam show -f x-ci-accept-failures: "$pkg" | grep -qF "\"%s\""; then
    echo "The package $pkg failed, but has been disabled for CI on %s using the x-ci-accept-failures field in its opam file.";
  fi;
done;
exit "$res";
```

kit-ty-kate:
> What is the policy of extra meta-data being added to opam files? cc @kit-ty-kate @dra27

It’s on a case by case basis

novemberkilo:
I started out a bit twitchy about setting up ci specific metadata in the definition of a package (i.e. the `opam` file). It seemed like we were tucking in metadata that is unrelated to a package definition. After reminding myself that the process of submitting an opam package involves putting it through repo-ci, it didn't seem as unrelated after all.

If I understand correctly, it seems that by using `x-ci-accept-failures` we are following a path that has been blazed by repo-ci. Thus this approach to flagging experimental builds on ocaml-ci is consistent and sensible. 👍🏾

dra27:
I'm with @maiste on this one - it makes me uneasy! opam-repo-ci isn't allowed to be as opinionated as ocaml-ci - the `x-ci-accept-failures` is a workaround there. In the cases where this is expected to be used at the moment, is the issue that _our workers_ don't correctly build it (in which case, wouldn't having the configuration in ocaml-ci as a list of exceptions be better - we already have "configuration" for the list of monitored users, etc.). On the other hand, if the package itself actually doesn't work on, say, macOS, then that really ought to be reflected in the `available` field (and the solver should recognise that)?

benmandrew:
This is meant to be used in the case where the package itself is not intended for the specified platform. @dra27 for the '_our workers_' point you mentioned we already have the experimental builds, which is currently applied to `macos-homebrew`.

I was not aware of the `available` field, is there any reason opam-repo-ci doesn't use it?

tmcgilchrist:
The functionality I want to allow is speculative builds on potentially failing platforms eg Windows or un-common architectures eg s390x. Such that authors don't constrain away less common platforms which never get looked at again. This feature came up with some https://ci.ocamllabs.io/gitlab/nomadic-labs projects that were failing on s390x and ppc64 that the author would like to support but doesn't want to fail the build in the meantime. I am aware there are other options available :-)

The `available` flag is a stricter constraint that says this package only works on these platforms (or does not work on others) as claimed by the package author.  For example bindings to a C library that only works on Windows.

It seems like this `x-ci-accept-failures` flag is more opam-repo-ci specific and we shouldn't be using it.

dra27:
(sorry - the close was entirely a mis-click!)

@tmcgilchrist - if the repo doesn't build on s390x or ppc64 then that _should_ be constrained away, though? If at some point we start doing opam-health-check on s390x and/or ppc64 (which we be great at some point 🤷) then our response to packages failing would exactly be to constrain them away. At the moment, it sounds as though the opam files are actually incorrect?

kit-ty-kate:
I personally think OCaml-CI should have `x-ci-accept-failures` for similar reason as stated in: https://github.com/ocaml/opam-repository/pull/23235#issuecomment-1428320525

shonfeder:
Given our need to clear up stalled WIP, I will tie break this in favor of closure. If we have another go at extending ocaml-ci with this functionality, it should be on the basis of a refactor to allow sharing the common logic between opam-repo-ci and ocaml-ci.

PR #755, opened 2023-02-03
Restore cancellable opam file lookup

MisterDA:
Now the opam file lookup is executed in a `Lwt_preemptive` thread and should not block the main execution thread, but it is not cancellable anymore.

Restore the possibility of cancelling the lookup by introducing a signalling atomic. Regularly check the atomic: if cancellation was requested, raise to exit early.

Should we raise the limit of preemptive threads? it defaults to 4.

Thanks for the pointers @talex5.

### DIFF
--- a/lib/analyse.ml
+++ b/lib/analyse.ml
@@ -255,7 +255,11 @@ module Analysis = struct
         |> fun s -> [ s ]
 
   let of_dir ~solver ~job ~platforms ~opam_repository_commit root =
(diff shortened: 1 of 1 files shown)


maiste (commented):
LGTM :+1: It works on my side. However, I'm not competent enough to give a point of view about the correctness of the solution.

talex5 (approved):

PR #758, opened 2023-02-10
Deploy using OCaml 5.0 (alea jacta est)

MisterDA:
:fire: :fire: :fire: 
:computer:


tmcgilchrist (changes requested):
I would like an overall plan for moving our services to OCaml 5 before doing this. 

We should definitely be looking at what can move to 5.0 and making plans for how other services can be moved. But I want a comprehensive plan for it.

MisterDA:
We'll retry this later, then.

PR #760, opened 2023-02-16
Fix trailing slash in url

maiste:
This is an attempt to fix the problem mentioned here: https://github.com/ocurrent/ocaml-ci/issues/607#issuecomment-1379604676.

The solution comes from @cuihtlauac at https://github.com/ocaml/ocaml.org/pull/866.

### DIFF
--- a/web-ui/main.ml
+++ b/web-ui/main.ml
@@ -1,5 +1,17 @@
 module Backend = Controller.Backend
 
+module Middleware = struct
+  let no_trailing_slash next_handler request =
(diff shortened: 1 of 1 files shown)


cuihtlauac:
Let me know if it solves your problem, I'm curious

maiste:
Sure, no problem! On my workstation tests, it seems to fix the problem. Thanks for the solution :+1:

benmandrew (approved):
Works on my end 👍

maiste:
Let's try in production, then! :)

PR #761, opened 2023-02-16
Factorisation for router.ml

### LINKED ISSUE #746: Meta: track refactoring in the UI code.

maiste:
To keep conventions in routes, having a functor and building them is more accessible than having a specific list for each one. This also forces us to have the same function available for GitHub and GitLab without forgetting to write one (as it happened for some API calls). If some functionality is unavailable, there is the `extra_routes` variable to store unorthodox routes.

Finally, the code is more explicit by using a module to store the routes.

This is an attempt to fix #746.


benmandrew (approved):
Looks great to me, thanks for this 👍

PR #762, opened 2023-02-17
Change forge prefix in breadcrumbs to read 'Organisations'

benmandrew:
Currently on the far left of the breadcrumbs we show the forge that the organisation belongs to (`github` or `gitlab`). However, clicking on this link simply takes the user back to the organisations page containing orgs of both forges. Changing the text to read `Organisations` potentially makes this more consistent.

The actual link being the forge prefix is still preserved due to the way that the breadcrumbs links are constructed. Changing the links entirely to remove the forge would be a breaking change. I'm open to other ways of solving this as this isn't perfect.

Screenshot 2023-02-17 at 12 41 54

### DIFF
--- a/web-ui/view/history.ml
+++ b/web-ui/view/history.ml
@@ -157,7 +157,7 @@ module Make (M : Git_forge_intf.Forge) = struct
       | None -> [ (M.prefix, M.prefix); (org, org); (repo, repo) ]
       | Some commit ->
           [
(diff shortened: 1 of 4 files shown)


maiste (approved):
Thanks for the change! It's good to go :+1:.
I agree, the other solution would break too many links.. So let's ship this one.

PR #764, opened 2023-02-20
Cut yaml lines

MisterDA:
for readability


maiste (approved):
Thanks for improving the readability. I can't tell about the deployment but on the content part LGTM :+1:

tmcgilchrist:
Only way to tell is to run it and check.

PR #768, opened 2023-02-27
Revert back to local solver service

maiste:
Deploy on `live-engine` to mitigate the `Auto-cancelling` error.


tmcgilchrist:
Approved.

It would be helpful to include some context to why this is necessary.

maiste:
This PR is necessary as the `solver-service` started the analysis before flagging them as unnecessary. All the  `analysis` just stopped because of that. Moving back to the local version solved the problem.

PR #772, opened 2023-03-06
Switch ocaml-version from git submodule to opam package.

novemberkilo:
Minor gardening update (that came about from updating my environment after a few weeks away). 

I noticed that `ocaml-version` was pinned to a git submodule that was pointing to the `3.6.0` release of `ocaml-version` (which is the latest version on opam). Since we _should_ be depending on the opam package, I'm removing the submodule. This should have no effect on the project.


tmcgilchrist (approved):

MisterDA:
I mostly agree, but it's often (not always) to have the submodules when implementing features that cross ocaml-ci and its dependencies, or use code before it gets published to opam-repository.

tmcgilchrist:
There is an OCaml-version minor release that needs to be included in opam
versions and the docker files. @novemberkilo

Issue #632, opened 2022-11-24
Cached steps should identify as cached


novemberkilo:
Currently we show cached steps to have a queued time of 0s and a run time of 0s which is confusing because the logs clearly show that the step has a start and end time.

We should identify steps that are cached with `Ran for 0s (cached)` or simply `Cached`

novemberkilo:
Closed by #775

PR #775, opened 2023-03-07
Cached steps should identify themselves as such.

novemberkilo:
iiuc we've been working with the principle that a cached step has no time in the queue and also has 0 running time. This seems to be corroborated by [this](https://github.com/ocurrent/ocaml-ci/blob/fe87959f28ddf1bd94abb1577a8b1f56e875b316/web-ui/view/timestamps_durations.ml#L40) (which predates the redesign work). So I have just worked with that logic to implement a simple solution.

WDYT @maiste?

### DIFF
--- a/web-ui/view/build.ml
+++ b/web-ui/view/build.ml
@@ -108,6 +108,11 @@ let title_card ~status ~card_title ~hash_link ~ref_links ~first_created_at
       ])
 
 let step_row ~step_title ~created_at ~queued_for ~ran_for ~status ~step_uri =
+  let cached = queued_for = "0s" && ran_for = "0s" in
+  let queued_for_txt =
+    if cached then "cached" else Fmt.str "%s in queue" queued_for
+  in
+  let ran_for_txt = if cached then "Cached" else Fmt.str "Ran for %s" ran_for in
   let step_row_id = step_title in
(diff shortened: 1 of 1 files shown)


novemberkilo:
This closes #632

web-ui/view/build.ml
```diff
@@ -108,6 +108,11 @@ let title_card ~status ~card_title ~hash_link ~ref_links ~first_created_at
       ])
 
 let step_row ~step_title ~created_at ~queued_for ~ran_for ~status ~step_uri =
+  let cached = queued_for = "0s" && ran_for = "0s" in
+  let queued_for_txt =
+    if cached then "cached" else Fmt.str "%s in queue" queued_for
+  in
+  let ran_for_txt = if cached then "Cached" else Fmt.str "Ran for %s" ran_for in
```
maiste, line 115: Shouldn't the "Cached" and "cached" use the same case?

maiste (approved):
Minor question otherwise, LGTM :+1:
Thanks for the change :)

web-ui/view/build.ml
maiste: You are perfectly right! Closing the conversation: feel free to merge :)

benmandrew (approved):
Looks good to me

PR #779, opened 2023-03-08
Adding tooltips to the repo page. 

### LINKED ISSUE #739: Timestamp being confusing.
From a discussion with @emillon and @samoht, two issues with timestamps were raised:
1.  The naming "Speed" is confusing:  users can't know what it really is.
> As an example, "higher speed" means a decrease in this number
2. The `Total build time` metric should be displayed only if the user requests. Indeed, maintainers need information about the efficiency of their builds, not about how much it "cost" to run their programs.
> as a maintainer, the metric I'm after is how much time I have to wait before having useful info. so it would be the max of all builds. (or P95 of this). the sum of build times is a useful metric for €/energy/CO2 but not the first thing I'm after as a maintainer

> in terms of UX (for someone that wants to test the system), this is quite confusing and not very appealing. In CI, the first thing people care about is how long they have to wait to get their feedback. That should be main metrics we push forward.

novemberkilo:
Following on from [this proposal](https://github.com/ocurrent/ocaml-ci/issues/739#issuecomment-1459115407) this PR implements the changes to the titles of the table on the repo page and introduces tooltips.

Closes #739


benmandrew:
I'm unsure of the HTML and CSS difficulties, but would it be possible to have the tooltip be above (or below) the text so that it's still readable while the tooltip is active? All tooltips I can think of will generally do this as it's nice for the user to be able to see what the tooltip describes while reading it.

Example, but many others exist:
Screenshot 2023-03-08 at 12 51 32

maiste:
I agree with Ben's suggestions. As the position is absolute, moving the tooltip from its origin should be possible.

novemberkilo:
With stylistic improvements:


https://user-images.githubusercontent.com/181086/224584178-ce97ce7b-3197-4818-8397-37c59fe5cae0.mp4

novemberkilo:
I've also taken out "Total build run time" from the UI. We can still report on it via prometheus metrics.

maiste (approved):
LGTM :+1: 

I have just updated the code a bit to display the tooltip on the center and make sure it doesn't wrap. It looks like this: 
image

@novemberkilo WDYT? If you think it's not right, feel free to revert the commit :smile:

Issue #756, opened 2023-02-03
Runners for s390, arm32, x86_32, ppc64 are missing!?


hannesm:
Hello,

first of all thanks for your work and OCaml-CI service. Back some days ago (Sep 13th 2022), at the latest mirage-crypto release (https://github.com/mirage/mirage-crypto/runs/8338362322), OCaml-CI had various non-amd64 runners.

Now, a fresh OCaml-CI run only lists amd64 and arm64 runners. Is that intentional? It would be great to get the other architectures back (esp. s390 due to endianness, and some 32 bit platform for testing 32bit support).

Thanks a lot,

Hannes

dra27:
It's not a final answer, but those architectures appear to be unavailable for OCaml 5.0, and so are being eliminated - e.g. https://ci.ocamllabs.io/github/mirage/mirage-crypto/commit/b9414b37752c325295445f1aca0571cf5916cd22/variant/(analysis)#L59-59

dra27:
The underlying issue is the dependency on `num`, which doesn't support bytecode-only architectures without a new release.

tmcgilchrist:
David is correct the analysis is choosing to try finding a solution for the
latest release of OCaml which is 5.0. This doesn’t support those extra
architectures like s390x and ppc64. Would it make sense to make 4.14 a
special case and include it for the variants as well?

dra27:
It's in the same ball-park as issues like #381. We could have a slightly more principled rule, rather than a special case: the package is _supposed_ to install in OCaml 5 (ocaml-ci is of course opinionated about supporting the _latest_ version of OCaml) but the upper-bound could be relaxed when it's the _dependencies_ which are causing the problem. In other words, instead of testing packages on the _latest_ version of OCaml (by architecture) we instead test by the latest version of OCaml supported by their _dependencies_.

hannesm:
Thanks for your quick reply. I'll wait for a release of `num` that supports OCaml 5.

And thanks for pointing into the log, I would have never figured that out by myself (I can't find any reason for "eliminated all possibilities", and from the huge list of which packages to install I wouldn't have figured out that `num` is the one not bytecode-only compatible).

[...]

PR #780, opened 2023-03-08
Support exotic architectures removed in OCaml 5.0

benmandrew:
Support for s390, arm32, x86_32, and ppc64 have not yet been added to OCaml 5.0, but we attempt to build these combinations anyway and fail, leaving the architectures untested. This PR uses a change in `ocaml-version` (https://github.com/ocurrent/ocaml-version/pull/61) to choose the latest version of OCaml that is supported by a given architecture.

This cannot be merged until the `ocaml-version` change is merged and released.

Fixes https://github.com/ocurrent/ocaml-ci/issues/756.

### DIFF
--- a/service/conf.ml
+++ b/service/conf.ml
@@ -117,7 +117,13 @@ let pool_of_arch = function
   | `Riscv64 -> "linux-riscv64"
 
 let platforms ~include_macos opam_version =
-  let v ?(arch = `X86_64) label distro ocaml_version =
+  let v ?(arch = `X86_64) ?ocaml_version label distro =
+    let ocaml_version =
+      match ocaml_version with
+      | None -> Ocaml_version.Until.arch arch
+      | Some ov -> ov
+    in
+    let ocaml_version = OV.with_just_major_and_minor ocaml_version in
     {
       arch;
       label;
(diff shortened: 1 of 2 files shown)


tmcgilchrist:
@benmandrew re-vendor `ocaml-version` and try this out on ocaml-ci against [https://github.com/mirage/mirage-crypto](https://github.com/mirage/mirage-crypto/)

kit-ty-kate (changes requested):
OCaml-CI should instead test 4.14 and 5.0 as if they were both latest versions, given 4.14 is LTS. This is what opam-repo-ci does for example (see https://github.com/ocurrent/opam-repo-ci/pull/196)

benmandrew:
I've vendored ocaml-version pointing to my own fork.

Basic question, but if the changes are OCaml-CI specific then what's the standard way to integrate a submodule in? Pointing to my fork isn't a permanent solution, but it also seems strange to maintain a separate branch on [`ocurrent/ocaml-version`](https://github.com/ocurrent/ocaml-version) specifically for usage in OCaml-CI.

service/conf.ml
```diff
@@ -52,7 +52,9 @@ end
 module OV = Ocaml_version
 module DD = Dockerfile_opam.Distro
 
-let default_compiler = OV.(Releases.latest |> without_patch)
+let default_compilers =
+  OV.(List.map with_just_major_and_minor Releases.[ v4_14; v5_0 ])
```
kit-ty-kate, line 56:
```suggestion
  OV.(List.map with_just_major_and_minor Releases.[ v4_14; latest ])
```

kit-ty-kate:
I don't understand why this fork is necessary

kit-ty-kate:
> OCaml-CI assumes that all architectures work fully on 5.0+. It tries to build these architectures with the master distro and latest version, which fails at some point leading to those arches not being tested. 

They all do work though, only in bytecode-only mode (which is a valid state of the OCaml compiler).

The issue in https://github.com/ocurrent/ocaml-ci/issues/756 is just that when a package depends on `num` then it won't be tested due to `num` being unavailable in bytecode-only mode. So simply testing both 4.14 and 5.0 fixes this issue and the ocaml-version fork is not needed.

benmandrew:
Ahh I understand what you mean, thanks for the explanation.

kit-ty-kate (approved):

Issue #783, opened 2023-03-13
Historical build chart is incomplete. Name of default branch is wrong.


novemberkilo:
## Context

The historical build chart should include the latest build but does not. This can cause confusion. Also, the default branch name is wrong.

## Current behaviour (examples provided)

![Screenshot 2023-03-13 at 10 58 51 am](https://user-images.githubusercontent.com/181086/224584567-fceea4fb-e0a8-4571-afe4-0110d5243954.png)

This shows a green tick icon to say that the latest build has passed. This is corroborated by the build page. Note also that the default branch is incorrectly `master`

![Screenshot 2023-03-13 at 10 59 08 am](https://user-images.githubusercontent.com/181086/224584563-38b693aa-c01e-4e7b-9634-92e9dc6c7c45.png)

The build history for `main` shows
![Screenshot 2023-03-13 at 10 59 38 am](https://user-images.githubusercontent.com/181086/224584557-8d071057-a779-4cd6-b039-3344f2d66d6f.png)

## Expected behaviour

The build chart is meant to provide an overview of the recent builds of the default branch of a repo. It should include the latest build. The default branch should also be correct.

Reported by @tmcgilchrist

novemberkilo:
It looks like for some repos, the history includes the latest commit, and for some it does not. 

This can be seen by visiting https://ci.ocamllabs.io/github/ocurrent and hovering over the last column of the charts. For example, `ocaml-ci` shows the latest ref as also being the last ref of its chart, but `mirage-ci` does not. This is consistent with the history page of the default branch of the repo. 

Any ideas about what's going on here @maiste @benmandrew?

novemberkilo:
Thanks @benmandrew - that was useful information but after chasing it around, I think the issue here is a bit different. The build history comes from the `ci_build_summary` table of the database and so I spent some time simulating the scenario above (same commit on multiple grefs) to see if I could determine if the information was not being written correctly, or not being read correctly.

My conclusion at this point is that the information is being persisted. The same commit appears against multiple grefs. However, the query being used to read history is not picking up the commit on all the grefs.

I'm making progress and expect to have a fix soon. cc/ @maiste and @tmcgilchrist for information.

PR #784, opened 2023-03-13
Fix default branch in build chart.

novemberkilo:
Addresses #783 

Looks like we've been carrying a hard-coded "master" label for the default branch in the list of repos :) This PR fixes that.


maiste (changes requested):
Thanks for the fix; everything looks great :+1: 
Could you generate the new opam file with the change from the `dune-project`?

benmandrew (approved):
Great, with @maiste's changes it's good to go

PR #786, opened 2023-03-14
Fix build history query.

### LINKED ISSUE #783: Historical build chart is incomplete. Name of default branch is wrong.

novemberkilo:
We were getting mangled history results (disjoint sets of commits) for branches that collect a similar set of commits because we were not including the gref in the clause that selects the maximum build number.

Fixes #783

### DIFF
--- a/lib/index.ml
+++ b/lib/index.ml
@@ -148,8 +148,8 @@ let db =
           ran_for, total_queued_for, message FROM ci_build_summary c1 WHERE \
           owner = ? AND name = ? AND gref = ? AND build_number = (SELECT \
           MAX(build_number) FROM ci_build_summary c2 WHERE c1.owner = c2.owner \
-          AND c1.name = c2.name AND c1.hash = c2.hash) ORDER BY created_at \
-          DESC"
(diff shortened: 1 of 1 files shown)


novemberkilo:
I tested this manually by rebasing a couple of branches on my fork of `ocaml-ci` and letting my test ocaml-ci app process them so that I could collect data in my local sqlite db. This fix appears to make the histories align -- the shared commits appear on the histories of all the branches as expected.

tmcgilchrist (approved):

PR #787, opened 2023-03-16
Update to tailwindcss 3.2.7

MisterDA:
Switched to downloading the tar archive instead of cloning opam-tailwindcss we don't need the history, it's slightly lighter.


MisterDA:
oh no, only git pins are supported :/

maiste:
There seems to be an issue with pinning using only http for the analysis...

maiste:
Indeed, that's sad it was a good idea :/

PR #788, opened 2023-03-16
Remove intermediate line breaks from in-progress logs

benmandrew:
Fixes the issue raised in https://github.com/ocurrent/ocaml-ci/issues/607#issuecomment-1446745199 where extra line breaks appear in the logs. This happens because when data is streamed in, it has a newline added to it so that there is a bit of space at the bottom of the logs; when new data comes in that padding remains.

One potential issue is that there is now no longer much padding at the bottom of the logs (but seemingly only for in-progress builds...). I spent too long trying to figure out how the CSS works so I've just left it as-is.

Screenshot 2023-03-16 at 15 32 43


novemberkilo:
🎉 thanks @benmandrew

PR #790, opened 2023-03-17
Tests for platforms

benmandrew:
Add tests so that we're more likely to notice an accidentally changed set of supported architectures. Currently there is no such check, and both increasing and decreasing the number of tested platforms can have negative impacts if not noticed.


MisterDA (approved):
looks great, thanks!

PR #791, opened 2023-03-17
Allow using tar.gz archives from GitHub in opam pin-depends

MisterDA:
Today's hack!
Sometimes it's nice not to have to pin a git repository: for instance with opam-tailwindcss opam has to download the entire git history, which includes all the executables committed. I thought I'd use the tarball generated by `git archive` instead. We get the snapshot from a commit, so there's still reproducibility and the possibility to bisect changes made to the opam file, with the bonus of very few metadata and unwanted data from the dependency to download.

For GitHub hosted projects, the code involves downloading a tarball and following a redirection, then extracting the opam files from a compressed tarball. The downloaded tarball is cached. Decompression is done on the fly, without decompressing on disk.

I've tried to make the download and iteration over the data efficient, there might be better ways. There's no cancellation mechanism for now, perhaps the job should be checked periodically for cancellation in the loops?

There's a bug in `tar` preventing reading the data: https://github.com/mirage/ocaml-tar/issues/113, and it's now fixed!

### DIFF
--- a/lib/pin_depends.ml
+++ b/lib/pin_depends.ml
@@ -29,6 +29,12 @@ module Cmd = struct
     | _ -> false
     | exception Unix.Unix_error (Unix.ENOENT, _, _) -> false
 
+  let file_exists path =
+    match Unix.lstat (Fpath.to_string path) with
+    | { Unix.st_kind = S_REG; _ } -> true
+    | _ -> false
+    | exception Unix.Unix_error (Unix.ENOENT, _, _) -> false
+
   let git ~cancellable ~job ?cwd args =
     let args =
       match cwd with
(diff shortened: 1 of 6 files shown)


maiste:
I'm wondering the performance gain over the complexity this PR introduces...

MisterDA:
It's possible to speed up searching the tar by using `lseek` instead of reading the whole file.
https://gitlab.com/tezos/tezos/-/blob/6cf735462b2e45ce48672f7296177873d698060e/src/lib_store/unix/snapshots.ml#L1365

lib/pin_depends.ml
```diff
@@ -106,3 +108,124 @@ let get_opam ~job ~pkg url =
...
+
+let get_opam ~job ~pkg url =
+  match url.OpamUrl.backend with
+  | `git -> get_opam_git ~job ~pkg url
+  | `http -> get_opam_http ~job ~pkg url
+  | _ ->
+      Fmt.failwith
+        "Only Git or GitHub https pin-depends are supported (got %S for %s)"
+        (OpamUrl.to_string url)
+        (OpamPackage.to_string pkg)
```
tmcgilchrist, line 231: Possible to add a test around this functionality?

tmcgilchrist (approved):

lib/pin_depends.ml
MisterDA: Well if we switch ocaml-ci to pin-depend on the tailwindcss tarball, ocaml-ci's own deployment will test the feature!

PR #794, opened 2023-03-22
Logs over websockets

novemberkilo:
This PR changes a few things.

1. It introduces the machinery for `jsoo` into the project
2. A new route is introduced that provides a websocket connection for log data
3. The logic for constructing the appropriate markup for streaming log data is moved over to javascript by way of jsoo
4. A websockets client is also introduced in jsoo to consume the log data and present it within the `step` page
5. Following https://css-tricks.com/books/greatest-css-tricks/pin-scrolling-to-bottom/ once we scroll to the bottom of the page, the scrolling stays pinned there and there is no need to scroll further as logs continue to appear

Along the way I found a bug in `Brr` that has resulted in us having to hand roll percent-encoding of some things (so that the parentheses that show up in variants don't get lost in the routing). I will file an issue on `Brr` accordingly.


MisterDA (commented):
Looking great! That will definitively be an improvement.

web-ui/jsoo/main.ml
```diff
@@ -0,0 +1,80 @@
...
+  let s_lp_encoded =
+    Re.Str.global_replace regexp_left_paren encoded_left_paren s'
+  in
+  let s_rp_encoded =
+    Re.Str.global_replace regexp_right_paren encoded_right_paren s_lp_encoded
+  in
+  Jstr.of_string s_rp_encoded
+
+(* let js_str = Js.string *)
+let document = Dom_html.document
```
patricoferris, line 26: It might be worthwhile going _all in_ on `Brr` so you can drop the dependency on `js_of_ocaml` and `js_of_ocaml-ppx`? In which case this could turn into something like:

```ocaml
let inject_log_lines data =
  match Document.find_el_by_id G.document scroller_id with
  | None -> assert false
  | Some scroller ->
    let new_div = El.div [ El.txt' data ] in
    El.append_children scroller [ new_div ]
```

web-ui/jsoo/main.ml
```diff
@@ -0,0 +1,80 @@
...
+        let data = Brr_io.Message.Ev.data (Ev.as_type e) in
+        inject_log_lines
+        @@ Js.string
+        @@ Process_chunk.go line_number (Js.to_string data))
+      target
+  in
+  ignore result;
+  Js._true
+
+let _ = Dom_html.window##.onload := Dom_html.handler go
```
patricoferris, line 80: Similar to my other comment, with Brr this could be:

```ocaml
let () =
  ignore (Ev.listen Ev.load go (Window.as_target G.window) : Ev.listener)
```

patricoferris (commented):
A few comments on the Brr code :))

web-ui/jsoo/main.ml
```diff
@@ -0,0 +1,80 @@
...
+  let s_rp_encoded =
+    Re.Str.global_replace regexp_right_paren encoded_right_paren s_lp_encoded
+  in
+  Jstr.of_string s_rp_encoded
+
+(* let js_str = Js.string *)
+let document = Dom_html.document
+
+let inject_log_lines data =
+  let new_div = Dom_html.createDiv document in
```
benmandrew, line 29: At least on Safari, adding this extra `div` around the logs creates ugly spacing between the log lines. I'm unsure why.

Screenshot 2023-03-22 at 11 49 44

benmandrew, line 29: At least on Safari, adding this extra div around the logs creates ugly spacing between the log lines. I'm unsure why.

Screenshot 2023-03-22 at 11 49 44

benmandrew (changes requested):
Thanks for this! One note is that on Safari the `overflow-anchor` property is not supported. It has no effect so it's good to keep it in for the other browsers.

web-ui/jsoo/main.ml
novemberkilo: Thanks @patricoferris I had intended to get us to just using Brr but since I am still quite unfamiliar with the API, had deferred completing that. This will help a lot 🙂

tmcgilchrist: Just using `Brr` looks like a good option for making this code simpler.

web-ui/router.ml
```diff
@@ -308,6 +321,14 @@ let build_github_route github =
...
             in
             Dream.redirect request target);
+        (* Dream.get "/ws/github/:org/:repo/commit/:hash/variant/:variant"
+           (fun request ->
+             Ws_controller.Github.show_step
+               ~org:(Dream.param request "org")
+               ~repo:(Dream.param request "repo")
+               ~hash:(Dream.param request "hash")
+               ~variant:(Dream.param request "variant")
+               request backend); *)
```
maiste, line 331: This can be removed as you move it to the generic router :)

web-ui/static/js/step-page-websocket-client.js
maiste, line 1: Don't we want this to be built with `jsoo`?

web-ui/view/step.mli
```diff
@@ -32,5 +32,6 @@ module Make : functor (_ : Git_forge_intf.Forge) -> sig
     can_cancel:bool ->
     ?flash_messages:(string * string) list ->
     string * int64 ->
-    Dream.response Lwt.t
+    string
+    (* Dream.response Lwt.t *)
```
maiste, line 36: Why does this have to be a string and not a `Dream.response`?

web-ui/static/js/step-page-websocket-client.js
novemberkilo: yes that's cruft now and will be deleted. sorry forgot to do that prior to last commit.

maiste: No problem! :smile: This is not urgent and this is a massive piece of work already :wink:

web-ui/jsoo/main.ml
```diff
@@ -0,0 +1,80 @@
...
+  let result =
+    Ev.listen Brr_io.Message.Ev.message
+      (fun e ->
+        let data = Brr_io.Message.Ev.data (Ev.as_type e) in
+        inject_log_lines
+        @@ Js.string
+        @@ Process_chunk.go line_number (Js.to_string data))
+      target
+  in
+  ignore result;
```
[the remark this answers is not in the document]
patricoferris: Agreed! I'm also quite partial to adding type annotations to ignored values (here ` let _result : Ev.listener =`). Too many times have I ignored a partially applied function ^^"

web-ui/jsoo/main.ml
novemberkilo: Wondering whether this has to do with the padding that is added to the first and last child ... might try taking that out and solving for neatness differently (like adding a margin on the container).

novemberkilo: I think I have fixed this now @benmandrew - would you please run up my branch locally and verify that it renders well for you in Safari?

web-ui/jsoo/dune
```diff
@@ -0,0 +1,15 @@
...
+  (pps js_of_ocaml-ppx))
+ (libraries ansi astring brr))
+
+; (executable
+;  (modules steps_to_reproduce_build)
+;  (name steps_to_reproduce_build)
+;  (modes js)
+;  (preprocess (pps js_of_ocaml-ppx))
+;  (libraries
+;   brr))
```
tmcgilchrist, line 15: Remove this commented out chunk?

web-ui/jsoo/main.ml
```diff
@@ -0,0 +1,79 @@
...
+let regexp_left_paren = Re.Str.regexp_string "("
+let regexp_right_paren = Re.Str.regexp_string ")"
+let encoded_left_paren = "%28"
+let encoded_right_paren = "%29"
+
+let encode_parens s : Jstr.t =
+  let s' = Jstr.to_string s in
+  let s_lp_encoded =
+    Re.Str.global_replace regexp_left_paren encoded_left_paren s'
+  in
```
tmcgilchrist, line 16:
```suggestion
  let s_lp_encoded = 
    Jstr.to_string s 
    |> Re.Str.global_replace regexp_left_paren encoded_left_paren
  in
```
Consider avoiding the `s'` variable using `|>`?

web-ui/jsoo/main.ml
```diff
@@ -0,0 +1,79 @@
...
+        [
+          Jstr.of_string "ws:/";
+          Jstr.concat ~sep:(Jstr.of_string ":") [ hostname; Jstr.of_int port ];
+          encode_parens @@ Jstr.append (Jstr.of_string "ws") pathname;
+        ]
+
+(* It looks like parens are not being percent encoded - will file an issue with Brr. Then the code below should work
+   Brr.Uri.with_uri ~scheme:(Jstr.of_string "ws")
+   ~path:(encode_parens @@ Jstr.append (Jstr.of_string "ws") pathname)
+   location *)
```
tmcgilchrist, line 56: Could you include a link to the Brr issue here?

web-ui/view/git_forge_intf.ml
```diff
@@ -97,7 +97,8 @@ module type View = sig
     can_cancel:bool ->
     ?flash_messages:(string * string) list ->
     string * int64 ->
-    Dream.response Lwt.t
+    string
+  (* Dream.response Lwt.t *)
```
tmcgilchrist, line 101:
```suggestion
```

web-ui/view/step.ml
```diff
@@ -193,460 +193,359 @@ module Make (M : Git_forge_intf.Forge) = struct
...
-                a_class
-                  [
-                    "shadow-sm rounded-lg overflow-hidden border \
-                     border-gray-200 dark:border-gray-400 divide-x \
-                     divide-gray-20";
-                  ];
-                a_style "display: none";
-                a_id "build-repro-container";
-              ]
+    (* let header, footer = *)
```
tmcgilchrist, line 200:
```suggestion
```

tmcgilchrist (approved):
Looks good. 

A few tidy up comments required.

PR #795, opened 2023-03-24
Moving Opam-specific spec file functionality to OBuilder (WIP)

benmandrew:
Depends on this OBuilder PR: https://github.com/ocurrent/obuilder/pull/146.

Addresses https://github.com/ocurrent/ocaml-ci/issues/682.


shonfeder:
Between the conflicts of this WIP and the upstream PR it depends on being closed, I don't think we'll be able to move forward with this work as is. We'll figure out how to prioritize work on the motivating issue tho, as I think this is an important abstraction to address.

PR #797, opened 2023-03-30
Fix install command in readme

gridbugs:
Fixes two things:
 - ocaml-version is not a local package, so remove the leading "./"
 - the ./solver-service local package was missing from the command

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Get the code with:
 ```sh
 git clone --recursive https://github.com/ocurrent/ocaml-ci.git
 cd ocaml-ci
-opam install --deps-only ocaml-version ./ocaml-dockerfile ./ocluster ./ocurrent ./solver-service .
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,7 +38,7 @@ Get the code with:
 ```sh
 git clone --recursive https://github.com/ocurrent/ocaml-ci.git
 cd ocaml-ci
-opam install --deps-only ./ocaml-version ./ocaml-dockerfile ./ocluster ./ocurrent .
+opam install --deps-only ocaml-version ./ocaml-dockerfile ./ocluster ./ocurrent ./solver-service .
```
benmandrew, line 41:
```suggestion
opam install --deps-only ./ocaml-dockerfile ./ocluster ./ocurrent ./solver-service .
```

benmandrew (approved):
Thanks for this! You can remove the explicit `ocaml-version` as the opam solver finds the dependency itself.

PR #799, opened 2023-04-03
Add step-logs.js to crunch.

novemberkilo:


web-ui/dune
```diff
@@ -2,6 +2,7 @@
  (target static.ml)
  (deps
   %{workspace_root}/web-ui/static/css/main.css
+  %{workspace_root}/web-ui/static/js/step-logs.js
```
novemberkilo, line 5: This was missing and was causing a 404 on `step-logs.js` when I deployed this last week.

web-ui/static/js/step-page-poll.js
```diff
@@ -1,3 +1,32 @@
+// Called via jsoo in the generated step-logs.js
+function extractStepsToReproduce(startIndex, finishIndex) {
+  if ( // return if already done
+    !!document.getElementById("build-repro-container") &&
```
MisterDA, line 4: not sure if the double bang is necessary, isn't the value already truthy/falsy?

web-ui/view/step.ml
```diff
@@ -193,460 +193,358 @@ module Make (M : Git_forge_intf.Forge) = struct
...
+        Common.form_cancel_step ~variant ~csrf_token ~show:can_cancel ();
+        Common.form_rebuild_step ~variant ~csrf_token ~show:show_rebuild ();
+      ]
+    in
+    let branch =
+      if refs = [] then ""
+      else
+        match Astring.String.cuts ~sep:"/" (List.hd refs) with
+        | "refs" :: "heads" :: branch -> Astring.String.concat ~sep:"/" branch
+        | _ -> ""
```
MisterDA, line 211: Maybe check that `"refs/heads/" is a prefix of `List.hd refs` instead.

MisterDA (commented):
Looks nice, thanks!

web-ui/static/js/step-page-poll.js
novemberkilo: The double bang is "idiomatic" js for `Boolean(object)`

web-ui/jsoo/steps_to_reproduce_build.ml
```diff
@@ -0,0 +1,15 @@
+module Ev = Brr.Ev
+module El = Brr.El
+module At = Brr.At
+module Document = Brr.Document
+module Window = Brr.Window
```
maiste, line 5:
```suggestion
```
Can't this be remove?

ocaml-ci-web.opam
maiste, line 1: Shouldn't we fix the version of `dream` to ensure we have no broken change here?

tmcgilchrist: Yes we should have the latest alpha5 as the lower bound on Dream.

dune-project
```diff
@@ -115,7 +115,7 @@
   (ansi (>= 0.6.0))
   (capnp-rpc-unix (>= 1.2))
   (crunch (and (>= 3.2.0) :build))
-  dream
+  (dream (>= 1.0.0alpha5))
```
tmcgilchrist, line 118:
```suggestion
  (dream (>= 1.0.0~alpha5 ))
```

tmcgilchrist (approved):

PR #800, opened 2023-04-03
Add opam-repository commit sha to logs.

novemberkilo:
When a solve fails it's useful to know what commit of `opam-repository` was used. This can help detect when something has gone wrong with the webhook driven update machinery - specifically, if we've stopped processing input webhooks from opam-repository.


novemberkilo:
See also https://github.com/ocurrent/ocurrent/pull/419

tmcgilchrist (approved):

PR #808, opened 2023-04-14
Test OCaml 5.1

kit-ty-kate:


tmcgilchrist:
We can include this once there is an official 5.1 release. Unless someone has time to mark it as an experimental build.  There are docker base images for Linux but we will need manual intervention to update the macOS workers. cc @mtelvers

benmandrew:
I'm unsure if I have permissions to commit to another's PR (and got a bit carried away with other changes) so I have a new PR #813 where 5.1 is marked as experimental.

The MacOS workers already have 5.1~alpha1 (and 5.2!) images thanks to @mtelvers.

benmandrew:
#813 is merged, thanks.

PR #814, opened 2023-05-17
Update vendored dependencies

moyodiallo:
Fix the issue that leads to #768, about "Auto-cancelling" error.

### DIFF
--- a/lib/backend_solver.ml
+++ b/lib/backend_solver.ml
@@ -4,7 +4,7 @@ let ( >>!= ) = Lwt_result.bind
 
 type t =
   | Remote of Current_ocluster.Connection.t
-  | Local of Ocaml_ci_api.Solver.t Lwt.t
+  | Local of Solver_worker.Solver_request.t Lwt.t
 
 let switch = Current.Switch.create ~label:"solver-remote" ()
 let config = Current.Config.v ()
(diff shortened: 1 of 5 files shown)


benmandrew (approved):
Works on my end 👍

maiste (approved):
Work fine on `live-engine` it can be merged.

lib/backend_solver.ml
```diff
@@ -34,14 +34,15 @@ let remote_solve con job request =
   Capnp_rpc_lwt.Capability.with_ref build_job
     (Current_ocluster.Connection.run_job ~job)
 
-let local ?solver_dir () : t =
-  Local
-    (Lwt.return (Solver_worker.spawn_local ?solver_dir ~internal_workers:20 ()))
+let local () : t =
+  Local (Lwt.return (Solver_worker.Solver_request.create ~n_workers:20 ()))
```
talex5, line 38: Note that this is quite a significant change: `spawn_local` created a child process once at startup, which then spawned solver worker processes as needed. Whereas `Solver_request.create` runs the manager in the ocaml-ci process, spawning child processes when the opam-repository commit changes.

The reason for the previous (more complicated) design was to avoid accidentally inheriting file descriptors that either ocaml-ci or the solver forgot to mark as close-on-exec. However, I guess it's been working OK so maybe it's not a problem

PR #815, opened 2023-05-24
Signpost experimental variants in web UI

benmandrew:
Failures of experimental variants can sometimes be due to bugs in OCaml-CI. This should be signposted to users who would otherwise think that there is a definite issue with their project.

I'm open to rewording of the explanatory text at the bottom of the page, which is currently:
> *Variants labelled `(experimental)` are still undergoing testing; if they have failed it may be a bug in OCaml-CI.

---

Screenshot 2023-05-24 at 11 24 36

### DIFF
--- a/service/api_impl.ml
+++ b/service/api_impl.ml
@@ -2,6 +2,7 @@ module Rpc = Current_rpc.Impl (Current)
 module Raw = Ocaml_ci_api.Raw
 module String_map = Map.Make (String)
 module Index = Ocaml_ci.Index
+module Pipeline = Ocaml_ci.Pipeline
 module Run_time = Ocaml_ci.Run_time
 open Capnp_rpc_lwt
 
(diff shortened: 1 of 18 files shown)


service/api_impl.ml
```diff
@@ -2,6 +2,7 @@ module Rpc = Current_rpc.Impl (Current)
 module Raw = Ocaml_ci_api.Raw
 module String_map = Map.Make (String)
 module Index = Ocaml_ci.Index
+module Pipeline = Ocaml_ci.Pipeline
```
maiste, line 5: Shouldn't this type be extracted in another module to avoid importing the pipeline? It doesn't seem that bad to have an `Experimental` module.

web-ui/static/js/build-page-poll.js
```diff
@@ -72,6 +72,7 @@ function poll(api_path, timeout, interval) {
           const hyphen = document.createElement("div");
           const step_queued_for = document.createElement("div");
           const right_matter = document.createElement("div"); // ran_for, right_arrow
+          const is_experimental = document.createElement("div"); // TODO
```
maiste, line 75: Maybe this TODO should be removed :wink:

PR #818, opened 2023-05-30
Update distributions list

MisterDA:


tmcgilchrist:
Could you enumerate the distribution additions and removals for visibility
here?

MisterDA:
- added: Ubuntu 23.04, Fedora 38, Alpine 3.18, OpenSUSE Tumbleweed
- removed: Fedora 36

tmcgilchrist:
@MisterDA Seems like the `capnp` packages are not available on Tumbleweed and the builds are failing, could you check whether they truely are unavailable or we need to patch the `conf-capnproto` package in opam-repository.

PR #821, opened 2023-06-01
Revert "Update distributions list (#818)"

novemberkilo:
This reverts commit 9408e5f35cbcaad2825777f05772a9ad763e7c88.

Reverting so that we can fix the failures from #818 and retry in a separate PR.


tmcgilchrist (approved):
Looks good, we can restore it when we fix the capnp error.

novemberkilo:
FYI https://github.com/ocaml/opam-repository/pull/23852

PR #823, opened 2023-06-08
Add reynir in allowlist

reynir:


benmandrew:
Could you also add your username to the `deploy-data/github-organisations.txt` file? Thanks!

PR #828, opened 2023-06-19
Update macOS to 5.1~alpha2

mtelvers:


benmandrew (approved):
Thanks for this

samoht:
Is there a way to make sure this is always in sync?

tmcgilchrist:
@samoht tracking issue for extracting this code into a common library https://github.com/ocurrent/ocaml-ci/issues/682. We need to work with obuilder-specs rather than `ocaml-dockerfile` for these new platforms (macOS and FreeBSD) that don't support docker natively.

PR #830, opened 2023-06-20
Update dune-project file with maintainer information

tmcgilchrist:
Updates the supported ocaml version to 4.14, since that is the version used in production there is no point in supporting earlier versions and doing builds for them.


benmandrew (approved):

novemberkilo (approved):
👍🏾 thanks @tmcgilchrist

moyodiallo (commented):
:+1:

PR #835, opened 2023-07-03
Extend experimental to OCaml 5.1~alpha2

tmcgilchrist:


benmandrew (approved):
LGTM. The 5.1~alpha2 builds already show up as experimental because they're all `macos-homebrew`, but this is good for reliability.

PR #836, opened 2023-07-05
Update macos 5.1 builds to beta1.

tmcgilchrist:
Requires updating the base images on macOS first.


benmandrew (approved):

benmandrew:
Could you also add 5.1~beta1 to the list of experimental variants? As in https://github.com/ocurrent/ocaml-ci/pull/835.

PR #841, opened 2023-07-07
Lint on linux (the sequel)

benmandrew:
Prefer running `(lint-fmt)` on Linux x86 platforms. Completion of #839.


tmcgilchrist:
Deployed on `live-engine` 
Thanks @benmandrew

tmcgilchrist:
This is not quite right it was still choosing macOS to run the linting steps on.

PR #853, opened 2023-08-09
Linting without installing dependencies of a project

moyodiallo:
[opam-dune-lint](https://github.com/ocurrent/opam-dune-lint/pull/46) doesn't build a project when linting, so installing the dependencies isn't necessary.

_To note this is on experimentation._


moyodiallo:
Works fine in prod, `live-engine branch`.

tmcgilchrist:
The other piece for general ocaml-ci linting is to run both opam lint on all the opam files and opam-dune-lint. From my limited local testing they check different things and we should be running both.

PR #854, opened 2023-08-17
Add experimental builds for Opam 2 2

tmcgilchrist:
Adds experimental builds for opam.2.2 using the pre-built opam-dev (master) builds in the docker base images. 
Tagged as a draft because:
1. Some variants like flambda don't work with the opam list command to get opam-vars
2. opam-vars query is hardcoded for macOS and FreeBSD, when it should be run on those cluster builds.

### DIFF
--- a/lib/build_info.ml
+++ b/lib/build_info.ml
@@ -21,15 +21,17 @@ let experimental_variant s =
     match s.variant with
     | None -> false
     | Some v ->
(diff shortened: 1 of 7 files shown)


shonfeder:
Should be superseded by #953 iiuc.

tmcgilchrist:
Looking forward to opam 2.2 support 👍🏻

PR #868, opened 2023-09-11
Add edwintorok as ocaml-ci user

edwintorok:
Please add me to ocaml-ci.

I was looking for a way to run something similar to what 'opam-ci' does (so I can discover and fix the bugs without having to do a release for each step), and I think this would allow me to do that. If I should be looking elsewhere instead then please let me know, thanks.


tmcgilchrist:
Please follow the [Getting Started guide](https://ocaml.ci.dev/getting-started) and please let us know if you have any issues.

edwintorok:
Thanks, looks like it has built most things successfully for this repo, but there are some jobs that show an 'error 500':
https://ocaml.ci.dev/github/edwintorok/lintcstubs-arity/commit/1aa77defdfffe7ae0087666fb0e9091485559bab

For example this one, although the dashboard shows 'in queue', so maybe just a UI bug?
https://ocaml.ci.dev/github/edwintorok/lintcstubs-arity/commit/1aa77defdfffe7ae0087666fb0e9091485559bab/variant/ubuntu-22.10-4.14_opam-2.1

mtelvers:
In this case, these stuck tests are for Ubuntu 22.10, which is now end-of-life.  It has been removed from OCaml-CI this morning.  https://github.com/ocurrent/ocaml-ci/pull/870.  We should improve the error message.

PR #878, opened 2023-09-20
Ensure linting specs are using upper-bound selections

benmandrew:
The `(lint-doc)` spec on [ocurrent/clarke](https://github.com/ocurrent/clarke) was failing due to it using a lower-bound selection reserved for the `(lower-bound)`  spec. This PR filters out lower-bound selections from the list of candidates.


moyodiallo (approved):
LGTM 👍

PR #883, opened 2023-09-28
Add `PLATFORMS` environment variable to support local Docker stack

benmandrew:
Currently the `CI_PROFILE` env controls both where capnp secrets are located (relative path for local testing or absolute path for Docker volume), and what platforms are chosen to test on (a minimal subset, or all of them).

I am working on a `docker-compose.yml` for testing an entire OCaml-CI stack locally, where `CI_PROFILE=production` is needed. However, for this sort of local testing a minimal subset of platforms is preferred. This PR separates the platforms behaviour of `CI_PROFILE` into `PLATFORMS` so that they can be set independently.


benmandrew:
Renamed `CI_PROFILE` to `CAPNP_PROFILE` as that is all that the environment variable is now responsible for.

PR #886, opened 2023-10-02
Exclude macos and freebsd builds when running ocaml-ci-local.

tmcgilchrist:
ocaml-ci-local pipelines assume you are building using docker which only support Linux.

### DIFF
--- a/service/conf.ml
+++ b/service/conf.ml
@@ -68,7 +68,8 @@ end
 module OV = Ocaml_version
 module DD = Dockerfile_opam.Distro
 
-(** Default supported compilers. Defined as 4.14 Long Term Support and latest 5.* release. *)
+(** Default supported compilers. Defined as 4.14 Long Term Support and latest
+    5.* release. *)
 let default_compilers =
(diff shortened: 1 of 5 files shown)


benmandrew (approved):

service/conf.ml
```diff
@@ -229,10 +205,11 @@ let platforms ~profile ~include_macos opam_version =
       let[@warning "-8"] (latest :: previous :: _) =
         List.rev OV.Releases.recent
       in
-      let macos_distros = if include_macos then macos_distros else [] in
       let ovs = [ latest; previous ] in
       let releases = List.map make_release ovs in
-      releases @ macos_distros @ freebsd_distros
+      releases
+      |> List.append (if include_macos then macos_distros else [])
+      |> List.append (if include_freebsd then freebsd_distros else [])
```
benmandrew, line 212: As an addition to the PR could you remove these two lines? To stop including Macos and FreeBSD distros on the 'minimal' set of platforms (used for local testing of `ocaml-ci-service`).

PR #888, opened 2023-10-05
Fixing the issue of the CI not responding

moyodiallo:
With @mtelvers  we identify where the issue about the CI not responding when there's a lot of activity, meaning when there's a new push in opam-repo the analysis of all user repositroy is restarted.

An analysis makes a tmp copy of an user repository in order to extract all the information needed for the solver-service. There's some IO access involved and take time when there's a lot of analysis running at same time in the CI.

This is an attempt to fix that, by cache the extraction without keep doing it for each new opam repository commit.


benmandrew (approved):
I don't have the background to tell if this solves the problem, but it at least looks sensible and works on my machine 👍

mtelvers:
Some more context here.  Looking at a run such as [run 51](https://ocaml.ci.dev:8100/job/2023-10-05/162038-ci-analyse-d69a3b) nearly an hour is lost between the start and the `cp` line.  At this time, OCaml-CI is cloning the repo from GitHub.  At other times, such as [run 52](https://ocaml.ci.dev:8100/job/2023-10-05/214231-ci-analyse-e32c9f) or [run 50](https://ocaml.ci.dev:8100/job/2023-10-05/114627-ci-analyse-c3ce4e) the `git clone` is very fast.  It's not clear why there is such a difference, but it could be down to rate limiting from GitHub.  This PR should cache the result of the clone to avoid repeating it with the same commit.  I suggest we push it to `live-engine` and see how it performs in the real world.

moyodiallo:
>  I suggest we push it to live-engine and see how it performs in the real world.

I pushed it on `live-engine`. It is deploying right now.

mtelvers:
This appears to be working.  The `Fetch` and `Extract` stages are not repeated but `Analyse` is run as expected.

image

benmandrew:
@moyodiallo can this be merged?

moyodiallo:
> @moyodiallo can this be merged?

Yes.

PR #891, opened 2023-10-09
Add smuenzel to stack.yml

smuenzel:
PS: maybe this allowlist should be sourced from a separate file, the line is getting a little long.


benmandrew:
Could you also add your username to `deploy-data/github-organisations.txt`?

That's a good suggestion. We currently have GitHub usernames duplicated over `stack.yml` and the `txt` file for historical reasons, but it'd be good to smooth out the onboarding process!

benmandrew:
@smuenzel the change is deployed, please let us know if there's any issues. Thanks!

PR #892, opened 2023-10-09
Have GitHub organisations list in one place

benmandrew:
It is becoming unwieldy to have the GitHub orgs stored in two different places (`stack.yml` and `deploy-data/github-organisations.txt`) as well as making it more of a pain for new users to onboard. More often than not, they only put their username in one place.

This PR makes it so `stack.yml` is generated from `stack.yml.in`, with the orgs substituted in from `github-organisations.txt`. This is done when `make deploy-stack` is run.

`github-organisations.txt` is now the unique source of GitHub orgs, and is where new users should add their usernames. I will update the documentation to reflect this.

Currently, the new script substitutes the org list into `stack.yml.in` with a `sed` call, replacing the placeholder string `GITHUB_ORGANISATIONS`. I'm open to suggestion for more robust or idiomatic ways of getting the org list into `stack.yml`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,5 +9,10 @@ deploy-backend:
 deploy-web:
 	env DOCKER_BUILDKIT=1 docker --context $(CONTEXT) build -f Dockerfile.web -t ocaml-ci-web .
 
-deploy-stack:
+orgs := $(shell cat deploy-data/github-organisations.txt | tr '\n' ',')
+
+stack.yml: stack.yml.in deploy-data/github-organisations.txt
+	sed 's/GITHUB_ORGANISATIONS/${orgs:,=}/' stack.yml.in > stack.yml
+
+deploy-stack: stack.yml
 	docker --context $(CONTEXT) stack deploy --prune -c stack.yml ocaml-ci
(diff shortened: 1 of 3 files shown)


mtelvers:
Two alternative options:

1) Update OCaml-CI to change the parameter `--github-account-allowlist` to accept a filename rather than a list of accounts.  The file `deploy-data/github-organisations.txt` could be included in the Docker image and used as the input file.  Perhaps as `--github-account-allowlist-file github-organisations.txt`.  Within OCaml-CI, this could easily be parsed into the account list.  The slight negative is that a change in accounts necessitates a rebuild of the code, but on the positive, Ocurrent Deployer will push the new image out, meaning that no additional access is required to update the account list beyond the ability to commit to the `live` branch.

2) An account list branch could be an input to the pipeline.  We could monitor an `accounts` branch of OCaml-CI repo, and on a commit to that branch, we could pull the latest list of accounts as the `github-organisations.txt` and update them in OCaml-CI.  This seems harder than option 1, but it doesn't need a rebuild of OCaml-CI just to update the list of accounts.  This seems the most *Ocurrent* way to handle it.

** Either option means we could remove `stack.yml` from this repository and include it in the Ansible playbook in the infrastructure repo.

mtelvers (changes requested):
Since we have a Makefile already we can use it to generate stack.yml if it doesn't exist and update it automatically when the dependencies change.

mtelvers (approved):
LGTM

PR #898, opened 2023-10-23
Add metrics for per-variant states

benmandrew:
This metric makes a DB query on every scrape, I don't know whether this will be too expensive.

Screenshot 2023-10-23 at 18 01 54

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@
 _build
 _opam
 var
-var_backup
 *.cap
 .*.swp
 .vscode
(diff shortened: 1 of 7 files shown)


benmandrew:
Even after optimisation of the initial DB query it is still too expensive due to the size of the database — it did not finish in 30 minutes in production. This might be able to be ameliorated by removing stale data from the `cache` OCurrent table and the `ci_build_index` table.

tmcgilchrist:
I think it is reasonable to prune the size of the database to bring it down
to a more reasonable size. Something along the lines of last 6 months and
the last n builds for branches.

Even with that I expect this query to continue to be too expensive. Having
it as a debug tool for adhoc use would be helpful.

PR #899, opened 2023-10-23
Revert names

mtelvers:
If we deploy current `stack.yml` with the changes implemented in #884, the services are renamed which prevents Caddy from finding them.

Caddy (under Docker) uses the service names, `ci` and `gitlab`, as DNS names.  Docker resolves these internally to the IP address of the container.  If we need to rename these, we need to update the corresponding `Caddyfile`.

This PR reverts the name changes in #884 while we discuss the way forward.

### DIFF
--- a/stack.yml
+++ b/stack.yml
@@ -24,7 +24,7 @@ secrets:
     external: true
 
 services:
-  service-github:
+  ci:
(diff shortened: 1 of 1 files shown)


benmandrew (commented):
My mistake, this looks good to me.

PR #901, opened 2023-11-10
Update opam variables for macOS Sonoma

mtelvers:


lib/platform.ml
```diff
@@ -300,9 +300,13 @@ let get_macos ~arch ~label ~builder ~pool ~distro ~ocaml_version ~opam_version
...
              os_distribution = "homebrew";
-             os_version = "13.4";
+             os_version = "14.1";
              ocaml_package = "ocaml-base-compiler";
-             ocaml_version = Fmt.str "%a" Ocaml_version.pp ocaml_version;
+             ocaml_version =
+               (match Ocaml_version.major ocaml_version with
+               | 4 -> "4.14.1"
+               | 5 -> "5.1.0"
+               | _ -> "5.2.0");
```
tmcgilchrist, line 309: Why the change here to not use `Ocaml_version.pp`? Based on the list here https://github.com/ocurrent/ocaml-ci/blob/master/service/conf.ml#L73 we should only see 4.14 and 5.1 versions in this code.

mtelvers: The value entered here goes directly to the solver.  Given `let ocaml_version = Ocaml_version.v 5 1`, either of these conversions return `5.1`, however, the solver expects a fully qualified version, such as `5.1.0`.

```
Fmt.str "%a" Ocaml_version.pp ocaml_version   (* 5.1 *)
Ocaml_version.to_string ocaml_version (* 5.1 *)
```

Possibly a neater solution would be `Ocaml_version.to_string Ocaml_version.Releases.v5_1`.  See my new commit.

PR #904, opened 2023-11-15
Update opam lint

moyodiallo:
Instead of using a specific commit of opam-dune-lint, now we're using the released version. And that was the goal since the opam lint job is added.


benmandrew (approved):
Looks great, thanks!

PR #905, opened 2023-11-15
Increase the analysis pool

moyodiallo:
The CI and Solver-service is stable now, this is why this increase.


mtelvers (commented):
The graph of incoming queue length on the scheduler shows that the solver queue never has any backlog of jobs. This is because the solver pool capacity is 392, and the analyse pool is only 150. The graph of the number of running jobs shows a maximum of ~150. We would expect to see shorter, overall, solve times by increasing the size of the analyse pool. The value could be increased to 400 rather than 300.

PR #906, opened 2023-11-20
Add Cumulus as an organization allowed to use ocaml-ci

dinosaure:
Cumulus is currently the owner of `Syndic` which is a library used by `ocaml.org`. It will be nice to have an access to `ocaml-ci` and check if Syndic is in a good shape.

### DIFF
--- a/deploy-data/github-organisations.txt
+++ b/deploy-data/github-organisations.txt
@@ -94,3 +94,4 @@ moyodiallo
 edwintorok
 smuenzel
 9glenda
+Cumulus


mtelvers:
We would be delighted for you to run OCaml-CI against `Syndic`.
We have recently changed the onboarding process of OCaml-CI.  See PR #892.  We now update only `deploy-data/github-organisations.txt` and rebuild `stack.yml` using a `Makefile`.  This ensures that everything remains correctly synchronised.

dinosaure:
I rebased this PR according to your request :+1:.

PR #907, opened 2023-11-21
Fix ocamlformat analysis limiting the other jobs

### LINKED ISSUE #893: Missing `.ocamlformat` version line causes analysis failure
Issue opened from comment by @gpetiot on issue #834, as the technical background is slightly different:

> We still observe this on ocamlformat's repository: https://github.com/ocaml-ppx/ocamlformat/runs/17517278251 (we don't have a set version in the .ocamlformat file because we use the version of the same commit to format ocamlformat's codebase)

I do wonder where the line is between supporting special use-cases and missing actual errors that should be surfaced in a normal use-case. You could imagine someone with an accidentally malformed `.ocamlformat` file wondering why it's not being tested.

@tmcgilchrist WDYT?

moyodiallo:
Attempt to fix #893, instead of failing on the analysis, the fmt job is spawned and the error is notified in it (suggestion from @Julow).


benmandrew:
This works if the `.ocamlformat` file has no `version=` line 👍.

However, if the version line contains an incorrect version &mdash; such as `version=0.99.1` &mdash; then the `(lint-fmt)` job is not spawned as no analysis solution is found. @moyodiallo do you think it'd be good to also have this case propagated as an error to a `(lint-fmt)` job? We'd have to be careful to do this only when the `.ocamlformat` file is present, so that we don't fail CI for projects without formatting.

benmandrew (approved):
If not, I'm happy to have this merged.

moyodiallo:
> However, if the version line contains an incorrect version — such as version=0.99.1 — then the (lint-fmt) job is not spawned as no analysis solution is found. @moyodiallo do you think it'd be good to also have this case propagated as an error to a (lint-fmt) job? We'd have to be careful to do this only when the .ocamlformat file is present, so that we don't fail CI for projects without formatting.

Yeah I think you're right it is logic, when we have `.ocamlformat` file present in a project, the purpose is to format the project. And also (lint-fmt) job is always spawned.

moyodiallo:
@benmandrew  Is this OK to your suggestion ?

benmandrew:
I'm not sure if the behaviour's changed or I missed it before, but if the `.ocamlformat` file doesn't have a `version=` line then the job fails in the `extract` stage with:

```
2023-11-29 13:22.17: Exec: "git" "ls-files" "**/.ocamlformat-enable" 
                           ".ocamlformat-enable"
2023-11-29 13:22.17: Job failed: Missing 'version=' line in .ocamlformat
```

Screenshot 2023-12-11 at 16 36 42

My test repo/branch is https://github.com/benmandrew/ci-test-repo/tree/ocamlformat-no-version.

moyodiallo:
I think you're using the `master` branch, not the PR one.

benmandrew:
My mistake, I was looking at an old job! Feel free to merge.

PR #913, opened 2023-12-14
Improve Docker Compose setup

benmandrew:
Changes:
- Use Caddyfile in repository instead of requiring the user to create their own in `/etc/caddy/Caddyfile`
- Move `rsync` installation into `docker-compose.yml` for the worker so that we don't need the shim Dockerfile
- Add log verbosity levels to `service` and `web-ui`


benmandrew:
It sounds obvious when you say it like that! I originally copied a lot from `stack.yml`, so good catch, thank you.

I have made the same correction to the opam-repo-ci Docker Compose PR.

PR #929, opened 2024-04-23
Update submodules to include Fedora 40

mtelvers:


test/service/test_spec.ml
```diff
@@ -7,7 +7,7 @@ let expected_macos_spec =
  (env OPAMCOLOR always)
  (run (shell "ln -f ~/local/bin/opam-2.1 ~/local/bin/opam"))
  (run (shell "opam init --reinit -ni"))
- (run (shell "opam exec -- ocaml -version && opam --version"))
+ (run (shell "uname -rs && opam exec -- ocaml -version && opam --version"))
```
shonfeder, line 10: Given the name of the value, `expected_macos_spec`, this doesn't seem related to the Fedora update? Is it?

What is this testing against that requires the kernel name to printed here?

shonfeder (commented):
Just one question on the test updates for my edification.

test/service/test_spec.ml
mtelvers: These lines are the obuilder job specification, which are sent to obuilder when a OCaml-CI tests a package.  We recently added `uname -rs` to the spec so that the log includes the output from that command.  e.g. https://ocaml.ci.dev:8100/job/2024-04-07/070141-ci-ocluster-build-ac53b3#L93.  This can be helpful on FreeBSD or macOS where the base images are manually updated and we only support the _current_ release.  I noticed the test failed when I updated the submodules so added a commit to fix it.

shonfeder: I see! Thank you for explaining.

shonfeder: Ah, I see even better after digging in a bit. IIUC, this will fix the failing build that is showing up on master here, e.g., https://github.com/ocurrent/ocaml-ci/runs/23123013559

Issue #931, opened 2024-05-02
debian-12 x86_32 workers fail due to old capnproto version


shonfeder:
An example of a failing build can be seen at 
https://ocaml.ci.dev/github/ocurrent/docker-base-images/commit/d91f3800a8897fd11753fd312f5ea11f240363bd/variant/debian-12-4.14_x86_32_opam-2.1/-/refs/pull/274/head

The relvent error is
```
# *** Uncaught exception ***
# kj/filesystem-disk-unix.c++:305: failed: ::fstat(fd, &stats): Value too large for defined data type
```

@benmandrew noted that this is the same problem discussed at
https://github.com/mirage/capnp-rpc/issues/273 

@mtelvers pointed out that the problem here is simply that package repo for
debian 12 has not updated to use the fixed capnproto version (e.g., 1.0.2)
https://tracker.debian.org/pkg/capnproto .

I'm trying to get that update moved along, but in the meantime I believe we can
fix this by [installing the latest release](https://capnproto.org/install.html)
in our images before running CI on them. Iiuc, https://github.com/ocurrent/ocaml-ci/pull/894 would have fixed the problem while waiting for the upstream updates, but it was closed because

> it is preferable to not add patched system libraries in the CI

as per https://github.com/ocurrent/ocaml-ci/pull/894#issuecomment-1808344120

However, at this point a proper release is out, and in my view it is preferable to special case install of a package rather than to leave builds red, so I will prep a PR following that example but using the latest install.

shonfeder:
Thanks to the debian IRC, TIL:

> package versions are frozen for every debian stable release, so unless the fix can be backported non-destructively to 0.9.2, bookworm will not get it

So if we want to continue support this build target,our (non-exclusive) options are:

1. Install the latest version in our CI pipeline
2. Work towards, and wait for, a backport to 0.9.2

However, we have now been letting this build fail for months, iiuc, and it seems like everything is moving along ok. So maybe the best option is:

3. Just remove the tests for debian-12 on x86_32. 

WDYT, @mtelvers?

mtelvers:
These tests only fail for projects which use [capnproto](https://capnproto.org) as that is the outdated package.  Therefore, I do not think we should disable the test as it will work on most projects.

shonfeder:
[...]

PR #932, opened 2024-05-15
Exclude debian-12 x86_32 from builds with capnproto

shonfeder:
Debian 12, "bookwork" is currently on a stable package repo, and package versions are frozen. The frozen version of capnproto is 0.9.2. This version does not support large files on 32 bit arch, and this can result in specious test and build failures.

These false positives have been giving us noisy red trunks for a long while, but this fix should clean up our builds until we either drop support for this target or the fix needed is backported to 0.9.2.

Mitigates #931, however that should  remain open until we have a fix, rather than a workaround.

### DIFF
--- a/lib/pipeline.ml
+++ b/lib/pipeline.ml
@@ -77,7 +77,7 @@ let get_job_id x =
    platforms unsupported by some required package. *)
 let excluded_selection =
   let is_debian_12_x86_32 (v : Variant.t) =
-    match Variant.distro v, Variant.arch v with
+    match (Variant.distro v, Variant.arch v) with
     | "debian-12", `X86_64 -> true
     | _ -> false
   in
--- a/lib/pipeline.ml
+++ b/lib/pipeline.ml
@@ -73,6 +73,27 @@ let get_job_id x =
   let+ md = Current.Analysis.metadata x in
   match md with Some { Current.Metadata.job_id; _ } -> job_id | None -> None
 
+(* Identifies "selections" on which we don't want to run the CI, e.g., due to
+   platforms unsupported by some required package. *)
+let excluded_selection =
+  let is_debian_12_x86_32 (v : Variant.t) =
+    match Variant.distro v, Variant.arch v with
+    | "debian-12", `X86_64 -> true
+    | _ -> false
+  in
+  let is_conf_capnproto package_name =
+    match Astring.String.cut ~sep:"." package_name with
+    | Some ("conf-capnproto", _) -> true
+    | _ -> false
+  in
+  fun (s : Selection.t) ->
+    (* TODO Remove when (a) we stop building on 32bit Debian 12 or (b) Debian
+       bookworm backports the need fix to capnproto to support large files on 32
+       bit systems.
+
+       See https://github.com/ocurrent/ocaml-ci/issues/931 *)
+    is_debian_12_x86_32 s.variant && List.exists is_conf_capnproto s.packages
+
 let docker_specs ~analysis =
   let+ analysis = Current.state ~hidden:true analysis in
   match analysis with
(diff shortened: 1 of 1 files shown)


lib/pipeline.ml
```diff
@@ -89,22 +110,8 @@ let docker_specs ~analysis =
...
           :: Spec.opam_monorepo builds
       | `Opam_build selections ->
-          let selections =
-            (* TODO: This filter will be removed when a new version of debian in which capnproto has this fix "https://github.com/capnproto/capnproto/pull/1830".
-               * Related to this issue "https://github.com/mirage/capnp-rpc/issues/273" *)
-            List.filter
-              (fun s ->
-                not
-                  (Variant.arch s.Selection.variant == `I386
-                  && Variant.arch s.Selection.variant == `Aarch32
```
shonfeder, line 99: Unless I am missing something, this filter could never match anything as is, since it is impossible for the same value to `==` both of these variants.

lib/pipeline.ml
```diff
@@ -89,22 +110,8 @@ let docker_specs ~analysis =
             ~analysis (`Lint `Fmt)
           :: Spec.opam_monorepo builds
       | `Opam_build selections ->
-          let selections =
-            (* TODO: This filter will be removed when a new version of debian in which capnproto has this fix "https://github.com/capnproto/capnproto/pull/1830".
```
shonfeder, line 93: The referenced issue was resolved last year: https://github.com/capnproto/capnproto/pull/1830

lib/pipeline.ml
```diff
@@ -89,22 +110,8 @@ let docker_specs ~analysis =
...
-                  && List.find_opt
-                       (fun pkg ->
-                         Astring.String.cut ~sep:"." pkg
-                         |> Option.map fst
-                         |> Option.equal String.equal (Some "conf-capnproto"))
-                       s.Selection.packages
-                     |> Option.is_some))
-              selections
+          let included_selections =
+            List.filter (fun s -> not (excluded_selection s)) selections
```
shonfeder, line 114: An assumption I am making, based on https://github.com/ocurrent/ocaml-ci/pull/902 being proposed and deployed, is that this filter will take account of all of a package's dependencies, and not just direct dependencies. If this assumption is wrong, then we may still hit build failures in CI due to transitive deps.

I have spent a bit of time tracing thru the code, but it hasn't been clear to me one way or another. Any input correcting my assumption if its in error would be helpful and save some time.

shonfeder:
There's a CI failure on the linting job I'm troubleshooting (tho which I think must be unrelated):

```
/src: (run (cache (opam-archives (target /home/opam/.opam/download-cache)))
           (network host)
           (shell "opam depext -i conf-m4 && opam depext -i dune 'odoc>=1.5.0'"))
Opam plugin "depext" may require upgrading/reinstalling. Reinstall the plugin on the current switch? [Y/n] y
[WARNING] Opam package conf-git.1.1 depends on the following system package that can no longer be found: git
The following actions will be performed:
  - recompile opam-depext 1.2.1-1


-> retrieved opam-depext.1.2.1-1  (cached)
-> removed   opam-depext.1.2.1-1
-> installed opam-depext.1.2.1-1
Done.


<><> opam-depext.1.2.1-1 installed successfully <><><><><><><><><><><><><><><><>
=> opam-depext is unnecessary when used with opam >= 2.1. Please use opam install directly instead
# Run eval $(opam env) to update the current shell environment


<><> Carrying on to "opam depext -i conf-m4" ><><><><><><><><><><><><><><><><><>


You are using opam 2.1+, where external dependency handling has been integrated: consider calling opam directly, the 'depext' plugin interface is provided for backwards compatibility only
# Detecting depexts using vars: arch=x86_64, os=linux, os-distribution=alpine, os-family=alpine
[WARNING] Opam package conf-git.1.1 depends on the following system package that can no longer be found: git
# The following system packages are needed:
m4
[WARNING] Opam package conf-git.1.1 depends on the following system package that can no longer be found: git
opam: PACKAGES... arguments: Invalid character in package name
      "\027[01;04mconf-git\027[0m\027[04m"
Usage: opam install [OPTION]... [PACKAGES]...
Try `opam install --help' or `opam --help' for more information.
"/usr/bin/env" "bash" "-c" "opam depext -i conf-m4 && opam depext -i dune 'odoc>=1.5.0'" failed with exit status 2
2024-05-17 15:44.40: Job failed: Failed: Build failed
```

But this should be ready for review in the meanwhile.

benmandrew (approved):
Looks good to me

shonfeder:
The failing linting error here is not related to this change, but instead (I think) must be due to a change in the container the job is running in. I just reran the linting job on the latest commit to master here, and got the same error (https://ocaml.ci.dev/github/ocurrent/ocaml-ci/commit/f544ddc4809c029767774e6cd1964b6a7d24c673/variant/%28lint-doc%29) even tho that error was not being produced when the commit was first merged. I'll track this new problem in its own issue.

PR #941, opened 2024-05-31
Add Puneeth and Shon as admins

shonfeder:
We aren't now, but we should be. While reviewing this bit of code, maybe good to see if anyone is currently listed as an admin who shouldn't be?

### DIFF
--- a/service/github.ml
+++ b/service/github.ml
@@ -4,9 +4,10 @@ let has_role user = function
   | `Builder | `Admin -> (
       match Option.map Current_web.User.id user with
       | Some
-          ( "github:talex5" | "github:avsm" | "github:kit-ty-kate"
-          | "github:mtelvers" | "github:samoht" | "github:tmcgilchrist"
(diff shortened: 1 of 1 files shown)


punchagan (approved):

service/github.ml
```diff
@@ -1,14 +1,26 @@
+let administrators =
+  [
+    "github:avsm";
+    "github:benmandrew";
+    "github:dra27";
+    "github:kit-ty-kate";
```
mtelvers, line 6:
```suggestion
```

Perhaps take the opportunity to remove Kate

mtelvers (approved):
I'm not sure that the refactoring was necessary, and this function now differs from the other implementations across the projects such as service/main.ml in opam-repo-ci.

shonfeder:
I opted to make the smaller change Mark advised. A refactoring pass to clean up and remove duplication can be done later. Thanks for the reviews!

Issue #944, opened 2024-06-04
Promote FreeBSD and MacOS to non-exprimental 


shonfeder:
MacOS and FreeBSD are currently tagged as "experimental platforms":

https://github.com/ocurrent/ocaml-ci/blob/bd28414e04b4d743f290a0f6e3ff27ad8797cef4/lib/build_info.ml#L10-L37

Which means that even if those CI checks have failed or are not finished, the overall status of a job can show as green. This recently led to a problem, reported by @polytypic, where the development CI was showing all green, masking non-completing FreeBSD jobs that were actually signs of a problem. 

MacOS was added in cffaa14c320cc2803722f36bc1bcb9d208b86264, in Oct of 2022 and marked as experimental in Dec with b8f1733 .

FreeBSD was added more recently, just under a year ago, in c2df7fea51f0f166c06c19d2ff0f4fcb73a7ed0b.

However, we think it may be time to promote these variants to non-experimental targets.

Thoughts?

polytypic:
Yes.  If some platform on OCaml-CI is too unstable for some reason for some project I'm developing, I might, for example, exclude that platform (at the package level) or skip tests via other means on OCaml-CI for that platform.

mtelvers:
I would like to see both FreeBSD and macOS promoted out of `experimental`.

tmcgilchrist:
I agree they are both stable and have sufficient capacity to be promoted
out of experimental.

PR #949, opened 2024-06-11
Remove experimental tag from macOS and FreeBSD

mtelvers:
Closes https://github.com/ocurrent/ocaml-ci/issues/944

### DIFF
--- a/lib/build_info.ml
+++ b/lib/build_info.ml
@@ -12,12 +12,9 @@ let of_label label = { label; variant = None }
     If it is experimental we allow those builds to fail without failing the
     overall build for a commit. *)
 let experimental_variant s =
-  if
-    Astring.String.(
-      is_prefix ~affix:Variant.lower_bound_label s.label
-      || is_prefix ~affix:Variant.opam_label s.label)
-  then true
-  else false
+  Astring.String.(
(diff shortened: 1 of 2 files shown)


lib/build_info.ml
```diff
@@ -17,12 +17,7 @@ let experimental_variant s =
       is_prefix ~affix:Variant.lower_bound_label s.label
       || is_prefix ~affix:Variant.opam_label s.label)
   then true
-  else
-    match s.variant with
-    | None -> false
-    | Some v ->
-        Astring.String.equal "freebsd" (Variant.distro v)
-        || Astring.String.equal "macos-homebrew" (Variant.distro v)
+  else false
```
shonfeder, line 20: You could remove the conditional if you felt inclined, and just use the disjunction, as on lines 29 and 30, but I could also see an argument for leaving space for else conditions, and it's not a biggie either way.

shonfeder (approved):
Looks great! Thanks for taking care of this :)

lib/build_info.ml
mtelvers: I have removed it - my original thinking was it would have been nice to leave the logic in place to help a future person add a new experimental platform, but they can just check the old commits/git blame.

PR #942, opened 2024-05-31
Fix order of steps in lint-fmt job

shonfeder:
Reworking fixing an error I introduced in
https://github.com/ocurrent/ocaml-ci/pull/937

PR #959, opened 2024-08-02
Fix order of steps in lint-fmt job

mtelvers:
Reworking fixing an error Shon introduced in
https://github.com/ocurrent/ocaml-ci/pull/937

This commit is on `live-engine` but does not appear on `master`

### DIFF
--- a/lib/lint.ml
+++ b/lib/lint.ml
@@ -64,18 +64,20 @@ let fmt_spec ~base ~ocamlformat_source ~selection =
           Option.value ~default:commit (commit_from_ocamlformat_source source)
         in
         let network = [ "host" ] in
-        install_ocamlformat source
+        [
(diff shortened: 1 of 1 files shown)


shonfeder:
The PR to fix this initially in https://github.com/ocurrent/ocaml-ci/pull/942 was into master. Do happen to know how it got removed?

shonfeder:
Huh... looking at the history here, I am quite confused about why I didn't merge this into master in the first place via my PR, why it was deployed an not merged etc... :/ 

I clearly messed up there, so thanks for the fix. IMO, this points to the need for a cleaner CD story here (and a staging env so we don't need to test on prod). But those needs are tracked already. Thank you for the fix and sorry for the trouble!

shonfeder (approved):

PR #966, opened 2024-08-27
Put an upper bound on tar-unix

shonfeder:
Corresponds to https://github.com/ocaml/opam-repository/pull/26331/files

Fix the breakage in master present since https://github.com/ocurrent/ocaml-ci/commit/0679e7f05749f70bd07597f7ff2aa03d03886fef

I'll also do followups to the ocurrent stack.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -44,8 +44,8 @@
   (opam-0install (>= 0.4.3))
   (ppx_deriving (>= 5.1))
   (ppx_deriving_yojson (>= 3.7))
-  (tar (>= 2.4.0))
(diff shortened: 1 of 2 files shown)


mtelvers (approved):
Reworking OCaml-CI to use tar-unix >= 3.0.0 looks non-trivial.  Thus this seems a suitable workaround, particularly given the deprioritisation of OCaml-CI.

shonfeder:
Yeah. We'll have to start with updating obuilder too.  Ill open an issue to track that but its not a hi priority item to my mind. Thanks for the review!

PR #972, opened 2024-09-14
Add license

shonfeder:
Adds an MIT license, as per Tarides's policy.

If any contributors feel that they should be included in the copy right notice explicitly,
please add yourself, in this PR or in a followup!

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,21 @@
+# MIT License
+
+Copyright (c) 2019-2024 Tarides <contact@tarides.com>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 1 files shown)


punchagan (approved):
I wonder if maybe we should switch `opam-repo-ci` to the same license too, given we are using the MIT license for `ocaml-ci` from which the project was forked, IIUC.

shonfeder:
Yeah, we could go either way, if we want to reconcile them.

Also, this license addition will be our most complex, I think, as we have quite a few contributors here outside of the org.

If we can identify that their contributions are just to add their org/name, maybe we can make a case for not having to contact all? But otherwise, I think we'll need that dance again :)

shonfeder:
IIUC, looking at the contributor graph, all substantial contributions here came from people employed at Tarides, and therefore working with an explicit understanding that the work must bear an OSS license. The rest of the contribs are people adding they accounts to opt in to the service.

If anyone takes issues with the addition of this license, please let us know, and we can either remove your contributions here or adjust the license as needed.

PR #977, opened 2024-10-16
Support architectures not present in the master distro.

tmcgilchrist:
e.g. RiscV on Ubuntu

This needs more testing but it should allow RiscV Linux builds (if we have capacity).


tmcgilchrist:
No longer needed we get debian-13-4.14_riscv64_opam-2.5 workers now.

PR #979, opened 2024-10-18
Add personal projects to ocaml-ci

raphael-proust:
Not sure if there's a procedure for this, do let me know if that's the case.

My use case is partly to get familiar with this system and partly because those personal projects are published to opam so I'd like to smooth out the release process.


mtelvers (approved):
The procedure to follow for getting started is given here: https://ocaml.ci.dev/getting-started. Please let us know if you have any problems

PR #982, opened 2024-11-18
Use opam 2.3 and rename freebsd to freebsd-14.1

mtelvers:


shonfeder (approved):
Thanks for continuing to propagate this update :)

Issue #990, opened 2024-12-04
ocaml-ci is breaking builds that use vendored sub modules


shonfeder:
## Context

This is showing up in CI jobs for projects that use vendored dependencies (via git submodules) as

```
...
[ERROR] The compilation of ocluster-api.dev failed at "dune subst".
...
#=== ERROR while compiling ocluster-api.dev ===================================#
# context     2.3.0 | linux/x86_64 | ocaml-base-compiler.5.2.1 | pinned(file:///src/ocluster)
# path        ~/.opam/5.2/.opam-switch/build/ocluster-api.dev
# command     ~/.opam/5.2/bin/dune subst
# exit-code   1
# env-file    ~/.opam/log/ocluster-api-1-306113.env
# output-file ~/.opam/log/ocluster-api-1-306113.out
### output ###
# File ".", line 1, characters 0-0:
# Error: There is no dune-project file in the current directory, please add one
# with a (name <name>) field in it.
# Hint: 'dune subst' must be executed from the root of the project.
```

I've seen it in these two builds, so far:

- https://ocaml.ci.dev/github/ocurrent/ocaml-ci/commit/cc9bac43c99dedc42a48b4dc1982157f87a98132/variant/alpine-3.20-5.2_opam-2.3
- https://github.com/ocurrent/ocurrent-deployer/runs/33726197377

## Step(s) to reproduce

This is showing up in ocaml-ci running on itself, starting with https://github.com/ocurrent/ocaml-ci/commit/cc9bac43c99dedc42a48b4dc1982157f87a98132

## Expected behaviour

It should build vendored dependencies OK.

## Technical details

Since I *think* the earliest this showed up is in the CI results run on https://github.com/ocurrent/ocaml-ci/commit/cc9bac43c99dedc42a48b4dc1982157f87a98132 it's possible that some of the upstream changes pulled in from that change are responsible. But this is just a guess as I have not had time to investigate properly yet.

Issue #991, opened 2024-12-04
The 'lint-doc' job is failing on odoc.


jonludlam:
## Context

The 'lint-doc' job is failing on odoc, meaning we have a red cross on all ocaml-ci jobs right now.
See https://ocaml.ci.dev/github/ocaml/odoc/commit/822d266232fccdffbd4922434c81c45ab6d583f4/variant/%28lint-doc%29 as an example.

The failure seems to be happening while it's trying to build odoc-parser, and is complaining about the lack of a `dune-project` file. 

To be honest, the 'lint-doc' job is of very limited value right now, though we can do a way better job of it when we release `odoc-driver`.

punchagan:
Thanks for the report @jonludlam! The current failure seems like a duplicate of #990 which @shonfeder is investigating. The note about switching to `odoc-driver` could be left as an issue for tackling, if you like.

jonludlam:
I did a bit of investigation. It's failing because:

1. It pins odoc, odoc-parser, etc to /src/ without installing anything but the opam files
2. It then tries to 'opam install' odoc, which with dune 3.16 completed without error, even though it just installed a totally empty package. On dune 3.17 it complains (not unreasonably!).
3. Only then does it actually install the source.

It seems likely this is the cause of the problems in #990 too. 

Permuting the two lines (install src, then install odoc) almost works - I needed to explicitly ask to install `odoc-parser` as well as `odoc`. However, if you do this you lose an awful lot of the caching that ocaml-ci gives you. 

One possibility is to install the deps before pinning, then pin, then copy the files, then install.

shonfeder:
Fixed (for now) by https://github.com/ocurrent/ocaml-ci/pull/993.  Thanks for the report!

PR #993, opened 2024-12-04
Fake `dune-project` file

mtelvers:
Closes https://github.com/ocurrent/ocaml-ci/issues/990 and https://github.com/ocurrent/ocaml-ci/issues/991

This PR creates a minimal `dune-project` file which is good enough for `dune subst` not to fail.

An empty file is not good enough.  It must have at least the `lang` stanza.  However, which `lang` version should be specified as potentially, if a project specifically requests dune version 2, then this hack won't work.


mtelvers:
Some issues with this on Windows

mtelvers:
I have tried copying all the `dune-project` from the source tree along with the `.opam` files but these complete `dune-project` files fail to build (understandably!). This approach is also problematic as not every project will have a `dune-project` file.

shonfeder (approved):
LGTM as a workaround and it is proven to solve the problem currently. Thanks!

PR #1004, opened 2025-03-20
Update to FreeBSD 14.2

MisterDA:


mtelvers:
Thanks.  I'll need to update the FreeBSD workers before this can be committed.

mtelvers (approved):

PR #1030, opened 2025-10-16
Updates (Dockerfile, cmdliner shell completions, …)

MisterDA:


mtelvers:
Thanks for this, and apologies it sat for so long. Revisiting it now, most of the changes have since been overtaken by other work on `master`:

- The Dockerfiles are already on debian-13 / opam-2.4 with updated opam-repository pins.
- The vendored `ocurrent/*.opam` pinning has been replaced by released `current` packages.
- GitLab support has been removed entirely (#1052), so `Dockerfile.gitlab` and `ocaml-ci-gitlab.opam` no longer exist.
- The bulk of the remaining diff is ocamlformat 0.27.0 reflow.

Rather than rebase this across all the conflicts, I'm going to close it. If we want to adopt cmdliner 2.0 (shell completions) or bump ocamlformat to 0.27.0, those are best done as their own focused PRs. Thanks again!

Issue #1045, opened 2026-04-20
OCaml 5.5 is not tested


kit-ty-kate:

mtelvers:
https://github.com/ocurrent/ocaml-ci/pull/1050 adds OCaml 5.5.0~beta1

PR #1050, opened 2026-04-28
Platform matrix trim

mtelvers:
- Build only `OV.Releases.significant` rather than `OV.Releases.recent`
- Drop i386 and Aarch32 builds


mtelvers:
Closes https://github.com/ocurrent/ocaml-ci/issues/1045

PR #1052, opened 2026-04-30
Drop gitlab support

tmcgilchrist:
OCaml CI's GitLab pipeline is not in use anymore and can be dropped. The permissions setup has been broken for some time and only I've noticed :-) 

@mtelvers b416602 drops references to gitlab pipeline configuration, along with that we need to remove that configuration from the server.


mtelvers (approved):
LGTM. A couple of tiny references remain:

These two doc-comments still mention GitLab:
- lib/build.mli:12 — (e.g. GitHub or GitLab)
- lib/cluster_build.mli:21 — (e.g. GitHub or GitLab)

These two static files have no references after the diff:
- web-ui/static/images/gitlab-logo-500.png
- web-ui/static/images/index-page-search.png (its only reference in web-ui/view/documentation/index_page.ml was removed here)

PR #1055, opened 2026-05-12
Reset opam-repository before opam init --reinit

mtelvers:
Fix for opam 2.4+ ignoring opam-repository updates after format-upgrade init. See commit message for the full diagnosis and reproduction notes.


jonludlam:
:+1: we hit this failure mode in the odoc tests, and it was very confusing!

